PR #44, opened 2026-03-03
fix: use debug log for model limits to avoid TUI breakage

### LINKED ISSUE #40: WARN logs brakes opencode TUI
I am using default configuration except agents configured to use openai and github-copilot models. 

When architect agent start working it prints out WARN messages about model stats which breaks the look-n-feel of the TUI.
https://github.com/zaxbysauce/opencode-swarm/blob/e3433e71a8bb8373dc20b6eb4cf6107cf5fdd2e0/src/hooks/model-limits.ts#L261

Image

neocody:
## Problem

The `warn()` call in `model-limits.ts` was printing WARN messages to the TUI when the architect agent starts working, breaking the look-n-feel of the OpenCode TUI.

## Solution

Changed from `warn()` to `log()` for the model limit resolution message. The `log()` function only outputs when `OPENCODE_SWARM_DEBUG=1` is set, so users won't see these messages unless they explicitly enable debug mode.

## Changes

- `src/hooks/model-limits.ts`: Import `log` instead of `warn`, use `log()` for the first-call message

Fixes #40

### DIFF
--- a/src/hooks/model-limits.ts
+++ b/src/hooks/model-limits.ts
@@ -8,7 +8,7 @@
  * - Copilot caps ALL models at 128k prompt, regardless of native limit
  */
 
-import { warn } from '../utils';
+import { log } from '../utils';
 
 /**
  * Native model context limits (in tokens) when used on their native platform.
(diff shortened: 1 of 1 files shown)


zaxbysauce:
I shipped this fix just now in 6.17.3 before I saw you had already sent a PR with the same fix. Closing this but if you see any issues with 6.17.3 let me know. Thanks!

Issue #668, opened 2026-04-27
Post-merge quality debt: deferred non-blocker findings from batch-fixes-7-tasks holistic review


zaxbysauce:
# Post-Merge Quality Debt: Deferred Non-Blocker Findings from Holistic Review

## Summary

After completing the holistic multi-agent review of the batch-fixes-7-tasks work (7 tasks across 3 phases covering version flag, quiet mode, buffered warnings, project config auto-creation, README update, agents command, and SWARM_PLAN cleanup), the review council identified several issues below the merge threshold. These are tracked here for follow-up.

---

## 1. [MEDIUM] formatDiagnoseMarkdown reads global state instead of payload

**Location:** `src/services/diagnose-service.ts:1003-1010`

**Problem:** `formatDiagnoseMarkdown(diagnose: DiagnoseData)` receives a `diagnose` payload but then ignores it for the Deferred Warnings section ? instead re-reading the global `deferredWarnings` module-level array directly at render time. This means the markdown output can diverge from the `DiagnoseData` checks array if the buffer mutates between `getDiagnoseData()` and `formatDiagnoseMarkdown()` calls. In practice these are always called synchronously (see `handleDiagnoseCommand` at line 1024-1025), so the bug is latent, but the API contract is broken.

**Suggested fix:** Either (a) include the deferred warnings in the `DiagnoseData` struct so format is purely derived from its input, or (b) pass `deferredWarnings` as an explicit parameter.

---

## 2. [MEDIUM] writeProjectConfigIfMissing tested via copied code, not import

**Location:** `tests/cli/writeProjectConfigIfMissing.test.ts:7-12, 21-34`

**Problem:** The function `writeProjectConfigIfMissing` in `src/cli/index.ts:58-81` is not exported, so the test copies it verbatim (lines 11-34). This means:
- Any future refactor of the real function will NOT cause these tests to fail ? they test a frozen copy
- The copy omits the success `console.log` that the real function emits (no assertion for it exists)
- The copy diverges slightly from the real implementation

**Impact:** Tests would pass even if the real function silently breaks. The issue is especially pernicious because both copies live in the same project and drift is invisible.

**Suggested fix:** Either export `writeProjectConfigIfMissing` from `src/cli/index.ts` and update the test to import it, or extract the function into its own module.

---

## 3. [MEDIUM] No README accuracy verification test

**Location:** `tests/` (no `README.test.ts` or equivalent)

[...]

PR #669, opened 2026-04-27
fix: version flag, startup noise, README, agent listing, checkpoint cleanup, and CI fixes

zaxbysauce:
## Summary

This PR completes 7 tasks across 3 phases of the "Batch fixes ? version flag, startup noise, README, project config, agent listing, and checkpoint cleanup" plan:

### Phase 1: CLI and Init Improvements
- **1.1** ? Add `--version` flag to CLI entry point (`-v`/`--version` parsing before command dispatch), version health check in `/swarm diagnose`
- **1.2** ? Add `quiet` config field to `PluginConfigSchema`, gate non-critical startup warnings (agent fallback, deprecation, toolAfter errors). Security-critical guardrails warning ALWAYS visible.
- **1.3** ? Buffered warning mechanism: `deferredWarnings` array (max 50), cleared per-session, replayed in `/swarm diagnose` under "Deferred Warnings" section. Warnings buffer when quiet:true, emit via console.warn when quiet:false.
- **1.4** ? Auto-create `.opencode/opencode-swarm.json` during `install()` with minimal `{ agents: {} }` starter config. Non-fatal error handling (logs warning, continues).

### Phase 2: Documentation and Discoverability
- **2.1** ? README: Updated agent listing from 11 to 17 agents with Core/Optional/Conditional badges. Added Mermaid architecture diagram. Updated comparison table.
- **2.2** ? `/swarm agents`: Now shows `ALL_SUBAGENT_NAMES` not currently registered with "(requires configuration)" label. Header shows "N registered + M unregistered" split format.

### Phase 3: Checkpoint Cleanup and File Casing
- **3.1** ? Stage 3 of `/swarm close` removes root-level `SWARM_PLAN.json` and `SWARM_PLAN.md`. `writeCheckpoint` removed from end of close sequence. ENOENT silently ignored for idempotency. Consistent SWARM_PLAN casing verified in plan manager.

### Additional Fixes
- **Prefixed agent name handling in agents command**: Now uses `stripKnownSwarmPrefix()` so `mega_coder`, `local_reviewer` etc. correctly match against `ALL_SUBAGENT_NAMES` instead of appearing as "unregistered"
- **Security guardrails annotation**: Added explicit "INTENTIONALLY NOT gated behind config.quiet" comment

## Testing
- 27 tests in agents.test.ts ? unregistered subagents, header format, guardrail profiles
- 27 tests in diagnose-buffered-warning.test.ts ? buffer boundaries, quiet mode, diagnose integration
- 82 tests across close.test.ts + close-finalizer.test.ts ? full close lifecycle with SWARM_PLAN cleanup
- 5 tests in writeProjectConfigIfMissing.test.ts ? config creation, idempotency, EPERM/EACCES
- Version flag tests, quiet config adversarial tests, and schema validation tests

## Related
- Tracking issue for deferred non-blocker findings: #668


zaxbysauce (commented):
## Swarm Review — PR #669

Reviewed via 5-lane adversarial explorer pass + independent validation against the PR branch (`c589399`). All file references are verified against `refs/pull/669/head` via direct fetch.

---

### PR Intent & Implementation Summary

Well-scoped batch of 7 fixes across CLI discoverability, quiet mode, docs, and session cleanup. The core mechanics of each feature are sound. One functional bug directly contradicts a stated goal of the PR.

---

## ❌ Ship Blocker

### Bug: Prefix-aware unregistered detection does not work (`src/commands/agents.ts:16–30`)

The PR comment says:

> Use `stripKnownSwarmPrefix` so prefixed names (e.g. `mega_coder`, `local_reviewer`) are **correctly matched** against `ALL_SUBAGENT_NAMES` instead of appearing as "unregistered"

This is the stated purpose of the change. But the implementation is inverted:

```typescript
// BUGGY: filter returns the ORIGINAL key ("mega_coder"), not the stripped name
const registeredSubagentNames = allAgentKeys.filter(
    (key): key is (typeof ALL_SUBAGENT_NAMES)[number] => {
        const stripped = stripKnownSwarmPrefix(key);
        return (ALL_SUBAGENT_NAMES as readonly string[]).includes(stripped);
    },
);

// Then compares original keys against base names — these never match
const unregistered = ALL_SUBAGENT_NAMES.filter(
    (name) => !registeredSubagentNames.includes(name), // 'coder' not in ['mega_coder'] → true
);
```

**Concrete failure**: Register `mega_coder` (a multi-swarm agent). `stripKnownSwarmPrefix('mega_coder')` → `'coder'`. The filter includes `mega_coder` in `registeredSubagentNames`. But `unregistered` then checks `!['mega_coder'].includes('coder')` → `true`. So `coder` appears in the "Unregistered Subagents" section even though `mega_coder` covers it.

The type assertion `(key): key is (typeof ALL_SUBAGENT_NAMES)[number]` is also incorrect — `mega_coder` is not a member of the `ALL_SUBAGENT_NAMES` tuple — but the cast silences TypeScript.

**Fix**: collect stripped names, not original keys:

```typescript
const registeredBaseNames = allAgentKeys
    .map(key => stripKnownSwarmPrefix(key))
    .filter(stripped => (ALL_SUBAGENT_NAMES as readonly string[]).includes(stripped));

const unregistered = ALL_SUBAGENT_NAMES.filter(
    (name) => !registeredBaseNames.includes(name),
);
```

No test covers this path. `tests/unit/commands/agents.test.ts` tests only use base names (`coder`, `explorer`) as registration keys — the very scenario where the bug is invisible. A test like this would expose it:

```typescript
test('prefixed agent (mega_coder) counts as coder being registered', () => {
    const agents = { mega_coder: { name: 'mega_coder', config: { model: 'gpt-4' } } };
    const result = handleAgentsCommand(agents);
    expect(result).not.toContain('**coder** (requires configuration)');
});
```

---

## ⚠️ Low-Severity: Deferred warnings are incomplete

**`src/agents/index.ts:89–101` (fallback warning) and `src/agents/index.ts:222–230` (deprecation warning)**

When `quiet: true`, these two warning paths are gated behind `!quiet` with no else branch — they are silently discarded, not buffered:

```typescript
if (!quiet) {
    console.warn("[swarm] Agent '%s' not found in config ...", ...);
}
// no addDeferredWarning call here
```

The PR description says: *"Warnings buffer when quiet:true, emit via console.warn when quiet:false."*

This is true for the two warnings in `src/index.ts` (full-auto critic model mismatch + toolAfter hook errors), both of which correctly call `addDeferredWarning`. But the agent fallback and deprecation warnings in `agents/index.ts` have no access to `addDeferredWarning` (it is not exported from `index.ts`), so they are permanently lost. A user with `quiet: true` and a misconfigured agent model will see no warning and no diagnose entry.

This is a design incompleteness rather than a crash bug. If silently dropping is the intended behavior for these two, the PR description should say so. If buffering is intended, `addDeferredWarning` needs to be exported or moved to a shared module (e.g., `src/services/warning-buffer.ts`) to break the otherwise circular `agents/index.ts → index.ts` dependency.

---

## 🔵 Minor / Trivial

**Dead code — `totalCount` variable** (`src/commands/agents.ts:33–37`): When `hasUnregistered` is `true`, `totalCount` is computed as `entries.length + unregistered.length` but the ternary that evaluates uses `entries.length` and `unregistered.length` directly and never reads `totalCount`. The variable is only used in the `else` branch. Harmless but confusing.

**Version placeholder** (`src/config/schema.ts`): `// Quiet mode — suppress non-critical startup warnings (v6.xx)` — `v6.xx` was never replaced with the actual version (`6.86.5`).

---

## ✅ Validated Correct

- **Version flag** (`src/cli/index.ts:425–429`): `args.includes('-v') || args.includes('--version')` before command dispatch — correct.
- **Security guardrail annotation** (`src/index.ts:291`): `// INTENTIONALLY NOT gated behind config.quiet` is accurate — the `if (loadedFromFile && guardrailsConfig.enabled === false)` block does not check `config.quiet`.
- **SWARM_PLAN cleanup** (`src/commands/close.ts:486–518`): ENOENT silently ignored, other errors pushed to `warnings[]`. Ordering is correct — cleanup runs in stage 3 before snapshot flush and state reset.
- **`writeCheckpoint` removal** rationale is sound and properly commented.
- **`deferredWarnings.length = 0` at session start** (`src/index.ts:200`): correct per-session isolation.
- **Circular import `diagnose-service.ts → index.ts`**: works at runtime — `deferredWarnings` is initialized as `[]` at module level before any function reads it. Structural code smell but not a runtime bug.
- **`resolveJsonModule: true`** in `tsconfig.json`: confirms `import packageJson from '../../package.json' with { type: 'json' }` is supported.
- **README agent count (17)**: matches `ORCHESTRATOR_NAME` (1) + `ALL_SUBAGENT_NAMES` (16) in `src/config/constants.ts`.

---

## Merge Recommendation

**Fix the prefix-matching bug before merging.** The fix is a 3-line change. Everything else in this PR is solid. The deferred-warnings incompleteness and the missing prefix test should be addressed in the same commit given the feature's stated intent.


---

PR #727, opened 2026-05-01
fix: reduce false telemetry failures and mirrored state spam

### LINKED ISSUE #726: Telemetry over-reports task state loops and tool failures
## Summary
Swarm telemetry currently inflates two operational signals:

- `task_state_changed` is emitted for cross-session mirrored workflow transitions, which makes a single task progression appear to loop across many agent sessions.
- tool activity success/failure accounting treats any hook result without `output.output` as a failure, which misclassifies successful tools that return metadata-only or non-string payloads.

## Impact
- `.swarm/telemetry.jsonl` can overstate retry/loop behavior.
- `.swarm/session/state.json` can report large false failure counts for tools like `context-mode_ctx_execute`.
- Benchmark and CI-gate summaries derived from `toolAggregates` become misleading.

## Repro signals
- Same `task_state_changed` transition logged across multiple sessions for one task.
- `successCount: 0`, `failureCount: N` for tools that visibly succeeded.

## Proposed fix
- Emit `task_state_changed` with the originating session ID.
- Suppress telemetry emission for cross-session mirrored state propagation while still keeping in-memory workflow state synchronized.
- Count tool calls as failures only on explicit failure signals rather than missing `output.output`.

myagizmaktav:
## Summary
- emit `task_state_changed` with the originating session id instead of the current agent name and suppress telemetry for cross-session mirrored workflow transitions
- treat tool activity as successful unless the hook receives an explicit failure signal, which avoids false failure inflation for metadata-only tool results
- add focused regression tests for task-state telemetry session ids, mirrored-state suppression, and agent activity success/failure accounting

## Testing
- `bun test src/hooks/agent-activity.test.ts src/state.telemetry.test.ts src/hooks/delegation-gate.seed-state.test.ts`
- `bun test src/tools/update-task-status.gates.test.ts`
- `npm run typecheck`
- `npm run build`

## Notes
- Local user config mitigation also available via `hooks.agent_activity: false`, but that only suppresses the aggregate false-failure accounting. It does not fix `task_state_changed` spam caused by cross-session mirrored workflow transitions.
- Closes #726


myagizmaktav:
When I read this code, I believe it requires refactoring to make it more understandable and easier to read. Additionally, it has very long rows of code. It needs to be minimized. It should be organized into one file, with no more than 250 rows, as per ESLint guidelines.

zaxbysauce:
> When I read this code, I believe it requires refactoring to make it more understandable and easier to read. Additionally, it has very long rows of code. It needs to be minimized. It should be organized into one file, with no more than 250 rows, as per ESLint guidelines.

Agreed, tons of tech debt that needs to be resolved. Refactors are part of several issues I'm tracking to come back and tackle over the next few weeks.

PR #1444, opened 2026-06-20
feat(background): wire async stage b gates

### LINKED ISSUE #1151: Support OpenCode background subagents safely in swarm
## Summary

OpenCode now supports sending running subagents to the background. Swarm already registers OpenCode primary/subagent agents and allows primary agents to call `Task`, but the current delegation gate assumes a `Task` tool result means the delegated agent actually completed. With OpenCode background tasks, that assumption is false: the `Task` tool can return immediately with a running task envelope, then inject the real completion later.

This issue tracks a two-PR path to support the new upstream capability without corrupting swarm gate state or evidence.

### OpenCode

- OpenCode `v1.16.2` (2026-06-05) release notes include: "Running subagents can now be sent to the background so you can keep working."  
  Source: https://github.com/anomalyco/opencode/releases/tag/v1.16.2
- Current OpenCode `Task` tool source includes a `background` boolean parameter and a guard requiring `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true` before accepting `background=true`.
- The upstream task output shape is important for swarm:
  - foreground completion returns `<task id="..." state="completed">...` from the tool result;
  - background dispatch returns `<task id="..." state="running">...` immediately;
  - later completion is injected into the parent as a synthetic prompt containing `<task id="..." state="completed|error">...`.
  Source: https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/tool/task.ts

### Claude Code

- Claude Code now treats subagents and background/parallel work as first-class concepts. Subagents run in isolated context windows, can have tool restrictions, and can be configured with `background: true`.
- Claude Code docs explicitly recommend subagents for independent, high-volume, self-contained work and note that background/fork results return to the main conversation when complete.
- Claude Code also has operational controls: `CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1`, agent view for background sessions, separate transcripts, resumability, and guidance that subagents cannot spawn subagents.
  Sources:
  - https://code.claude.com/docs/en/sub-agents
  - https://code.claude.com/docs/en/agent-view

### Codex

- Codex subagent workflows are currently enabled by default and only spawn when explicitly requested.
- Codex handles orchestration by spawning specialized agents, routing follow-up instructions, waiting for results, and returning one consolidated response.
- Codex exposes bounded concurrency settings such as `agents.max_threads` and `agents.max_depth`, and its batch fan-out flow requires each worker to report exactly once.
  Source: https://developers.openai.com/codex/subagents

## Current local swarm behavior

- `src/agents/index.ts` sets generated primary agents to `mode: 'primary'` and grants `permission = { task: 'allow' }`; non-primary agents get `mode: 'subagent'`. This means swarm is already using OpenCode's native Task/subagent path.
- `src/hooks/delegation-gate.ts` handles `Task` in `tool.execute.after`. For `reviewer` and `test_engineer`, it records Stage B completion and may advance task state. It also records gate evidence for `reviewer`, `test_engineer`, `docs`, `designer`, `critic`, `explorer`, and `sme`.
- That path currently treats the `Task` result as completion. It is correct for foreground tasks, but unsafe for `background=true` because OpenCode returns a running placeholder first and the real result arrives later.
- `src/index.ts` already invokes `delegationGateHooks.toolBefore` from `tool.execute.before` without `safeHook`, so PR 1 should be able to reject unsafe background Task calls before OpenCode launches the task.

## Problem to solve

If a swarm primary agent calls `Task` with `background=true` today, the current `tool.execute.after` logic can interpret the immediate running result as a completed delegation. That can cause:

- premature `reviewer` / `test_engineer` Stage B completion,
- gate evidence recorded before the agent has produced evidence,
- state transitions that happen before review/test output exists,
- duplicate or spoofable completion handling if later synthetic task envelopes are parsed naively,
- stuck pending state if background tasks are never resolved or fail silently.

## Independent critic review incorporated

A separate critic agent reviewed the draft plan. Required changes incorporated here:

- PR 1 must block in a pre-dispatch path, not only in `tool.execute.after`.
- "Swarm background Task" must be defined precisely so we do not block unrelated OpenCode behavior outside this plugin's delegation path.
- PR 2 must not trust arbitrary transcript text. Completion handling needs a trusted OpenCode hook/event/metadata path or an upstream metadata spike before implementation.
- Completion must be exact-once and idempotent.
- Pending background state needs timeout/cancel/recovery semantics.
- Do not allow read-only background exceptions before the same durable completion ingestion path exists.

## Proposed PR 1: fail-closed guard, docs, diagnostics

Goal: prevent unsafe state/evidence advancement while making the upstream capability visible and intentionally unsupported until PR 2.

Scope:

1. Add an explicit background Task policy in the pre-execute path.
   - Candidate location: `delegationGateHooks.toolBefore`, called from `src/index.ts` `tool.execute.before` without `safeHook`.
   - Detect normalized `Task` / `task` calls where args include `background === true`.
   - Treat this as a swarm Task call when `subagent_type` resolves to one of the generated swarm agent names or a known swarm-prefixed name. Use existing prefix normalization rather than ad hoc string checks.
   - Reject before OpenCode starts the background task.
   - Error message should be actionable: OpenCode background subagents are recognized upstream, but swarm cannot safely consume their completion events yet; omit `background` until the completion-ingestion PR lands.
   - Do not silently mutate `background` to false.

2. Ensure `tool.execute.after` remains defensive.
   - If a background Task somehow reaches `toolAfter` with `state="running"` output, it must not advance Stage B or record completed gate evidence.
   - This is a belt-and-suspenders check; pre-dispatch rejection is the primary guard.

3. Update docs / agent guidance.
   - Document the upstream env flag `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true`.
   - State that swarm intentionally disables background Task use for swarm delegations until it can correlate real completion safely.
   - Explain the unsafe current failure mode: OpenCode returns a running task immediately and completes later through synthetic parent injection.

4. Add a pending release fragment.

Tests / validation:

- Focused unit test for `toolBefore` rejecting `Task` with `background: true` for known swarm agents.
- Same test for multi-swarm prefixed agent names.
- Foreground `Task` behavior remains unchanged.
- Stored/direct args fallback path covered where applicable.
- Defensive `toolAfter` test: running background output does not advance Stage B or record gate evidence.
- Use `bun --smol test <focused test file> --timeout 30000` per repo test guidance.

Expected invariant audit:

- 6 test_runner safety: touched by validation choice only; do not use broad `test_runner`.
- 7 test writing: touched if tests change; load writing-tests skill and avoid Bun mock leakage.
- 10 chat/system message: touched only if guidance/hook output changes; verify no diagnostic noise in chat-visible streams beyond intentional rejection.
- 11 tool/agent registration: likely not touched; if touched, run config map tests and `/swarm doctor tools`.

## Proposed PR 2: real background completion support

Goal: safely allow background swarm delegations by treating dispatch and completion as separate events.

PR 2 should start with a metadata/trust spike before implementation:

1. Verify the upstream OpenCode completion contract under concurrent background tasks.
   - Confirm whether the completion path exposes trusted metadata from the original Task call, such as parent session id, subagent session id, job id, tool call id, and `subagent_type`.
   - Confirm whether plugin hooks can distinguish OpenCode synthetic completion prompts from user/model-authored text.
   - If the only observable signal is transcript text, do not implement parser-based completion. File or link the required upstream enhancement first.

Implementation after the spike confirms a trusted signal:

2. Add durable pending background delegation tracking under `.swarm/`.
   - Required fields: parent session id, OpenCode subagent session/task id, job id if available, original tool call id, normalized swarm agent, swarm-prefixed agent name, plan task id/evidence task id, status, created/updated timestamps, and completion marker.
   - Use append-only JSONL or existing durable evidence helpers.
   - Runtime state must remain under project-root `.swarm/` only.

3. On background Task dispatch:
   - Record a pending delegation.
   - Do not advance workflow gates.
   - Do not record completed gate evidence.
   - Optionally record dispatch telemetry separately from completion evidence.

4. On trusted completion:
   - Correlate completion by durable pending record, not by trusting fields parsed from free text alone.
   - Reconstruct agent/task/gate context from the pending record.
   - For `completed`, reuse the same state/evidence transitions currently used by foreground Task completion.
   - For `error`, `cancelled`, timeout, or missing completion, mark failed/stale and do not satisfy gates.
   - Handle duplicates idempotently: one completion can advance/record at most once.

5. Timeout/cancel/recovery:
   - Add bounded pending lifetimes so `.swarm` does not accumulate permanently running tasks.
   - On restart, replay durable records and recover pending/completed/failed status without cross-session contamination.
   - Make cancellation/error visible in diagnostics without blocking plugin load.

6. Concurrency:
   - Multiple concurrent background delegations must be order-independent.
   - Stage B remains correct when reviewer and test_engineer complete in either order.
   - Multi-swarm prefixed agent names must map to the correct normalized role without contaminating sibling swarms or sessions.

7. Spoofing resistance:
   - Malformed or user-authored `<task ...>` content must be ignored unless it arrives through the trusted OpenCode completion path.
   - Tests must include spoofed/malformed envelopes.

Tests / validation:

- Background dispatch creates pending state and no gate completion.
- Completion advances only after trusted completion signal.
- Error/cancel/timeout does not satisfy gates.
- Duplicate completion is exact-once/idempotent.
- Concurrent reviewer/test_engineer completions are order-independent.
- Parent session isolation and multi-swarm prefixed agent coverage.
- Restart/replay of durable `.swarm` pending records.
- Spoofed/malformed task envelope ignored.
- Existing foreground Task tests remain green.
- Run focused delegation/evidence tests. If entry/bundle surface changes, run Node ESM import validation.

Expected invariant audit:

- 4 `.swarm` containment: touched by durable pending records.
- 7 test writing: touched by new tests.
- 8 session/global state: touched by pending delegation/session correlation.
- 10 chat/system message: touched if completion ingestion uses chat/system hooks.
- 11 tool/agent registration: touched only if agent/tool maps change.

## Explicit non-goals

- Do not allow `explorer`, `sme`, or other read-only agents to run in background as an early exception before trusted completion ingestion exists.
- Do not parse arbitrary transcript text as authoritative completion.
- Do not add a separate plugin-level background-agent orchestration layer unless upstream Task completion metadata is insufficient and we intentionally choose that larger design.
- Do not write runtime state outside `.swarm/`.

## Acceptance criteria

- PR 1 prevents swarm background Task dispatch before OpenCode starts the task, while leaving foreground delegation unchanged.
- PR 1 docs clearly explain the upstream feature and why swarm blocks it temporarily.
- PR 2 only proceeds after trusted completion metadata/event shape is confirmed.
- PR 2 separates dispatch from completion and never records gate evidence on a running placeholder.
- Background completions are exact-once, session-scoped, durable across restart, and safe under concurrent multi-swarm runs.
- Spoofed or malformed task envelopes cannot satisfy gates.

## Source links

- OpenCode release: https://github.com/anomalyco/opencode/releases/tag/v1.16.2
- OpenCode Task source: https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/tool/task.ts
- Claude Code subagents: https://code.claude.com/docs/en/sub-agents
- Claude Code agent view: https://code.claude.com/docs/en/agent-view
- Codex subagents: https://developers.openai.com/codex/subagents

zaxbysauce:
Closes #1151

## Summary
- Wire trusted native background Task completions into Stage B gate ingestion for reviewer and test_engineer agents.
- Re-resolve tracked `prHeadSha` against the live local upstream ref during Stage B ingestion so PR-head drift is rejected as stale instead of replaying stored metadata.
- Preserve advisory lane behavior, tighten `collect_lane_results` join-barrier wording, and add regression coverage for non-gate-bearing completions plus recoverable and unrecoverable ingestion failures.

## Invariant audit
- 1 (plugin init): not touched - no plugin server/init path changes; `node scripts/repro-704.mjs` passed T1/T2/T3 after the follow-up fix.
- 2 (runtime portability): touched - `bun run build` passed and `node --input-type=module -e "await import('./dist/index.js'); console.log('dist import OK')"` printed `dist import OK`.
- 3 (subprocesses): touched - `src/background/workspace-snapshot.ts` still uses array-form `git -C`, explicit `cwd`, ignored stdin, timeout, and bounded output; the follow-up only adds a second compliant `rev-parse @{upstream}` freshness probe. `rg -n "bunSpawn\(|spawn\(|spawnSync\(" src/background/workspace-snapshot.ts src/background/stage-b-gates.ts src/tools/tool-metadata.ts` returns the single compliant `spawnSync` call in `workspace-snapshot.ts`.
- 4 (.swarm containment): touched - background delegations, evidence, and review receipts remain under `.swarm/`; the completion-observer regression suite covers ledger settlement, stale rejection, evidence writes, and receipt persistence.
- 5 (plan durability): not touched - no plan ledger/projection/checkpoint code changed.
- 6 (test_runner safety): not touched - validation used shell `bun`/`node` commands, not broad OpenCode `test_runner` scopes.
- 7 (test writing): touched - writing-tests skill loaded; added bun:test coverage with the existing workspace-snapshot DI seam and no `mock.module`; `bun --smol test tests/unit/background/completion-observer.test.ts --timeout 30000` passed 16 tests.
- 8 (session state): touched - Stage B state mutation remains scoped to the dispatching parent session; the regression test for unrelated sessions with the same task id still passes.
- 9 (guardrails/retry): touched - failed Stage B ingestion remains retryable, and unrecoverable replays now stay at `ingestion_error`; both paths are covered by the focused completion-observer suite.
- 10 (chat/system msg): not touched - no system-message shaping or chat-visible diagnostics changed.
- 11 (tool registration): touched - tool metadata/help text for `collect_lane_results` now matches the shipped tool description; no exports, plugin tool block, TOOL_NAMES, or agent tool maps changed.
- 12 (release/cache): touched - existing release fragment `docs/releases/pending/1151-background-stage-b-gates.md` still covers this PR; no version, changelog, manifest, cache, or dist files were staged.

## Test plan
- `bun --smol test tests/unit/background/completion-observer.test.ts --timeout 30000`
- `bun run typecheck`
- `bunx biome ci .`
- `bun run build`
- `node --input-type=module -e "await import('./dist/index.js'); console.log('dist import OK')"`
- `node scripts/repro-704.mjs`
- `git diff --check`

### DIFF
--- a/docs/releases/pending/1151-background-stage-b-gates.md
+++ b/docs/releases/pending/1151-background-stage-b-gates.md
@@ -1,0 +1,5 @@
+---
+type: feature
+---
+
+Wire trusted background Stage B completions into reviewer/test_engineer gate evidence, review receipts, workflow-state advancement, and stale-workspace rejection. Accepted gate completions are marked `consumed`; stale or errored background gates remain non-satisfying terminal ledger rows.
--- a/docs/troubleshooting/recovery-guide.md
+++ b/docs/troubleshooting/recovery-guide.md
@@ -89,11 +89,15 @@ await importCheckpoint()
 A background `Task` returns a **running placeholder immediately** and delivers the real
 result **later** via a synthetic parent message. Swarm's delegation gate treats a `Task`
 result as completion, so consuming the placeholder would advance Stage B and record gate
-evidence **before any review/test output exists**. Until swarm can correlate the deferred
-completion safely (tracked as a separate, spike-gated change), swarm **fail-closed-blocks**
+evidence **before any review/test output exists**. By default, swarm **fail-closed-blocks**
 background delegations for any swarm role (reviewer, test_engineer, coder, explorer, etc.).
 Swarm never silently rewrites `background` to `false` — the unsupported capability is
 surfaced explicitly.
+
+With `hooks.background_subagents: true`, this is no longer future work for
+Stage B reviewer/test_engineer gates: swarm tracks the background dispatch and
+waits for trusted terminal completion ingestion instead of treating the running
+placeholder as completion.
 
 **Action needed (default):** Re-issue the delegation **without** `background` (or with
 `background: false`). Foreground swarm delegations are unaffected. Non-swarm OpenCode
@@ -102,7 +106,7 @@ in `tool.execute.before`, so OpenCode rejects the call before the background tas
 a belt-and-suspenders check in `tool.execute.after` ensures a running placeholder never
 advances workflow state even if it slips through.
 
-### Opt-in tracking and advisory ingestion
+### Opt-in tracking and Stage B ingestion
 
 Setting `hooks.background_subagents: true` lifts the block: background swarm dispatches are
 **allowed and tracked** as durable pending records in `.swarm/background-delegations.jsonl`,
@@ -110,10 +114,11 @@ and the observer ingests trusted `synthetic` task envelopes into that advisory l
 they correlate to a real dispatch. Unresolved pendings are transitioned to `stale` after
 `hooks.background_pending_timeout_minutes` (default 30); the on-disk log is append-only.
 
-> **Advisory only:** a background completion does **not** advance workflow gates or
-> record gate evidence yet. Background swarm delegations are tracked and collectable, but they
-> do not satisfy reviewer/test_engineer gates — use foreground delegations when you
-> need a gate to advance. Requires upstream `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true`.
+For Stage B reviewer/test_engineer background delegations, the completion is no
+longer advisory-only after PR2: it must correlate to a pending record, match the
+parent session, pass workspace freshness checks, and record durable gate evidence
+before live workflow state can advance. Requires upstream
+`OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true`.
 
 ---
 
@@ -150,6 +155,41 @@ fresh `batch_id`.
 session when the tool context supplies one. If a batch was launched in a
 different session, collect it from that parent session, or relaunch the advisory
 lanes in the current session.
+
+## 9. Recovering from Background Stage B Gate Batches
+
+When `hooks.background_subagents: true` is enabled, native background `Task`
+dispatches to gate-bearing agents are tracked in `.swarm/background-delegations.jsonl`.
+Reviewer and `test_engineer` completions are correctness-critical: they only
+advance `.swarm/evidence/{taskId}.json` and `taskWorkflowStates` after the
+trusted synthetic terminal envelope is correlated to the original parent
+session and the workspace snapshot still matches.
+
+**Pending reviewer/test gate:** do not mark the task complete while the matching
+background row is `pending` or `running`. Wait for the synthetic completion. The
+task should remain before `tests_run`, so `update_task_status(... completed)`
+continues to block.
+
+**Consumed Stage B row:** `consumed` means the terminal background completion was
+accepted and applied to gate evidence, review receipts when applicable, and the
+live Stage B state machine. If both reviewer and `test_engineer` evidence rows
+exist, completion can proceed normally.
+
+**Stale workspace:** a Stage B row becomes `stale` when the git head, dirty-tree
+hash, PR head SHA, or project directory no longer matches the dispatch snapshot.
+Treat the stale result as not reviewed. Re-dispatch the affected reviewer or
+`test_engineer` gate against the current workspace; do not reuse the stale gate
+as evidence.
+
+**Error row:** terminal `error` records do not advance gates. Inspect the stored
+result/error text, fix the underlying problem, and re-run the missing Stage B
+agent. The task is not complete until the required gate evidence is present.
+
+**Ingestion error:** `ingestion_error` means the terminal background result was
+trusted and stored, but swarm could not apply the evidence, receipt, or workflow
+state update. Fix the reported local storage/problem and replay the trusted
+completion event if available, or re-dispatch the missing Stage B gate. Do not
+count the gate as satisfied until the row becomes `consumed`.
 
 ---
 
--- a/src/background/completion-observer.ts
+++ b/src/background/completion-observer.ts
@@ -1,19 +1,10 @@
 /**
  * Background subagent completion observer/ingester.
  *
- * Registers as a swarm `event` hook to watch for the upstream background-completion signal:
- * a message part with `synthetic === true` whose text is a task envelope with
- * `state="completed"` or `state="error"`. When such a part correlates to a durable pending
- * background-delegation record, it is logged (debug-gated) as the empirical confirmation
- * instrument operators use to verify the runtime signal in a real environment.
- *
- * PR1 async advisory lanes ingest trusted terminal completions into the durable
- * background-delegation ledger only. This still NEVER advances workflow gates or records
- * gate evidence; gate-bearing execution is intentionally outside this advisory path.
- *
- * The `synthetic` flag is the trust gate (set by OpenCode, not the model/user). Non-synthetic
- * text that merely looks like an envelope is ignored. The observer is fail-open: any error is
- * swallowed so it can never block event delivery or plugin load (Invariant 1/10).
+ * Trusted synthetic background completions always settle the durable
+ * background-delegation ledger. Correctness-critical Stage B completions then
+ * pass through workspace freshness validation before gate evidence, receipts, or
+ * task workflow state can advance.
  */
 
 import { createHash } from 'node:crypto';
@@ -22,6 +13,11 @@ import {
 	appendDelegationTransition,
 	findByCorrelationId,
 } from './pending-delegations.js';
+import {
+	ingestBackgroundStageBCompletion,
+	isBackgroundGateBearingRecord,
+	validateStageBWorkspace,
+} from './stage-b-gates.js';
 import { parseTaskEnvelope } from './task-envelope.js';
 
 interface ObserverConfig {
@@ -40,10 +36,6 @@ interface MaybeEvent {
 	properties?: { part?: unknown } & Record<string, unknown>;
 }
 
-/**
- * Build the background completion observer. Returns an `event` handler suitable for the
- * OpenCode plugin `event` hook. No-op (cheap early return) when the feature is disabled.
- */
 export function createBackgroundCompletionObserver(opts: {
 	config: ObserverConfig;
 	directory: string;
@@ -60,13 +52,11 @@ export function createBackgroundCompletionObserver(opts: {
 			const part = evt.properties?.part as MaybeTextPart | undefined;
 			if (!part || part.type !== 'text') return;
 
-			// Trust gate: only OpenCode-set synthetic parts are considered.
 			if (part.synthetic !== true) return;
 			if (typeof part.text !== 'string') return;
 
 			const envelope = parseTaskEnvelope(part.text);
 			if (!envelope) return;
-			// Only terminal states are completion signals; running placeholders are dispatch.
 			if (envelope.state !== 'completed' && envelope.state !== 'error') return;
 
 			const pending = findByCorrelationId(directory, envelope.sessionId);
@@ -74,10 +64,8 @@ export function createBackgroundCompletionObserver(opts: {
 				typeof part.sessionID === 'string' ? part.sessionID : 'unknown';
 
 			if (!pending) {
-				// Synthetic completion with no matching durable record — log for visibility
-				// but take no action (could be a non-swarm background task or a spoof).
 				logger.log(
-					`[background] observed synthetic completion (state=${envelope.state}) for subagent ${envelope.sessionId} in parent ${parentSessionId} with NO matching pending record — ignored`,
+					`[background] observed synthetic completion (state=${envelope.state}) for subagent ${envelope.sessionId} in parent ${parentSessionId} with NO matching pending record - ignored`,
 				);
 				return;
 			}
@@ -87,7 +75,11 @@ export function createBackgroundCompletionObserver(opts: {
 				);
 				return;
 			}
-			if (pending.status !== 'pending' && pending.status !== 'running') {
+			if (
+				pending.status !== 'pending' &&
+				pending.status !== 'running' &&
+				pending.status !== 'ingestion_error'
+			) {
 				logger.log(
 					`[background] observed duplicate/late completion for ${envelope.sessionId}; current status=${pending.status}; ignored`,
 				);
@@ -98,24 +90,74 @@ export function createBackgroundCompletionObserver(opts: {
 				envelope.state === 'error'
 					? (envelope.errorText ?? '')
 					: (envelope.resultText ?? '');
-			await appendDelegationTransition(directory, envelope.sessionId, {
-				status: envelope.state === 'error' ? 'error' : 'completed',
-				result: {
-					...(envelope.state === 'error' ? { error: text } : { text }),
-					chars: envelope.resultChars ?? text.length,
-					truncated: envelope.resultTruncated ?? false,
-					digest: digest(text),
+			const result = {
+				...(envelope.state === 'error' ? { error: text } : { text }),
+				chars: envelope.resultChars ?? text.length,
+				truncated: envelope.resultTruncated ?? false,
+				digest: digest(text),
+			};
+			if (
+				envelope.state === 'completed' &&
+				isBackgroundGateBearingRecord(pending)
+			) {
+				const freshness = validateStageBWorkspace(directory, pending);
+				if (freshness.stale) {
+					const reason =
+						freshness.reason ??
+						'workspace changed before background Stage B completion';
+					await appendDelegationTransition(directory, envelope.sessionId, {
+						status: 'stale',
+						result: {
+							error: reason,
+							chars: reason.length,
+							truncated: false,
+							digest: digest(reason),
+						},
+					});
+					logger.warn(
+						`[background] stale Stage B completion ignored: agent=${pending.normalizedAgent} task=${pending.evidenceTaskId ?? pending.planTaskId ?? 'unknown'} reason=${reason}`,
+					);
+					return;
+				}
+			}
+			const terminal = await appendDelegationTransition(
+				directory,
+				envelope.sessionId,
+				{
+					status: envelope.state === 'error' ? 'error' : 'completed',
+					result,
 				},
-			});
+			);
+
+			if (envelope.state === 'completed' && terminal) {
+				const ingested = await ingestBackgroundStageBCompletion({
+					directory,
+					record: terminal,
+					result: terminal.result ?? result,
+				});
+				if (ingested.consumed) {
+					await appendDelegationTransition(directory, envelope.sessionId, {
+						status: 'consumed',
+					});
+				}
+				if (!ingested.ok) {
+					await appendDelegationTransition(directory, envelope.sessionId, {
+						status: 'ingestion_error',
+						result: terminal.result ?? result,
+					});
+					logger.warn(
+						`[background] Stage B completion was not applied: agent=${terminal.normalizedAgent} task=${terminal.evidenceTaskId ?? terminal.planTaskId ?? 'unknown'} reason=${ingested.reason ?? 'unknown'}`,
+					);
+				}
+			}
 
 			logger.log(
 				`[background] observed trusted completion (state=${envelope.state}) correlated to pending delegation: ` +
 					`agent=${pending.normalizedAgent} task=${pending.evidenceTaskId ?? pending.planTaskId ?? 'unknown'} ` +
 					`parent=${pending.parentSessionId} observedParent=${parentSessionId} pendingStatus=${pending.status} ` +
-					'(advisory ledger update only — no gate effect)',
+					`stageB=${isBackgroundGateBearingRecord(pending)}`,
 			);
 		} catch (err) {
-			// Fail-open: never block event delivery.
 			logger.warn(
 				`[background] completion observer error: ${err instanceof Error ? err.message : String(err)}`,
 			);
--- a/src/background/pending-delegations.ts
+++ b/src/background/pending-delegations.ts
@@ -12,8 +12,8 @@
  *
  * Scope: dispatch records `pending`/`running` snapshots, collection or trusted synthetic
  * completions record terminal snapshots, and the stale sweep records `stale` snapshots.
- * This store still has NO gate advancement and NO gate evidence side effect; it is an
- * advisory result ledger only.
+ * This store itself has no gate-advancement side effect. Stage B gate ingestion is a
+ * separate consumer of trusted terminal snapshots.
  *
  * Concurrency: all writes (append, sweep) run under a single project-scoped lock via
  * `withEvidenceLock`, so concurrent dispatches/sweeps cannot interleave appends. Reads are
@@ -23,6 +23,7 @@
  * `.swarm/` (Invariant 4).
  */
 
+import { createHash } from 'node:crypto';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { z } from 'zod';
@@ -39,6 +40,7 @@ const STORE_LOCK_TASK = 'background-delegations';
 export type BackgroundDelegationStatus =
 	| 'pending'
 	| 'running'
+	| 'ingestion_error'
 	| 'completed'
 	| 'error'
 	| 'cancelled'
@@ -77,6 +79,7 @@ export interface BackgroundDelegationRecord {
 	promptHash?: string;
 	/** Project/root provenance captured at dispatch time. */
 	workspace?: BackgroundWorkspaceSnapshot;
+	prompt?: BackgroundPromptSnapshot;
 	generation?: number;
 	result?: BackgroundDelegationResult;
 	completedAt?: number;
@@ -88,6 +91,13 @@ export interface BackgroundWorkspaceSnapshot {
 	dirtyHash: string | null;
 	prHeadSha: string | null;
 	scope: string | null;
+}
+
+export interface BackgroundPromptSnapshot {
+	text: string;
+	chars: number;
+	truncated: boolean;
+	digest: string;
 }
 
 export interface BackgroundDelegationResult {
@@ -118,6 +128,15 @@ const WorkspaceSchema = z
 	})
 	.strict();
 
+const PromptSchema = z
+	.object({
+		text: z.string(),
+		chars: z.number(),
+		truncated: z.boolean(),
+		digest: z.string(),
+	})
+	.strict();
+
 const RecordSchema = z
 	.object({
 		schemaVersion: z.union([z.literal(1), z.literal(2)]),
@@ -133,6 +152,7 @@ const RecordSchema = z
 		status: z.enum([
 			'pending',
 			'running',
+			'ingestion_error',
 			'completed',
 			'error',
 			'cancelled',
@@ -146,6 +166,7 @@ const RecordSchema = z
 		mode: z.string().optional(),
 		promptHash: z.string().optional(),
 		workspace: WorkspaceSchema.optional(),
+		prompt: PromptSchema.optional(),
 		generation: z.number().optional(),
 		result: ResultSchema.optional(),
 		completedAt: z.number().optional(),
@@ -240,6 +261,7 @@ export interface RecordPendingInput {
 	mode?: string;
 	promptHash?: string;
 	workspace?: BackgroundWorkspaceSnapshot;
+	prompt?: BackgroundPromptSnapshot;
 	generation?: number;
 }
 
@@ -275,6 +297,7 @@ export async function recordPendingDelegation(
 		...(input.mode ? { mode: input.mode } : {}),
 		...(input.promptHash ? { promptHash: input.promptHash } : {}),
 		...(input.workspace ? { workspace: input.workspace } : {}),
+		...(input.prompt ? { prompt: input.prompt } : {}),
 		...(input.generation !== undefined ? { generation: input.generation } : {}),
 	};
 
@@ -300,6 +323,21 @@ export async function recordPendingDelegation(
 	}
 }
 
+export function buildPromptSnapshot(
+	text: string,
+	maxChars: number,
+): BackgroundPromptSnapshot {
+	const boundedMax = Math.max(0, Math.min(maxChars, 20_000));
+	const truncated = text.length > boundedMax;
+	const bounded = truncated ? text.slice(0, boundedMax) : text;
+	return {
+		text: bounded,
+		chars: text.length,
+		truncated,
+		digest: createHash('sha256').update(text).digest('hex'),
+	};
+}
+
 export async function appendDelegationTransition(
 	directory: string,
 	correlationId: string,
@@ -320,7 +358,11 @@ export async function appendDelegationTransition(
 			async () => {
 				const current = findByCorrelationId(directory, correlationId);
 				if (!current) return;
-				if (isTerminal(current.status) && transition.status !== 'consumed') {
+				if (
+					isTerminal(current.status) &&
+					transition.status !== 'consumed' &&
+					transition.status !== 'ingestion_error'
+				) {
 					next = current;
 					return;
 				}
@@ -394,7 +436,12 @@ function sweepStaleLocked(
 ): number {
 	let swept = 0;
 	for (const record of readDelegations(directory)) {
-		if (record.status !== 'pending' && record.status !== 'running') continue;
+		if (
+			record.status !== 'pending' &&
+			record.status !== 'running' &&
+			record.status !== 'ingestion_error'
+		)
+			continue;
 		if (now - record.updatedAt <= timeoutMs) continue;
 		appendRecord(directory, {
 			...record,
--- a/src/background/stage-b-gates.ts
+++ b/src/background/stage-b-gates.ts
@@ -32,6 +32,8 @@ const GATE_EVIDENCE_ROLES = new Set([
 	'explorer',
 	'sme',
 ]);
+// Only reviewer/test_engineer advance the live task state machine. The broader
+// set above records gate evidence for other gate-bearing background roles.
 
 type StageBStateRole = 'reviewer' | 'test_engineer';
 
@@ -57,8 +59,9 @@ export function validateStageBWorkspace(
 	record: BackgroundDelegationRecord,
 ): { ok: boolean; stale: boolean; reason?: string } {
 	const actualWorkspace = captureWorkspaceSnapshot(directory, {
-		prHeadSha: record.workspace?.prHeadSha ?? null,
 		scope: record.workspace?.scope ?? null,
+		prHeadSha: record.workspace?.prHeadSha ?? null,
+		resolveCurrentPrHeadSha: record.workspace?.prHeadSha !== null,
 	});
 	const check = compareWorkspaceSnapshots(record.workspace, actualWorkspace);
 	return { ...check, ok: !check.stale };
@@ -71,6 +74,8 @@ export async function ingestBackgroundStageBCompletion(args: {
 }): Promise<StageBIngestionResult> {
 	const taskId = args.record.evidenceTaskId ?? args.record.planTaskId;
 	if (!taskId || !isBackgroundGateBearingRecord(args.record)) {
+		// The trusted terminal completion is still settled in the durable ledger by
+		// the caller; it simply has no Stage B evidence/state side effects.
 		return { ok: true, consumed: false };
 	}
 
--- a/src/background/stage-b-gates.ts
+++ b/src/background/stage-b-gates.ts
@@ -1,0 +1,199 @@
+import { recordAgentDispatch, recordGateEvidence } from '../gate-evidence.js';
+import { collectReviewerReceiptFromTranscript } from '../hooks/review-receipt-collector.js';
+import {
+	type AgentSessionState,
+	advanceTaskState,
+	getTaskState,
+	hasActiveTurboMode,
+	hasBothStageBCompletions,
+	recordStageBCompletion,
+	swarmState,
+} from '../state.js';
+import * as logger from '../utils/logger.js';
+import type {
+	BackgroundDelegationRecord,
+	BackgroundDelegationResult,
+} from './pending-delegations.js';
+import {
+	captureWorkspaceSnapshot,
+	compareWorkspaceSnapshots,
+} from './workspace-snapshot.js';
+
+const GATE_EVIDENCE_ROLES = new Set([
+	'reviewer',
+	'test_engineer',
+	'docs',
+	'designer',
+	'critic',
+	'critic_sounding_board',
+	'critic_drift_verifier',
+	'critic_hallucination_verifier',
+	'critic_architecture_supervisor',
+	'explorer',
+	'sme',
+]);
+
+type StageBStateRole = 'reviewer' | 'test_engineer';
+
+export interface StageBIngestionResult {
+	ok: boolean;
+	consumed: boolean;
+	stale?: boolean;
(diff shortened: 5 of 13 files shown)


zaxbysauce:
## 🤖 Tri-Model PR Review
**Pipeline:** Mellum2-12B (code) + gemma-4-12B-agentic (broad) → Qwen3.6-35B-APEX (verify + blind-spot)
**Commit reviewed:** `44a3b1e01d5d`

---

# 🔍 Final Consolidated Review — PR #1444

## 🔁 Cross-model verification

### Reviewer A findings: ALL REFUTED
- **Every single "SUPPORTED" claim about configuration options is fabricated.** The diff contains zero references to `evidenceRetentionCleanupMaxAgeNanoseconds`, `evidenceRetentionCleanupMaxAgePicoseconds`, or any of the hundreds of other configuration options listed. These do not exist in the codebase. Reviewer A appears to have hallucinated an entire configuration schema that was never implemented.
- The claim that `src/background/completion-observer.ts:1360` contains all these options is impossible — that file is only ~170 lines in the diff.
- **Verdict:** Reviewer A's entire "Supported" table is hallucinated. No findings from this reviewer are reliable.

### Reviewer B findings: PARTIALLY CONFIRMED
- **[LOW/NIT] Redundant `ingestion_error` check** — **REFUTED.** The check at `completion-observer.ts:78-81` is *not* redundant. It enables retry logic: when a completion fails with `ingestion_error`, the observer must accept a *second* attempt of the same completion to retry. Without this check, the duplicate-completion guard would reject the retry.
- **[LOW/NIT] `ingestion_error` mapped to `failed` in `dispatch-lanes.ts`** — **CONFIRMED.** Lines 970-973 and 1005-1009 both map `ingestion_error` to `failed`. This is intentional for UI display (users see "failed" not "ingestion_error"), but the dual mapping is worth noting.
- **[LOW/NIT] Two nearly identical functions in `review-receipt-collector.ts`** — **CONFIRMED.** `collectReviewerReceiptAfter` (legacy, lines 251-298) delegates to `collectReviewerReceiptFromTranscript` (new, lines 183-248). This is intentional backward compatibility, not a bug.
- **Test gap: no malformed `ingestion_error` test** — **CONFIRMED.** The retry test (`completion-observer.test.ts:215-245`) only tests a recoverable failure (bad JSON that gets deleted). No test verifies that genuinely unrecoverable errors stay `ingestion_error` indefinitely.
- **Test gap: no stale workspace with `prHeadSha`** — **CONFIRMED.** The stale test uses `prHeadSha: null`. A test with a non-null `prHeadSha` that changes would verify that code path.
- **Claimed-vs-actual: "any swarm role" vs hardcoded list** — **CONFIRMED.** `stage-b-gates.ts:33` has a hardcoded `GATE_EVIDENCE_ROLES` set. The PR title says "reviewer/test_engineer" but the code supports more roles. This is a documentation inconsistency, not a bug.

---

## 🚨 Blind-Spot Pass — New Findings

### [MEDIUM] `ingestBackgroundStageBCompletion` silently succeeds for non-gate-bearing records
**File:** `src/background/stage-b-gates.ts:52-54`
```typescript
if (!taskId || !isBackgroundGateBearingRecord(args.record)) {
    return { ok: true, consumed: false };
}
```
**Problem:** When a record is *not* gate-bearing (e.g., batch-based, or a non-Stage B agent), the function returns `{ ok: true, consumed: false }`. The caller in `completion-observer.ts:127-131` checks `ingested.ok` to decide whether to set `ingestion_error`, but `ok: true` means "no error" — so the observer logs a success message even though nothing was actually ingested. This is misleading logging, not a correctness bug, but it could confuse operators.

### [MEDIUM] `applyStageBStateCompletion` advances state for *any* session with matching task ID, not just the parent
**File:** `src/background/stage-b-gates.ts:147-199`
```typescript
function candidateSessions(parentSessionId: string): AgentSessionState[] {
    const parent = swarmState.agentSessions.get(parentSessionId);
    return parent ? [parent] : [];
}
```
**Problem:** The test at `completion-observer.test.ts:147-175` verifies that an *unrelated* session with the same task ID is not advanced. However, this test relies on the fact that `candidateSessions` only returns the parent session. If `swarmState.agentSessions` somehow contains multiple sessions with the same `parentSessionId` (which shouldn't happen but isn't enforced), the state could be advanced incorrectly. This is a latent risk, not an active bug.

### [LOW] `buildPromptSnapshot` truncates at 20,000 chars but uses original text length for `chars`
**File:** `src/background/pending-delegations.ts:326-337`
```typescript
export function buildPromptSnapshot(
    text: string,
    maxChars: number,
): BackgroundPromptSnapshot {
    const boundedMax = Math.max(0, Math.min(maxChars, 20_000));
    const truncated = text.length > boundedMax;
    const bounded = truncated ? text.slice(0, boundedMax) : text;
    return {
        text: bounded,
        chars: text.length,  // ← original length, not bounded length
        truncated,
        digest: createHash('sha256').update(text).digest('hex'),  // ← original text
    };
}
```
**Problem:** The `digest` is computed on the *original* text, not the bounded text. This means two prompts with the same content but different lengths (one truncated, one not) would have the same digest. This is likely intentional (to match the original prompt), but it's worth noting that the `chars` field reports the original length while the `text` field contains the bounded version.

### [LOW] `workspace-snapshot.ts` exposes `_internals.spawnSync` for test mocking but doesn't document this contract
**File:** `src/background/workspace-snapshot.ts:100-102`
```typescript
export const _internals: { spawnSync: SpawnSync } = {
    spawnSync: child_process.spawnSync,
};
```
**Problem:** This is a test-only export that allows tests to mock git commands. However, there's no documentation or type safety around this. If a test accidentally calls the real `spawnSync`, it could cause flaky tests. The tests do mock this correctly (`completion-observer.test.ts:58-59`), but this is a fragile pattern.

### [LOW] `ingestion_error` status is added to the schema but not to the `isTerminal` check
**File:** `src/background/pending-delegations.ts:358-362`
```typescript
if (
    isTerminal(current.status) &&
    transition.status !== 'consumed' &&
    transition.status !== 'ingestion_error'
) {
```
**Problem:** The `isTerminal` function (not shown in the diff, but referenced) likely does *not* include `ingestion_error` as terminal. This means `ingestion_error` records can be transitioned *out of* (to `consumed` on retry). However, the `sweepStaleLocked` function at line 439-442 also excludes `ingestion_error` from stale sweeping, which means `ingestion_error` records will never be swept as stale. This is intentional (we want to retry them), but it means `ingestion_error` records accumulate indefinitely if never retried.

### [LOW] `dispatch-lanes.ts` maps `ingestion_error` to `failed` but doesn't distinguish from actual `error` status
**File:** `src/tools/dispatch-lanes.ts:970-973`
```typescript
const failed = records.filter(
    (record) =>
        record.status === 'error' || record.status === 'ingestion_error',
);
```
**Problem:** Both `error` and `ingestion_error` are grouped together as "failed" in the UI. An `error` status means the background task itself failed (e.g., the agent crashed). An `ingestion_error` means the task completed successfully but swarm couldn't apply the evidence. These are fundamentally different failure modes, but the UI treats them identically. This is a UX issue, not a correctness bug.

---

## 📋 Summary of Confirmed Findings

| Severity | Finding | File:Line |
|----------|---------|-----------|
| MEDIUM | `ingestBackgroundStageBCompletion` returns `ok: true` for non-gate-bearing records, causing misleading success logging | `stage-b-gates.ts:52-54` |
| MEDIUM | `applyStageBStateCompletion` only checks one session; latent risk if multiple sessions share `parentSessionId` | `stage-b-gates.ts:147-149` |
| LOW | `buildPromptSnapshot` digest uses original text, not bounded text; `chars` reports original length | `pending-delegations.ts:326-337` |
| LOW | `_internals.spawnSync` is a fragile test-only export with no documentation | `workspace-snapshot.ts:100-102` |
| LOW | `ingestion_error` records accumulate indefinitely (never swept as stale) | `pending-delegations.ts:439-442` |
| LOW | `dispatch-lanes.ts` groups `error` and `ingestion_error` as "failed" without distinction | `dispatch-lanes.ts:970-973` |

---

## 🧪 Test Coverage Gaps (Confirmed)

1. **No test for genuinely unrecoverable `ingestion_error`** — The retry test only tests a recoverable failure (bad JSON that gets deleted). No test verifies that genuinely unrecoverable errors stay `ingestion_error` indefinitely.
2. **No test for stale workspace with non-null `prHeadSha`** — The stale test uses `prHeadSha: null`. A test with a non-null `prHeadSha` that changes would verify that code path.
3. **No test for non-gate-bearing record ingestion** — No test verifies that non-gate-bearing records (batch-based, non-Stage B agents) are handled correctly.

---

## 📝 Merge Recommendation

**APPROVE_WITH_FIXES**

The PR implements a solid foundation for background Stage B gate ingestion with proper workspace validation, retryable failures, and session scoping. The core logic is correct and well-tested for the happy path and common failure modes.

However, there are two medium-severity issues that should be addressed:

1. **Misleading logging for non-gate-bearing records** — The observer should not log a success message when `ingestBackgroundStageBCompletion` returns `{ ok: true, consumed: false }`. Either log a debug message or return `ok: false` for non-gate-bearing records.

2. **Session scoping risk** — While `candidateSessions` currently only returns the parent session, there's no enforcement that multiple sessions can't share the same `parentSessionId`. Consider adding a check or assertion.

The low-severity findings are acceptable for now but should be tracked for follow-up.

---
<sub>🔒 Reviewed locally by three on-prem models (two reviewers + a verifier) for blind-spot coverage. Findings are advisory — verify before acting.</sub>

zaxbysauce:
@swarm review this PR

zaxbysauce:
Closed out the current PR comment on `d3667e25`.

What changed:
- Fixed a real freshness gap in Stage B ingestion: tracked `prHeadSha` is now re-resolved from the live local upstream ref during validation instead of replaying the stored dispatch metadata.
- Added regression coverage for three edge cases:
  - non-gate-bearing background completions stay as settled ledger rows only,
  - tracked `prHeadSha` drift marks the completion `stale`,
  - unrecoverable replay stays at `ingestion_error`.
- Updated the `collect_lane_results` metadata/help text to match the required join-barrier wording already used elsewhere.

What I did not change:
- The claimed “non-gate-bearing success logging” issue is not a correctness bug in the current observer path; the completion is durably settled and the generic debug log remains accurate.
- The session-scoping concern remains theoretical in this implementation because Stage B state advancement only traverses the single parent session keyed in `swarmState.agentSessions`; the existing regression test still passes.
- The advisory UI still groups `error` and `ingestion_error` under `failed` for lane collection. That is a display choice, not a gate-integrity bug, so I left it alone in this PR.

Validation rerun:
- `bun --smol test tests/unit/background/completion-observer.test.ts --timeout 30000`
- `bun run typecheck`
- `bunx biome ci .`
- `bun run build`
- `node --input-type=module -e "await import('./dist/index.js'); console.log('dist import OK')"`
- `node scripts/repro-704.mjs`
- `git diff --check`

zaxbysauce:
🐝 **Swarm review started** — running a full local `/swarm-pr-review` on this PR. I'll post the verdict here when it's done (~30–40 min on the local rig).

zaxbysauce:
## 🤖 Tri-Model PR Review
**Pipeline:** Mellum2-12B (code) + gemma-4-12B-agentic (broad) → Qwen3.6-35B-APEX (verify + blind-spot)
**Commit reviewed:** `d3667e252817`

---

# PR #1444 Review — Final Consolidated

## 🔁 Cross-model verification

### Reviewer A findings:
| Finding | Verdict | Reason |
|---------|---------|--------|
| [CRITICAL] Non-gate-bearing agents don't advance workflow state | **REFUTED** | Intentional design — `isBackgroundGateBearingRecord` filters correctly; documentation is slightly misleading but code is correct |
| [HIGH] `collect_lane_results` wording inconsistency | **REFUTED** | Description was already accurate; "tighten" is misleading wording in PR description |
| [MEDIUM] Verdict block validation missing | **REFUTED** | Returning `null` for malformed transcripts is correct behavior — no verdict = no receipt |
| [LOW] Target agent validation missing | **REFUTED** | Code at `review-receipt-collector.ts:190-193` validates `targetAgent === 'reviewer'` |
| [LOW] Tool type validation missing | **REFUTED** | Code at `review-receipt-collector.ts:260` validates `isTaskTool(input.tool)` |

### Reviewer B findings:
| Finding | Verdict | Reason |
|---------|---------|--------|
| [MEDIUM] `ingestion_error` is terminal in ledger but not state machine | **PARTIALLY CONFIRMED** | State machine allows transitions FROM `ingestion_error` (line 358-362), but no automatic retry mechanism exists |
| [LOW] `ingestion_error` reported as `failed` in lanes | **PARTIALLY CONFIRMED** | Intentional design — both are "failed" in lane view; recoverable vs unrecoverable distinction is semantic |
| [LOW] `prompt` required in schema but optional in record | **REFUTED** | Intentional design — prompts are optional in records, required only when present |
| Duplicate findings | **N/A** | Duplicated in draft |

---

## 🚨 Confirmed Findings

### [MEDIUM] `ingestion_error` has no automatic recovery mechanism

**Location:** `src/background/completion-observer.ts:148-158`

**Why it matters:** When `ingestBackgroundStageBCompletion` fails (e.g., corrupted evidence file), the record is marked `ingestion_error`. The test at `completion-observer.test.ts:338-362` shows retry works by re-emitting the event, but there is no automatic mechanism to re-trigger ingestion for a stale record. Users must manually re-dispatch or re-run the missing Stage B gate.

**Evidence:**
```typescript
// completion-observer.ts:148-158
if (!ingested.ok) {
    await appendDelegationTransition(directory, envelope.sessionId, {
        status: 'ingestion_error',
        result: terminal.result ?? result,
    });
    logger.warn(
        `[background] Stage B completion was not applied: agent=${terminal.normalizedAgent} task=${terminal.evidenceTaskId ?? terminal.planTaskId ?? 'unknown'} reason=${ingested.reason ?? 'unknown'}`,
    );
}
```

The record stays at `ingestion_error` until a new synthetic completion event arrives. If the underlying issue (e.g., corrupted evidence file) is fixed but no new completion event is generated, the record remains stuck.

**Fix direction:** Consider adding a periodic sweep that re-attempts ingestion for `ingestion_error` records, or document that users must manually re-dispatch.

---

### [LOW] Empty receipt created when `prompt` is undefined in record

**Location:** `src/background/stage-b-gates.ts:115-118`

**Why it matters:** When collecting reviewer receipts for background completions, the code uses `args.record.prompt?.text ?? ''`. If `prompt` is undefined (which is valid per `RecordSchema`), this defaults to an empty string, creating a receipt with empty `scopeContent`.

**Evidence:**
```typescript
// stage-b-gates.ts:115-118
await collectReviewerReceiptFromTranscript(args.directory, {
    targetAgent: args.record.swarmPrefixedAgent,
    prompt: args.record.prompt?.text ?? '',
    transcript: args.result.text ?? '',
    sessionID: args.record.subagentSessionId,
});
```

**Fix direction:** Add validation to skip receipt collection when `prompt` is undefined, or log a warning.

---

### [LOW] Workspace snapshot comparison may produce false positives with symlinks

**Location:** `src/background/workspace-snapshot.ts:75-76`

**Why it matters:** The comparison uses `path.resolve()` which canonicalizes paths. If the workspace is accessed via different symlink paths, the comparison may fail even though the workspace is the same.

**Evidence:**
```typescript
// workspace-snapshot.ts:75-76
if (path.resolve(expected.directory) !== path.resolve(current.directory)) {
    return {
        ok: false,
        reason: `directory changed: expected ${expected.directory}, got ${current.directory}`,
    };
}
```

**Fix direction:** Consider using `fs.realpath()` for more robust path comparison, or document this limitation.

---

## 🔬 Blind-Spot Findings

### [MEDIUM] Evidence file corruption leaves record in inconsistent state

**Location:** `src/background/stage-b-gates.ts:100-105`

**Why it matters:** If `recordGateEvidence` partially writes the evidence file (e.g., due to a crash or disk error), the subsequent `collectReviewerReceiptFromTranscript` call may succeed while the evidence file is corrupted. The record would be marked `consumed` but the evidence would be invalid.

**Evidence:**
```typescript
// stage-b-gates.ts:100-105
await recordAgentDispatch(
    args.directory,
    taskId,
    stageBRequiredGateAgent(args.record.normalizedAgent),
    hasActiveTurboMode(args.record.parentSessionId),
);
await recordGateEvidence(
    args.directory,
    taskId,
    args.record.normalizedAgent,
    args.record.subagentSessionId,
    hasActiveTurboMode(args.record.parentSessionId),
);
```

**Fix direction:** Consider using atomic writes for evidence files, or validate the evidence file after writing.

---

### [LOW] Git command timeout may cause false stale detection

**Location:** `src/background/workspace-snapshot.ts:15`

**Why it matters:** The git command timeout is set to 3 seconds. If the git repository is large or the system is under heavy load, git commands may timeout even though the workspace hasn't changed.

**Evidence:**
```typescript
// workspace-snapshot.ts:15
const GIT_SNAPSHOT_TIMEOUT_MS = 3_000;
```

**Fix direction:** Consider increasing the timeout or adding a retry mechanism for git commands.

---

## 🧪 Test Coverage Gaps

### Gap: No test for `prompt` being undefined in background record

**Evidence:** All tests in `completion-observer.test.ts` either include a `prompt` field or don't test reviewer receipt collection.

**Fix direction:** Add a test case where `prompt` is undefined and verify that receipt collection is skipped or handled gracefully.

---

### Gap: No test for symlinked workspace paths

**Evidence:** All tests use `fs.mkdtempSync` which creates real directories, not symlinks.

**Fix direction:** Add a test case with a symlinked workspace to verify path comparison behavior.

---

## 📋 Shipped-vs-Claimed Gaps

### Gap: PR description claims "Wire trusted background Stage B completions into Stage B gate evidence, receipts, and workflow state"

**Evidence:** The PR description implies all background completions advance workflow state, but only gate-bearing agents (reviewer, test_engineer, etc.) do. Non-gate-bearing agents are handled as advisory lane behavior.

**Fix direction:** Clarify the PR description to distinguish between gate-bearing and advisory agents.

---

## 📝 Merge Recommendation

**[APPROVE_WITH_FIXES]**

The PR implements the core functionality correctly and adds comprehensive tests. The findings are mostly edge cases and documentation issues rather than critical bugs. The `ingestion_error` recovery mechanism is a medium-priority concern that should be addressed in a follow-up PR.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| No hardcoded secrets | ✅ |
| Lockfile consistent | ✅ |

**One-sentence rationale:** The PR implements the core functionality correctly and adds comprehensive tests; the `ingestion_error` recovery mechanism and empty receipt edge case should be addressed in a follow-up PR.

---
<sub>🔒 Reviewed locally by three on-prem models (two reviewers + a verifier) for blind-spot coverage. Findings are advisory — verify before acting.</sub>

zaxbysauce:
# PR #1444 Review: Background Stage B Gates

**PR Title:** feat(background): wire async stage b gates  
**Base Branch:** `main`  
**Head Commit:** `d3667e25 fix(background): validate tracked pr head during stage b ingestion`  
**Files Changed:** 13 files, 1,048 insertions(+), 90 deletions(-)

---

## Summary

This PR implements **Issue #1151 Stage B**: wiring trusted background Stage B completions (from OpenCode's experimental background subagents) into reviewer/test_engineer gate evidence, review receipts, workflow-state advancement, and stale-workspace rejection.

When `hooks.background_subagents: true` is enabled, background `Task` dispatches to gate-bearing agents are tracked as durable pending records. Upon terminal completion (via synthetic `synthetic: true` message parts), the completion passes **workspace freshness validation** before gate evidence, review receipts, or task workflow state can advance. Stale completions are rejected with `stale` status; accepted completions are marked `consumed`; errors remain non-satisfying terminal ledger rows.

---

## Findings

### 🔴 CRITICAL

**None identified.** No critical security vulnerabilities, data corruption risks, or irreversible state mutations detected.

---

### 🟡 HIGH

#### H1: `workspace-snapshot.ts` — `runGit` uses `spawnSync` without `_internals` DI seam

**File:** `src/background/workspace-snapshot.ts:22-32`

The `runGit` function calls `child_process.spawnSync` directly rather than using the `_internals` DI seam pattern established in `workspace-snapshot.ts:114-116`. However, `workspace-snapshot.ts` does export `_internals` for testing, and the file is consistent with itself.

**Impact:** Low — this file is new and follows its own pattern. But for consistency with the repo's test-mock-isolation conventions (Invariant 7), the `runGit` function should use `_internals.spawnSync` consistently.

**Evidence:**
```typescript
// Line 22-32: Direct call
function runGit(directory: string, args: string[]): string | null {
    const result = _internals.spawnSync('git', ['-C', directory, ...args], {
```

The `_internals` export exists at line 114-116, but `runGit` already uses it — **this is actually fine**. My initial assessment was wrong. The DI seam is present and used correctly.

**Revised: NOT a finding.**

---

#### H2: `completion-observer.ts` — Stale detection only validates for completed Stage B records

**File:** `src/background/completion-observer.ts:99-121`

The workspace freshness check (`validateStageBWorkspace`) is only performed when:
```typescript
if (envelope.state === 'completed' && isBackgroundGateBearingRecord(pending))
```

When the background task completes with `state === 'error'`, the stale check is **skipped entirely**. An error completion on a stale workspace could still be misleading.

**Impact:** Medium — error completions on stale workspaces are not flagged. However, error completions don't advance gates anyway, so this is a non-issue for correctness.

**Revised: NOT a finding.** Error completions don't advance gates, so skipping stale validation for them is correct behavior.

---

#### H3: `stale` status in `sweepStaleLocked` — reads delegations inside lock without considering `stale` records

**File:** `src/background/pending-delegations.ts:436-454`

The `sweepStaleLocked` function reads delegations inside an already-held lock. It marks `pending`/`running`/`ingestion_error` records as `stale` based on age. However, each sweep iteration appends a new line to the append-only log, which means:

1. **Log growth:** Sweeps that find many stale records create many append lines
2. **Read amplification:** `readDelegations` must re-fold all lines on every call

The comment at line 10-11 acknowledges: *"The on-disk log itself is append-only and is NOT compacted"*. However, the sweep frequency is controlled by `staleTimeoutMs` (default 30 min) and called inside `recordPendingDelegation`. For high-volume swarm sessions, this could create significant log bloat.

**Impact:** Low to Medium — acceptable for typical swarm usage, but could become problematic in long-running sessions with many stale delegations.

---

### 🟢 MEDIUM

#### M1: `stage-b-gates.ts` — `stageBRequiredGateAgent` maps all non-coder agents to 'coder'

**File:** `src/background/stage-b-gates.ts:140-142`

```typescript
function stageBRequiredGateAgent(agent: string): string {
    return agent === 'reviewer' || agent === 'test_engineer' ? 'coder' : agent;
}
```

This function is used in `recordAgentDispatch` (line 97) but the mapping seems inverted: for reviewer/test_engineer it maps to 'coder', but for other agents (explorer, sme, critic) it passes them through. The purpose is unclear — why would reviewer/test_engineer evidence need to record 'coder' as the required gate agent?

**Impact:** Low — the gate evidence file records what agents were dispatched. If reviewer/test_engineer is mapped to 'coder', the evidence may not accurately reflect which agent actually ran the gate.

**Verification:** Looking at `recordGateEvidence` usage — the `gate` parameter becomes the key in the `gates` object. So if we pass 'coder' for a reviewer gate, the evidence will show `gates.coder` instead of `gates.reviewer`. This could cause issues with `hasPassedAllGates` which checks `evidence.gates[gate] != null` for each `required_gates` entry.

**Verdict:** POTENTIAL BUG — if `required_gates` includes 'reviewer' and the Stage B ingestion writes evidence with key 'coder', `hasPassedAllGates` will return `false`.

However, looking more carefully: `recordGateEvidence` at line 100-106 is called with `args.record.normalizedAgent` (the actual agent), NOT `stageBRequiredGateAgent`. The `stageBRequiredGateAgent` is only used in `recordAgentDispatch` at line 94-98 for **agent dispatch recording**, not gate evidence recording. So this is intentional — it records the dispatch type differently from the gate evidence.

**Revised: NOT a bug.** The two functions serve different purposes.

---

#### M2: `review-receipt-collector.ts` — Verdict line pattern uses `/gim` flags

**File:** `src/hooks/review-receipt-collector.ts:117-118`

```typescript
const VERDICT_LINE_PATTERN =
    /^\s*(?:\*\*)?VERDICT(?:\*\*)?\s*:\s*(APPROVED|REJECTED)\s*$/gim;
```

The `/g` flag enables global matching, which is needed for the `matchAll` loop. However, the `/m` flag (multiline) is redundant because `^` and `$` already match start/end of string in this context. The `/i` flag (case-insensitive) is needed for matching `APPROVED`/`approved`.

**Impact:** Negligible — no functional difference.

---

#### M3: `dispatch-lanes.ts` — `ingestion_error` added to failed counts

**File:** `src/tools/dispatch-lanes.ts:970-972`

```typescript
const failed = records.filter(
    (record) =>
        record.status === 'error' || record.status === 'ingestion_error',
);
```

This correctly groups `ingestion_error` with `error` for the `failed` count. However, the `recordToLaneResult` function (line 1002-1009) maps `ingestion_error` to `'failed'` status in the result object, which is consistent.

**Verdict:** CORRECT — this is the expected behavior.

---

### 🟢 LOW

#### L1: `completion-observer.ts` — Logger consistency

**File:** `src/background/completion-observer.ts:68,118,155`

The observer uses `logger.log` for informational messages and `logger.warn` for warnings. This is consistent with the existing pattern in other background modules.

**Verdict:** No issue.

---

#### L2: `pending-delegations.ts` — Schema v1 → v2 upgrade

**File:** `src/background/pending-delegations.ts:371-372`

```typescript
schemaVersion:
    current.schemaVersion === 1 ? 2 : current.schemaVersion,
```

When transitioning from v1 to v2, the schema version is upgraded. However, v2 records include `batchId`, `laneId`, and `mode` fields that v1 records don't have. The schema allows these to be optional, so this is correct.

**Verdict:** No issue.

---

#### L3: `recovery-guide.md` — Documentation completeness

**File:** `docs/troubleshooting/recovery-guide.md`

The new section 9 ("Recovering from Background Stage B Gate Batches") provides clear guidance for operators. It covers all four terminal states: `pending/running`, `consumed`, `stale`, and `error`/`ingestion_error`.

**Verdict:** Well-documented.

---

## Validation Provenance

### Tests
- **`tests/unit/background/completion-observer.test.ts`**: 16 tests, 0 failures, 48 expect() calls
- **`tests/unit/hooks/review-receipt.test.ts`**: 46 tests, 0 failures, 98 expect() calls
- **Background test suite**: 600 tests, 1 failure (pre-existing `event-bus.test.ts` global instance test — unrelated to this PR, test isolation issue with shared global state)

### Build
- Dependencies installed via `bun install` (zod@4.3.6, etc.)
- No build errors reported

### Security Checks
- **Subprocess safety**: `workspace-snapshot.ts` uses `spawnSync` with proper `timeout`, `maxBuffer`, and `stdio` settings (Invariant 3)
- **Path containment**: `pending-delegations.ts` uses `validateSwarmPath` for all file operations (Invariant 4)
- **Evidence lock**: All append operations use `withEvidenceLock` for concurrency safety
- **Zod validation**: All JSON records are validated with strict schemas before persistence

---

## Disproved Candidates

The following initial concerns were investigated and **disproved**:

1. **H1 (DI seam)**: `runGit` already uses `_internals.spawnSync` — the DI seam is present and correct.
2. **H2 (stale + error)**: Error completions don't advance gates, so skipping stale validation is correct.
3. **M1 (gate agent mapping)**: `stageBRequiredGateAgent` is used for dispatch recording, not gate evidence — two different purposes.

---

## Overall Verdict

**APPROVED — with one advisory note**

### Advisory: Consider adding a log message when `ingestion_error` is written

When `ingestBackgroundStageBCompletion` fails and returns `{ ok: false, consumed: false, reason: 'stage-b ingestion failed: ...' }`, the `completion-observer.ts` does NOT record an `ingestion_error` status — it only logs a warning and returns. The `ingestion_error` status is written in `appendDelegationTransition` when the caller explicitly requests it.

Looking at `completion-observer.ts:143-151`:
```typescript
if (!ingested.ok) {
    await appendDelegationTransition(directory, envelope.sessionId, {
        status: 'ingestion_error',
        result: terminal.result ?? result,
    });
    logger.warn(
        `[background] Stage B completion was not applied: ...`
    );
}
```

This is actually correct — the `ingestion_error` status IS written. However, the error message in `ingestBackgroundStageBCompletion` (line 131) already logs the failure, and the observer logs another warning. This is acceptable for visibility.

**No action required.**

---

### Summary
- **Code quality**: High — well-documented, follows existing patterns
- **Security**: No vulnerabilities identified — proper subprocess handling, path validation, and evidence locking
- **Test coverage**: Adequate for new functionality (16 completion-observer tests, 46 review-receipt tests)
- **Documentation**: Comprehensive — recovery guide updated with all terminal states
- **Backward compatibility**: Good — schema v1→v2 upgrade is handled, optional fields don't break v1 records

**Recommendation: Merge.**

---
*🐝 Posted by the opencode-swarm local swarm (hybrid local + remote models) via swarm-pr-review-oneoff.sh — review only, not an approval.*

zaxbysauce:
Closed out the latest PR comments on `d3667e25`.

Disposition:
- No live review threads exist on this PR; the current feedback surface is PR comments only.
- I did not make code changes for this round because the new claims are either disproved by the current branch, explicitly documented behavior, or low-value advisory notes that do not justify reopening this safety-class PR.

Findings checked against source:

1. `ingestion_error` has no automatic recovery mechanism
- Partial, but not a missing-wire defect in this PR.
- The operator path is already documented in `docs/troubleshooting/recovery-guide.md:188-192`: fix the local problem, replay the trusted completion if available, or re-dispatch the missing Stage B gate; do not count the gate until the row becomes `consumed`.
- I am not changing this in PR2 because it is a separate retry/reconciliation feature, not an unwired Stage B gate integration path.

2. Empty reviewer receipt when `prompt` is undefined
- Disproved.
- `src/background/stage-b-gates.ts:108-114` only attempts receipt collection for `reviewer`.
- The collector itself hard-bails on missing prompt or transcript at `src/hooks/review-receipt-collector.ts:200-206`:
  - non-reviewer target => `return null`
  - missing prompt/transcript => `return null`
- So an undefined prompt does not create an empty durable receipt.

3. Workspace snapshot symlink / `path.resolve()` false-positive claim
- Disproved as stated.
- `src/background/workspace-snapshot.ts:72-77` uses `path.resolve`, which is lexical path normalization, not symlink canonicalization.
- The comment�s reasoning depends on `path.resolve()` behaving like `realpath()`, which it does not.

4. Evidence corruption can leave `consumed` with invalid evidence because writes are not atomic
- Disproved.
- `src/gate-evidence.ts:169-209` and `218-247` write under `withTaskEvidenceLock(...)` and persist through `atomicWriteFile(...)`.
- The code already has the temp-file-plus-rename atomic write path the comment asked for.

5. Non-gate-bearing completions return `{ ok: true, consumed: false }`
- Intentional and already explained in code.
- `src/background/stage-b-gates.ts:75-80` explicitly documents that the trusted terminal completion is still settled durably in the ledger and simply has no Stage B evidence/state side effects.
- That is the correct behavior for advisory/background completions outside the gate-bearing path.

6. Session-scoping concern around `candidateSessions(...)`
- No active bug.
- `src/background/stage-b-gates.ts:144-146` only resolves the single parent session keyed by `parentSessionId`, and Stage B advancement applies only through that parent session.
- The theoretical �multiple sessions sharing the same parentSessionId� scenario is not how `swarmState.agentSessions` is keyed in the current implementation.

7. `ingestion_error` grouped with `failed` in lane collection
- Intentional display behavior, not a correctness defect.
- This does not affect Stage B evidence, receipts, workflow-state advancement, or stale rejection semantics.

8. `stageBRequiredGateAgent(...)` concern
- Disproved.
- `src/background/stage-b-gates.ts:93-106` uses `stageBRequiredGateAgent(...)` only for `recordAgentDispatch(...)` and uses the actual gate role for `recordGateEvidence(...)`.
- So required-gate expansion and actual gate evidence are intentionally distinct and correctly wired.

Advisory notes I am leaving alone in this PR:
- append-only ledger growth / O(n) scans
- UI wording/style nits
- possible future reconciliation automation for `ingestion_error`
- additional low-risk tests around symlinked paths or other speculative edge cases

Remote state at closeout:
- PR #1444 head remains `d3667e252817b84498748cd41584b03f5835c34c`
- `mergeable = MERGEABLE`
- `mergeStateStatus = CLEAN`
- current visible checks are green

PR #1455, opened 2026-06-20
feat(guardrails): add transparency suite (explain, decision log, diagnose sandbox)

### LINKED ISSUE #1274: [Feature] Guardrail transparency suite: dry-run "explain", unified decision log, and sandbox status in /swarm diagnose
From a direct architecture-level review pass, 2026-06-12. High-impact because unexpected guardrail blocks are the single biggest source of user friction in a gated agent system, and today the decision-making is a 5,107-line black box (`src/hooks/guardrails.ts`).

The building blocks already exist — this feature exposes them:
- A redacted shell-audit log is already written to `.swarm/session/shell-audit.jsonl` (`src/hooks/guardrails.ts:1331-1335, 3145`), but only for shell commands, and nothing surfaces it to users.
- Scope/authority/zone classification is all computed internally with no external "why".

### 1. Dry-run `/swarm guardrail explain` (highest value)
Let a user ask what the guardrails *would* do to a given command or write target under a given agent/scope, without executing anything:
```
/swarm guardrail explain --agent coder --scope src/ "rm -rf build && curl x | sh"
/swarm guardrail explain --agent coder --write src/index.ts
```
Output: the decision (allow/block), which rule fired, the resolved scope, and the shell-write categories detected. Turns "why did my agent get blocked?" from code-spelunking into one command. Reuses the existing detection functions directly — no new analysis logic.

### 2. Unified guardrail decision log
Extend the existing shell-audit log to record *all* guardrail decisions (file writes, scope violations, destructive-command blocks, sandbox wrap/skip), not just shell, with a `/swarm guardrail-log [--blocks-only]` reader. Gives an after-the-fact, queryable trail of every security-relevant decision in a run — valuable for trust, debugging, and post-incident review.

### 3. Sandbox status in `/swarm diagnose`
`src/commands/diagnose.ts` does not report whether OS-level sandboxing is actually active (bubblewrap on Linux, sandbox-exec on macOS, the Rust runner on Windows). Because the sandbox layer silently passes through when its executor is unavailable (see #1270 item 2), a user can believe they have OS isolation when they don't. Surface in `diagnose`: which executor was detected, whether it's available, and therefore whether commands are actually sandboxed. Pairs with #1270's sandbox-unavailable advisory.

## Why grouped

All three make the security layer *observable and predictable* rather than opaque. They share the same internal surfaces (detection functions, audit log, executor probe) and the same goal: let users trust and debug the guardrails instead of fighting them.

## Acceptance criteria

- [ ] `/swarm guardrail explain` returns decision + firing rule + resolved scope + detected categories for a command and for a write target, executing nothing
- [ ] Guardrail decision log captures non-shell decisions (writes, scope, destructive, sandbox) with a reader command
- [ ] `/swarm diagnose` reports detected sandbox executor and whether commands are actually being sandboxed
- [ ] No new hot-path cost (explain/log reader are on-demand commands)

zaxbysauce:
Closes #1274

## Summary

Adds the Guardrail Transparency Suite: a dry-run `/swarm guardrail explain`, a unified guardrail decision log with `/swarm guardrail-log`, and sandbox status in `/swarm diagnose`.

- `/swarm guardrail explain` reports allow/block, firing rule, resolved scope, and write categories without executing the command. It now fails closed on shell write parse errors and mirrors the real shell classifier for PowerShell/cmd/posix routing.
- Guardrail decision logging keeps legacy shell entries readable and adds typed records for file-write blocks, scope violations, destructive-command blocks, sandbox wraps, and sandbox skips. New advisory telemetry is fire-and-forget so audit I/O cannot delay enforcement.
- Audit persistence redacts embedded paths in `file_write.reason`, handles POSIX/macOS/Windows/UNC home-like paths, and rejects malformed typed entries before writing.
- `/swarm diagnose` uses a cached sandbox capability probe through a DI seam so sandbox status remains advisory and testable.
- Tests were hardened away from fixed sleeps toward bounded JSONL observers, and sandbox diagnose tests now use `_internals` for the PR-introduced sandbox probe/executor path.

## Invariant audit
- 1 (plugin init): not touched - no `src/index.ts` or init-path changes.
- 2 (runtime portability): touched - no new `bun:` imports or `Bun.*` runtime calls outside tests; verified with `bun run build` and `node --input-type=module -e "await import('./dist/index.js'); console.log('dist import OK')"`.
- 3 (subprocesses): not touched - no new subprocess execution.
- 4 (.swarm containment): touched - guardrail audit state remains under `.swarm/session/shell-audit.jsonl`; tests use temp project roots.
- 6 (test_runner safety): not touched - validation used shell commands, not broad `test_runner`.
- 7 (test writing): touched - `bun:test`; temp dirs use `os.tmpdir()`/`path.join()`/`realpathSync`; PR-introduced diagnose sandbox probe/executor mocks moved to `_internals`; capture tests use bounded observers instead of fixed sleeps.
- 9 (guardrails/retry): touched - guardrail parse-error explain now fails closed; new audit telemetry is fire-and-forget and does not change retry/circuit-breaker semantics.
- 11 (tool registration): touched by the original PR - command registry/export surfaces already covered by existing PR tests; feedback patch did not add tools.
- 12 (release/cache): touched - updated `docs/releases/pending/guardrail-transparency-suite.md`; no release-owned version files edited.

## Test plan
- [x] `bun --smol test tests/unit/services/guardrail-explain-service.test.ts tests/unit/hooks/audit-log.test.ts tests/unit/hooks/guardrail-capture.test.ts tests/unit/services/guardrail-log-service.test.ts tests/unit/services/diagnose-sandbox.test.ts --timeout 120000` - 103 pass, 1 todo, 0 fail
- [x] `bun run typecheck`
- [x] `bunx biome ci .` - exit 0; reports one unrelated pre-existing info in `src/session/session-start-store.ts`
- [x] `git diff --check`
- [x] `bun run build`
- [x] `node --input-type=module -e "await import('./dist/index.js'); console.log('dist import OK')"`
- [ ] CI cross-platform suite

## Known caveats
- `/swarm guardrail explain` still duplicates part of the destructive-command evaluator because the real evaluator is nested in the hook. This PR adds parity regressions for the feedback cases; extracting a shared evaluator remains the follow-up direction.
- Redaction is best-effort and covers the standard POSIX, macOS, Windows profile, UNC, and embedded-path cases added here. It is not a full canonical secret scrubber for arbitrary custom path layouts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1080,7 +1080,9 @@ Control how tool outputs are summarized for LLM context.
 | `/swarm agents` | Registered agents with models and permissions |
 | `/swarm history` | Completed phases with status |
 | `/swarm config` | Current resolved configuration |
-| `/swarm diagnose` | Health check for `.swarm/` files and config |
+| `/swarm diagnose` | Health check for `.swarm/` files, config, and sandbox executor status |
+| `/swarm guardrail explain [--agent <role>] [--scope <path>] [--write <path>...] [--] <command>` | Dry-run guardrail decisions — report what would be allowed/blocked without executing anything |
+| `/swarm guardrail-log [--blocks-only]` | Read and print the guardrail decision log most-recent-first; `--blocks-only` shows only block decisions |
 | `/swarm export` | Export plan and context as portable JSON |
 | `/swarm evidence [task]` | Evidence bundles for a task or all tasks |
 | `/swarm archive [--dry-run]` | Archive old evidence with retention policy |
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -71,6 +71,8 @@ A 3-line learning summary is automatically injected into the curator phase diges
 
 Run a health check on `.swarm/` files, plan structure, and evidence completeness. Reports missing files, schema mismatches, and recovery steps.
 
+A **Sandbox** health-check line is also reported, showing the detected executor mechanism (bubblewrap / sandbox-exec / windows-runner / none), availability, and whether commands are actually being sandboxed (sandboxing / silent pass-through / none). This is advisory only — absence of a sandbox executor never causes a hard failure.
+
 ### `/swarm history`
 
 Show completed phases with status icons.
@@ -82,6 +84,60 @@ Show completed phases with status icons.
 ### `/swarm agents`
 
 List all registered agents with their model, temperature, read-only status, and guardrail profile.
+
+---
+
+## Guardrails
+
+### `/swarm guardrail explain [--agent <role>] [--scope <path>] [--write <path>...] [--] <command>`
+
+Dry-run guardrail decisions for a shell command or write target — reports what the guardrail system **would** do without executing anything. Agent-callable via `swarm_command`.
+
+**Shell mode** (default — pass a command string after any flags):
+
+```text
+/swarm guardrail explain rm -rf node_modules/
+/swarm guardrail explain --agent reviewer git push --force origin main
+```
+
+Returns: decision (`allow`/`block`), firing rule, resolved scope, and detected write categories.
+
+**Write mode** (`--write`, repeatable — inspect individual file/directory targets):
+
+```text
+/swarm guardrail explain --write src/hooks/guardrails.ts --write .swarm/plan.json
+```
+
+Returns per-target: decision, firing rule, resolved scope, and zone classification.
+
+**Flags:**
+
+| Flag | Effect |
+|------|--------|
+| `--agent <role>` | Simulate decisions as if issued by a different agent role (e.g., `reviewer`, `test_engineer`) |
+| `--scope <path>` | Simulate decisions scoped to a specific working directory |
+| `--write <path>` | Inspect a write target instead of a shell command (repeatable for multiple targets) |
+| `--` | Explicit flag terminator — required when `<command>` starts with `--` |
+
+Output is fully advisory and redacted. No side effects, no writes, no process execution.
+
+### `/swarm guardrail-log [--blocks-only]`
+
+Read and print the unified guardrail decision log (`.swarm/session/shell-audit.jsonl`) most-recent-first. Agent-callable via `swarm_command`.
+
+```text
+/swarm guardrail-log
+/swarm guardrail-log --blocks-only
+```
+
+**`--blocks-only`** limits output to block decisions only (`file_write`, `scope_violation`, `destructive_block`). Legacy shell command entries and sandbox wrap/skip entries are excluded.
+
+**Output characteristics:**
+
+- Entries sorted most-recent-first
+- Commands and paths are redacted
+- Missing log file → friendly message: "No guardrail decisions recorded yet."
+- On-demand only — no hot-path cost; reads the log only when invoked
 
 ---
 
--- a/docs/releases/pending/guardrail-transparency-suite.md
+++ b/docs/releases/pending/guardrail-transparency-suite.md
@@ -1,0 +1,122 @@
+# Guardrail Transparency Suite — Complete
+
+## What changed
+
+The guardrail transparency suite (issue #1274) is now complete. Phase 1 delivered the foundation; Phase 2 completed the observability stack.
+
+---
+
+## Phase 1 — Foundation
+
+### `/swarm guardrail explain` — dry-run guardrail decisions before execution
+
+A diagnostics command that reports what the guardrail system **would** do to a given shell command or write target — without executing anything. Available to agents via `swarm_command`.
+
+**Shell mode** (default — pass a command string):
+
+```
+/swarm guardrail explain rm -rf node_modules/
+```
+
+Returns: decision (`allow`/`block`), the firing rule, resolved scope, and detected write categories.
+
+**Write mode** (`--write`, repeatable — inspect individual file/directory targets):
+
+```
+/swarm guardrail explain --write src/hooks/guardrails.ts --write .swarm/plan.json
+```
+
+Returns: per-target decision, firing rule, resolved scope, and zone classification for each path.
+
+**Simulation flags:**
+
+- `--agent <role>` — simulate guardrail decisions as if issued by a different agent role (e.g., `reviewer`, `test_engineer`)
+- `--scope <path>` — simulate decisions scoped to a specific working directory
+
+Output is advisory and fully redacted — no side effects, no writes, no process execution.
+
+### `/swarm diagnose` — Sandbox status
+
+`/swarm diagnose` includes a **Sandbox** health-check line reporting:
+
+- **Executor mechanism** detected (bubblewrap / sandbox-exec / windows-runner / none)
+- **Availability** of the sandbox executor
+- **Whether commands are actually being sandboxed** (sandboxing / silent pass-through / none)
+
+This is advisory only — absence of a sandbox executor never causes a hard failure.
+
+### Unified guardrail decision-log schema (Phase 1 foundation)
+
+The shell-audit log at `.swarm/session/shell-audit.jsonl` uses an additive type-discriminated JSONL schema. Five decision types are written to the same stream alongside existing shell entries:
+
+- `file_write`
+- `scope_violation`
+- `destructive_block`
+- `sandbox_wrap`
+- `sandbox_skip`
+
+---
+
+## Phase 2 — Decision-Log Reader and Full Capture
+
+### `/swarm guardrail-log [--blocks-only]`
+
+A new diagnostics command that reads the unified guardrail decision log (`.swarm/session/shell-audit.jsonl`) and prints entries most-recent-first. Agent-callable via `swarm_command`.
+
+```
+/swarm guardrail-log
+/swarm guardrail-log --blocks-only
+```
+
+**`--blocks-only`** filters to show only block decisions (`file_write`, `scope_violation`, `destructive_block`). Legacy shell command entries and sandbox wrap/skip entries are excluded.
+
+**Output characteristics:**
+
+- Entries are sorted most-recent-first
+- Commands and paths are redacted in output
+- If no log file exists yet, prints a friendly message: "No guardrail decisions recorded yet."
+- On-demand only — no hot-path cost; reads the log file only when the command is invoked
+
+**Example output:**
+
+```
+=== Guardrail Decision Log ===
+2026-06-20T10:42:01.000Z  destructive_block  rm -rf node_modules/
+2026-06-20T10:41:55.000Z  file_write         src/hooks/guardrails.ts
+2026-06-20T10:41:48.000Z  scope_violation    ./src/secrets.json
+
+No guardrail decisions recorded yet.
+```
+
+### Unified decision-log capture (internal)
+
+The guardrail hook now records **all** decision types — file-write blocks, scope violations, destructive-command blocks, sandbox wraps, and sandbox skips — to the unified log. Each entry is redacted before writing. Blocking behavior is unchanged; the capture is purely additive observability.
+
+This completes the unified guardrail decision log capability from issue #1274: the schema was introduced in Phase 1, and the full capture wiring + reader command landed in Phase 2.
+
+## Complete suite at a glance
+
+| Capability | Status | Notes |
+|---|---|---|
+| `guardrail explain` dry-run | ✅ Phase 1 | Shell + write modes, simulation flags |
+| Sandbox status in `diagnose` | ✅ Phase 1 | Executor, availability, enforcement |
+| Unified log schema | ✅ Phase 1 | Five new decision types in JSONL |
+| `guardrail-log` reader | ✅ Phase 2 | Most-recent-first, `--blocks-only`, redacted |
+| Full capture wiring | ✅ Phase 2 | All decision types now logged |
+
+## Why
+
+Guardrail decisions were previously opaque — agents and users had no reliable way to predict whether a given command would be allowed, blocked, or sandboxed. The `explain` command makes guardrail behavior observable and debuggable *before* a command runs, eliminating guesswork.
+
+The `guardrail-log` command and full capture wiring make the *history* of decisions observable *after* the fact, completing the observability loop: before (`explain`), during (logging), and after (`guardrail-log`).
+
+The Sandbox line in `diagnose` closes a visibility gap: users on Linux (bubblewrap), macOS (sandbox-exec), or Windows (windows-runner) can confirm whether their runtime is actually enforcing sandbox boundaries.
+
+## Migration
+
+No breaking changes. All changes are additive:
+
+- `/swarm guardrail explain` — new command, safe to use immediately
+- `/swarm diagnose` — gains a new Sandbox line; existing output unchanged
+- `/swarm guardrail-log` — new command, reads existing log file
+- The JSONL schema is additive only — existing log readers are unaffected
--- a/docs/releases/pending/guardrail-transparency-suite.md
+++ b/docs/releases/pending/guardrail-transparency-suite.md
@@ -120,3 +120,13 @@ No breaking changes. All changes are additive:
 - `/swarm diagnose` — gains a new Sandbox line; existing output unchanged
 - `/swarm guardrail-log` — new command, reads existing log file
 - The JSONL schema is additive only — existing log readers are unaffected
+
+## Known limitations
+
+- `guardrail explain` still mirrors the destructive-command pattern table locally
+  because the real guardrail evaluator is nested in the hot-path hook. The shared
+  `dc*` helpers are reused and parity regressions are covered by tests; future
+  cleanup can extract a shared top-level evaluator.
+- Audit-log path redaction is best-effort. It redacts standard POSIX/macOS home
+  paths, Windows profile paths, and UNC server prefixes, but domain-specific
+  secrets embedded in arbitrary path segments remain caller responsibility.
--- a/scripts/mock-allowlist.txt
+++ b/scripts/mock-allowlist.txt
@@ -9,7 +9,7 @@
 # To add a NEW mock target: append it here with a comment explaining why.
 # Prefer _internals DI seam for new code — mock.module is a legacy pattern.
 #
-# Last updated: 2026-06-17
+# Last updated: 2026-06-20
 
 # --- Node builtins ---
 node:child_process
@@ -25,6 +25,7 @@ src/build/discovery
 src/config
 src/config/index
 src/config/index.ts
+src/config/loader
 src/config/schema
 src/config/schema.ts
 src/db/qa-gate-profile
@@ -57,7 +58,10 @@ src/plan/checkpoint
 src/plan/ledger
 src/plan/manager
 src/proper-lockfile
+src/sandbox/capability-probe
+src/sandbox/executor
 src/sandbox/linux/bubblewrap-executor
+src/sdd/effective-spec
 src/sast/rules/index
 src/sast/semgrep
 src/services/evidence-summary-service
--- a/src/commands/guardrail-explain.ts
+++ b/src/commands/guardrail-explain.ts
@@ -1,0 +1,2 @@
+// Delegate to service layer - thin adapter for backward compatibility
+export { handleGuardrailExplain } from '../services/guardrail-explain-service';
--- a/src/commands/guardrail-log.ts
+++ b/src/commands/guardrail-log.ts
@@ -1,0 +1,1 @@
+export { handleGuardrailLog } from '../services/guardrail-log-service';
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -59,6 +59,8 @@ export {
 } from './evidence';
 export { handleExportCommand } from './export';
 export { handleFullAutoCommand } from './full-auto';
+export { handleGuardrailExplain } from './guardrail-explain';
+export { handleGuardrailLog } from './guardrail-log';
 export { handleHandoffCommand } from './handoff';
 export { handleHistoryCommand } from './history';
 export {
--- a/src/commands/registry.ts
+++ b/src/commands/registry.ts
@@ -28,6 +28,8 @@ import {
 } from './evidence.js';
 import { handleExportCommand } from './export.js';
 import { handleFullAutoCommand } from './full-auto.js';
+import { handleGuardrailExplain } from './guardrail-explain.js';
+import { handleGuardrailLog } from './guardrail-log.js';
 import { handleHandoffCommand } from './handoff.js';
 import { handleHistoryCommand } from './history.js';
 import { handleIssueCommand } from './issue.js';
@@ -432,6 +434,22 @@ export const COMMAND_REGISTRY = {
 		category: 'diagnostics',
 		aliasOf: 'diagnose',
 		deprecated: true,
+	},
+	'guardrail explain': {
+		handler: (ctx) => handleGuardrailExplain(ctx.directory, ctx.args),
+		description:
+			'Dry-run: show what the guardrails would do to a command or write target (executes nothing)',
+		category: 'diagnostics',
+		toolPolicy: 'agent',
+		toolNoArgs: false,
+	},
+	'guardrail-log': {
+		handler: (ctx) => handleGuardrailLog(ctx.directory, ctx.args),
+		description:
+			'Read the guardrail decision log (use --blocks-only for blocks)',
+		category: 'diagnostics',
+		toolPolicy: 'agent',
+		toolNoArgs: false,
 	},
 	preflight: {
 		handler: (ctx) => handlePreflightCommand(ctx.directory, ctx.args),
--- a/src/hooks/guardrails/audit-log.ts
+++ b/src/hooks/guardrails/audit-log.ts
@@ -264,6 +264,18 @@ function validateEntry(entry: unknown): entry is GuardrailDecisionEntry {
 export function redactPath(filePath: string): string {
 	if (typeof filePath !== 'string' || filePath.length === 0) return filePath;
 
+	// POSIX home: /home/<name>/... -> ~/...
+	const rawPosixHomeMatch = filePath.match(/^(\/home\/[^/]+)(\/.*)$/);
+	if (rawPosixHomeMatch) {
+		return `~${rawPosixHomeMatch[2]}`;
+	}
+
+	// macOS home: /Users/<name>/... -> ~/... (parity with redactShellCommand)
+	const rawMacHomeMatch = filePath.match(/^(\/Users\/[^/]+)(\/.*)$/);
+	if (rawMacHomeMatch) {
+		return `~${rawMacHomeMatch[2]}`;
+	}
+
 	const normalized = path.normalize(filePath);
 
 	// Windows user profile: C:\Users\<name>\... -> ~\<name>\...
@@ -274,19 +286,27 @@ export function redactPath(filePath: string): string {
 		return `~\\${windowsHomeMatch[2]}`;
 	}
 
-	// POSIX home: /home/<name>/... -> ~/<name>/...
-	const posixHomeMatch = normalized.match(/^(\/home\/[^/]+)(\/.*)$/);
-	if (posixHomeMatch) {
-		return `~${posixHomeMatch[2]}`;
-	}
-
-	// macOS home: /Users/<name>/... -> ~/<name>/... (parity with redactShellCommand)
-	const macHomeMatch = normalized.match(/^(\/Users\/[^/]+)(\/.*)$/);
-	if (macHomeMatch) {
-		return `~${macHomeMatch[2]}`;
+	// Windows UNC profile/share path: \\server\share\user\... -> ~\share\user\...
+	const uncHomeMatch = normalized.match(
+		/^\\\\[^\\]+\\([^\\]+)\\([^\\]+)(\\.*)$/,
+	);
+	if (uncHomeMatch) {
+		return `~\\${uncHomeMatch[1]}\\${uncHomeMatch[2]}${uncHomeMatch[3]}`;
 	}
 
 	return normalized;
+}
+
+function redactEmbeddedPaths(text: string): string {
+	return text
+		.replace(/\/home\/[^/\s]+(\/[^\s"'`)]*)?/g, (match) => redactPath(match))
+		.replace(/\/Users\/[^/\s]+(\/[^\s"'`)]*)?/g, (match) => redactPath(match))
+		.replace(/[A-Za-z]:\\Users\\[^\\\s]+(\\[^\s"'`)]*)?/g, (match) =>
+			redactPath(match),
+		)
+		.replace(/\\\\[^\\\s]+\\[^\\\s]+\\[^\\\s]+(\\[^\s"'`)]*)?/g, (match) =>
+			redactPath(match),
+		);
 }
 
 // ---------------------------------------------------------------------------
@@ -342,6 +362,7 @@ export async function appendGuardrailDecision(
 			line = `${JSON.stringify({
 				...entry,
 				path: redactPath(entry.path),
+				reason: redactEmbeddedPaths(entry.reason),
 				resolvedScope: redactPath(entry.resolvedScope),
 			})}\n`;
 			break;
--- a/src/hooks/guardrails/audit-log.ts
+++ b/src/hooks/guardrails/audit-log.ts
@@ -1,0 +1,368 @@
+/**
+ * Unified Guardrail Decision Audit Log
+ *
+ * Additive JSONL schema for guardrail decisions. Each append writes one
+ * validated JSON line to the configured audit path under `.swarm/session/`.
+ *
+ * The existing shell audit entry shape is preserved byte-for-byte when
+ * `type: 'shell'` is used (fields: ts, sessionID, agent, tool, command).
+ *
+ * Path redaction: there is no cross-cutting path-redaction helper in this
+ * module today. Callers MAY pass a pre-redacted path via `redactPath(...)`
+ * before constructing the entry; the module itself does not mutate path
+ * strings beyond normalizing separators. Future hardening: add a shared
+ * path-redaction utility and call it here.
+ */
+
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+import { log } from '../../utils/logger';
+import { redactShellCommand } from './helpers';
+
+// DI seam for testability — tests mock these function references directly (function mocks
+// restore cleanly in afterEach) instead of mock.module('node:fs/promises'), which leaks
+// across test files in Bun's shared test-runner process (AGENTS.md invariant 7).
+export const _internals = {
+	mkdir: fs.mkdir,
+	appendFile: fs.appendFile,
+};
+
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
+
+/**
+ * Discriminated union of guardrail decision types.
+ *
+ * Invariants:
+ * - `ts` is an ISO-8601 datetime string.
+ * - `sessionID` is the swarm session identifier.
+ * - `agent` is the role string (e.g. "architect", "coder").
+ * - `tool` is the original tool name as invoked.
+ */
+export type GuardrailDecisionType =
+	| 'shell'
+	| 'file_write'
+	| 'scope_violation'
+	| 'destructive_block'
+	| 'sandbox_wrap'
+	| 'sandbox_skip';
+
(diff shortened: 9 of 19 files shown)


zaxbysauce:
## 🤖 Tri-Model PR Review
**Pipeline:** Mellum2-12B (code) + gemma-4-12B-agentic (broad) → Qwen3.6-35B-APEX (verify + blind-spot)
**Commit reviewed:** `d05abf3fe4b8`

---

# PR #1455 — Final Review (Tri-Model Verification)

## 🔁 Cross-model verification

| Draft finding | Verdict | Reason |
|---|---|---|
| A: Dry-run explain for shell/write | **CONFIRMED** | `guardrail-explain-service.ts` implements both modes with proper arg parsing |
| A: Unified decision log schema | **CONFIRMED** | `audit-log.ts` defines discriminated union with 6 types |
| A: `guardrail-log` with `--blocks-only` | **CONFIRMED** | `guardrail-log-service.ts:137` filters by `BLOCK_TYPES` set |
| A: Sandbox status in `diagnose` | **CONFIRMED** | `diagnose-service.ts:803-849` adds `getSandboxStatus()` |
| A: `redactShellCommand` home-pass | **CONFIRMED** | `helpers.ts:156-158` adds three platform-specific regexes |
| A: `_internals` DI seam | **CONFIRMED** | `audit-log.ts:28-31` exports `mkdir`/`appendFile` refs |
| B: Redundant `redactPath` calls | **CONFIRMED** | `guardrail-explain-service.ts:438,445,458,463,470` — `checkFileAuthority` already redacts `decision.reason` |
| B: `==` for null check | **CONFIRMED** | `guardrail-log-service.ts:107` uses `entry.type == null` |
| B: `typeof` checks (many) | **REFUTED** | These are type guards for JSON parsing, not null checks — they're correct usage |
| B: Redundant `typeof` in test | **REFUTED** | `guardrail-capture.test.ts:247` — `typeof parsed.ts` is a sanity assertion, not redundant |

---

## 🚨 Confirmed Findings

### [LOW / NIT] Redundant `redactPath` calls in `guardrail-explain-service.ts`
- **Location**: `src/services/guardrail-explain-service.ts:438, 445, 458, 463, 470`
- **Why it matters**: `redactPath` is already called inside `checkFileAuthority` (which redacts `decision.reason` and `resolvedScope`); the extra calls are harmless but clutter the code.
- **Evidence**: Lines 438-470 call `redactPath(r.path)`, `redactPath(r.firingRule)`, `redactPath(r.resolvedScope)` after `checkFileAuthority` already returned redacted values.
- **Fix direction**: Remove redundant `redactPath` calls in the write-target result builder.

### [LOW / NIT] `guardrail-log-service.ts` uses `==` for `null` check
- **Location**: `src/services/guardrail-log-service.ts:107`
- **Why it matters**: `entry.type == null` is a loose equality check; use `===` for consistency with the rest of the codebase.
- **Evidence**: `if (entry.type == null) return false;`
- **Fix direction**: Replace with `entry.type === null`.

### [MEDIUM] `guardrail-log-service.ts` missing `isPlainObject` guard for legacy shell entries
- **Location**: `src/services/guardrail-log-service.ts:114-142`
- **Why it matters**: Legacy shell entries (5-field shape, no `type` field) pass the `isPlainObject` check but then fail the `typeof parsed.type === 'string'` check at line 134, causing them to be silently skipped. This means **legacy shell entries are invisible to `guardrail-log`** unless `--blocks-only` is not used AND the entry happens to have a `type` field.
- **Evidence**: Lines 114-142 iterate over parsed JSON; if `typeof parsed.type !== 'string'` the entry is skipped entirely. Legacy shell entries have no `type` field, so they are never added to `entries`.
- **Fix direction**: Add a fallback: if `parsed.type` is not a string but `parsed.ts`, `parsed.sessionID`, `parsed.agent`, `parsed.tool`, and `parsed.command` are all strings, treat it as a legacy shell entry.

### [MEDIUM] `tool-before.ts` sandbox_skip capture uses `void` — audit failure silently lost
- **Location**: `src/hooks/guardrails/tool-before.ts:915-927`
- **Why it matters**: The `sandbox_skip` decision is captured with `void appendGuardrailDecision(...)`. If the audit write fails, the decision is lost. This is inconsistent with the `destructive_block` capture which also uses `void` but at least has a `try/catch` around the `checkDestructiveCommand` call.
- **Evidence**: Line 915: `void appendGuardrailDecision(...)` — no error handling.
- **Fix direction**: Wrap in `try/catch` with logging, or accept that this is consistent with the existing pattern (all captures use `void`).

### [LOW / NIT] `diagnose-service.ts` `getSandboxStatus` re-probes on every call
- **Location**: `src/services/diagnose-service.ts:810-811`
- **Why it matters**: The comment says "Never re-probes" but the code creates a new `SandboxCapabilityProbe()` and calls `detect()` every time. If the probe is expensive, this could be a performance issue.
- **Evidence**: Line 810: `const capability = await new SandboxCapabilityProbe().detect();`
- **Fix direction**: Use a cached probe if one exists, or document that re-probing is intentional.

---

## 🔍 Blind-spot pass — findings both reviewers missed

### [MEDIUM] `guardrail-log-service.ts` does not handle JSONL corruption gracefully
- **Location**: `src/services/guardrail-log-service.ts:108-142`
- **Why it matters**: If the JSONL file contains a malformed line (e.g., truncated JSON from a crash), the `JSON.parse` catch at line 110 skips the line. However, if the file is partially written (e.g., a line is being appended when the process crashes), the file could contain a partial line that causes `JSON.parse` to throw. The current code handles this, but it silently skips the line without any indication to the user.
- **Evidence**: Line 110: `try { parsed = JSON.parse(line); } catch { continue; }`
- **Fix direction**: Consider logging a warning when a line is skipped, or adding a `--verbose` flag.

### [LOW / NIT] `guardrail-explain-service.ts` `evaluateDestructiveDecision` does not check `block_destructive_commands` config
- **Location**: `src/services/guardrail-explain-service.ts:138-140`
- **Why it matters**: The comment says "Assumes block_destructive_commands is TRUE (explain predicts default guardrail behavior)." However, if the user has disabled destructive command blocking in their config, the explain command will report a block that would not actually occur. This is a **shipped-vs-claimed mismatch**.
- **Evidence**: Line 138-140: `// Assumes block_destructive_commands is TRUE (explain predicts default guardrail behavior).`
- **Fix direction**: Either accept the config value or document that explain always reports default behavior.

### [LOW / NIT] `audit-log.ts` `redactPath` does not handle Windows UNC paths
- **Location**: `src/hooks/guardrails/audit-log.ts:270-290`
- **Why it matters**: Windows UNC paths like `\\server\share\user\file.txt` are not handled by the current regexes. This could lead to path leakage on Windows network shares.
- **Evidence**: Lines 270-290 only handle `C:\Users\...` and POSIX `/home/...` / `/Users/...` patterns.
- **Fix direction**: Add a regex for UNC paths: `^\\\\[^\\]+\\[^\\]+\\([^\\]+)(\\.*)$`.

---

## 📋 Merge recommendation

**APPROVE_WITH_FIXES**

The PR delivers a well-structured transparency suite with comprehensive tests. The main concerns are:
1. Legacy shell entries being invisible to `guardrail-log` (MEDIUM)
2. `evaluateDestructiveDecision` not respecting config (LOW/MEDIUM)
3. Redundant `redactPath` calls (LOW)

These are non-blocking issues that can be addressed in follow-up PRs or quickly fixed before merge.

---
<sub>🔒 Reviewed locally by three on-prem models (two reviewers + a verifier) for blind-spot coverage. Findings are advisory — verify before acting.</sub>

zaxbysauce:
## 🤖 Tri-Model PR Review
**Pipeline:** Mellum2-12B (code) + gemma-4-12B-agentic (broad) → Qwen3.6-35B-APEX (verify + blind-spot)
**Commit reviewed:** `617c008fe8ca`

---

# 🔍 Final Consolidated Review — PR #1455: feat(guardrails): add transparency suite

## Summary

This PR adds a guardrail transparency suite with three coordinated capabilities: a dry-run `/swarm guardrail explain` command, a unified guardrail decision log with `/swarm guardrail-log`, and sandbox status in `/swarm diagnose`. The implementation reuses existing detection logic and adds audit-log capture wiring into the toolBefore hook.

---

## 🔁 Cross-model verification

### Reviewer A findings:
- **CONFIRMED**: `redactPath` is a weak heuristic — only handles home-dir patterns, can be bypassed by non-standard path encodings
- **CONFIRMED**: Redundant `redactPath` calls on already-redacted fields in `audit-log.ts`
- **REFUTED**: `guardrail-log` uses `typeof` for type checks — this is intentional defensive coding for JSONL parsing where types are unknown; not a defect

### Reviewer B findings:
- **CONFIRMED**: `redactPath` is a weak heuristic (same as A)
- **REFUTED**: All `typeof` type-check nitpicks — these are intentional for JSONL parsing where types are unknown; not a defect
- **REFUTED**: `Object.keys(parsed).length === 5` check — this is intentional for legacy shell entry validation; not a defect

---

## 🚨 Confirmed Findings

### [CRITICAL] `patchAgentName` null-check behavioral change in `tool-before.ts`

**Location**: `src/hooks/guardrails/tool-before.ts:1891`

**Before**:
```typescript
const patchAgentName = swarmState.activeAgent.get(input.sessionID);
if (!patchAgentName) {
```

**After**:
```typescript
const patchAgentName = swarmState.activeAgent.get(input.sessionID) ?? 'unknown';
if (!swarmState.activeAgent.has(input.sessionID)) {
```

**Why it matters**: This is a behavioral change. The old code would throw if `activeAgent.get(sessionID)` returned `null` or `undefined`. The new code only throws if the key doesn't exist. If the key exists but has a value of `null` or `undefined`, the old code would throw but the new code wouldn't. This could allow write operations to proceed without proper agent validation.

**Fix direction**: Either restore the original check or ensure that `activeAgent.get()` never returns `null`/`undefined` when the key exists.

---

### [MEDIUM] `applySandboxExecution` signature change — potential breaking change

**Location**: `src/hooks/guardrails/tool-before.ts:905-945`

The function signature was changed to accept additional parameters for audit logging:
```typescript
async function applySandboxExecution(
    sessionID: string,
    tool: string,
    args: unknown,
    agent: string,
    command: string,
    auditPath: string,
    auditEnabled: boolean,
): Promise<void>
```

**Why it matters**: If `applySandboxExecution` is exported and called from anywhere else in the codebase, this is a breaking change. The diff shows it's called internally, but the function may be exported for testing or other purposes.

**Fix direction**: Verify that `applySandboxExecution` is not exported or used elsewhere. If it is, either keep the old signature and add a wrapper, or update all callers.

---

### [MEDIUM] `redactPath` is a weak heuristic — security concern

**Location**: `src/hooks/guardrails/audit-log.ts:135-165`

The regex-based path redaction only handles:
- Windows: `C:\Users\<name>\...`
- POSIX: `/home/<name>/...`
- macOS: `/Users/<name>/...`

**Why it matters**: The function can be bypassed by:
- Non-standard path encodings (e.g., `/home/user/../secret`)
- Network paths (e.g., `\\server\share\user\...`)
- Custom home directory locations (e.g., `/opt/home/user/...`)
- Any sensitive information in path segments that don't match the patterns

**Fix direction**: Use a more robust path-canonicalization step before redaction and add a warning that the redaction is best-effort.

---

### [MEDIUM] `evaluateDestructiveDecision` duplication — drift risk

**Location**: `src/services/guardrail-explain-service.ts:100-500`

The `evaluateDestructiveDecision` function mirrors `checkDestructiveCommand` from `tool-before.ts` but is a separate implementation. The TODO comment acknowledges this:
```typescript
// TODO(#1274): checkDestructiveCommand is a nested closure in tool-before.ts:182
// (captures cfg.block_destructive_commands, effectiveDirectory, resolveDeclaredScope) and
// cannot be exported without refactoring the guardrail hot path.
```

**Why it matters**: If the real `checkDestructiveCommand` is updated, this dry-run function won't be updated. This creates drift between what the dry-run predicts and what actually happens, undermining the purpose of the `explain` command.

**Fix direction**: Extract `checkDestructiveCommand` into a shared module that both the real handler and the dry-run can use.

---

### [LOW] `getSandboxStatus` creates new instance every time

**Location**: `src/services/diagnose-service.ts:800-850`

```typescript
async function getSandboxStatus(): Promise<HealthCheck> {
    try {
        const capability = await new SandboxCapabilityProbe().detect();
```

**Why it matters**: Creates a new `SandboxCapabilityProbe` instance every time `diagnose` is called. If the probe does heavy work (e.g., checking for binary availability), this could be expensive.

**Fix direction**: Use a cached instance or singleton pattern.

---

### [LOW] `isPlainObject` defined twice

**Location**: 
- `src/hooks/guardrails/audit-log.ts:65-70`
- `src/services/guardrail-log-service.ts:150-155`

**Why it matters**: Code duplication. If the implementation changes, both places need to be updated.

**Fix direction**: Move to a shared utility module.

---

## 📋 Unverified but plausible

- **`guardrail-log-service.ts`**: The `isPlainObject` check might not catch all edge cases (e.g., objects with `null` prototype). This is unlikely to cause issues in practice but worth noting.
- **`void` async calls**: The `void appendGuardrailDecision(...)` calls are intentional fire-and-forget, but could mask issues if the audit log becomes corrupted.

---

## 🎯 Merge Recommendation

**APPROVE_WITH_FIXES**

The PR adds valuable transparency features, but the following must be addressed before merge:

1. **CRITICAL**: Fix the `patchAgentName` null-check behavioral change in `tool-before.ts:1891`
2. **MEDIUM**: Verify `applySandboxExecution` is not exported or used elsewhere
3. **MEDIUM**: Improve `redactPath` to handle more path patterns
4. **MEDIUM**: Extract `checkDestructiveCommand` into a shared module to prevent drift

The `typeof` type-check nitpicks from the draft reviews are not defects — they are intentional defensive coding for JSONL parsing where types are unknown.

---
<sub>🔒 Reviewed locally by three on-prem models (two reviewers + a verifier) for blind-spot coverage. Findings are advisory — verify before acting.</sub>

zaxbysauce:
## PR feedback closure

Resolved pasted findings:

- F-001 MEDIUM - Fixed. `guardrail explain` now fails closed when POSIX write parsing reports `parseError`; covered by `F-001: malformed write syntax fails closed on parseError`.
- F-002 MEDIUM - Fixed. Explain shell-type routing now mirrors the real hook classifier for PowerShell/cmd patterns without adding unsupported `pwsh` behavior; covered by `$env:` and `Copy-Item` regression tests plus critic review.
- F-003 LOW - Fixed. `file_write.reason` is redacted for embedded POSIX/macOS/Windows/UNC-style paths before persistence.
- F-004 LOW - Fixed. `redactPath` now handles UNC-style paths and has a regression test.
- F-005 LOW - Fixed. Release fragment now documents known limitations around destructive evaluator duplication and best-effort redaction.
- F-006 LOW - Fixed for the PR-introduced sandbox diagnose probe/executor seams. Remaining `mock.module` calls in `diagnose-sandbox.test.ts` isolate broader pre-existing diagnose dependencies and were not introduced by this PR.
- F-007 LOW - Fixed. Capture tests now use bounded JSONL observers and type/predicate-specific waits instead of fixed sleeps.
- F-008 LOW - Fixed. Added per-type rejection tests for missing `file_write.reason`, `scope_violation.action`, `destructive_block.destructiveCategory`, `sandbox_wrap.command`, and `sandbox_skip.reason`.

Resolved live PR comments:

- Redundant `redactPath` in explain write-target builder - Fixed by storing raw decision fields once and formatting at output boundaries.
- Loose `== null` in `guardrail-log-service` - Fixed with explicit `null`/`undefined` checks.
- Legacy shell audit entries invisible - Already fixed on current PR head and preserved by tests.
- `sandbox_skip`/audit capture durability - Addressed without blocking enforcement: new advisory telemetry remains fire-and-forget, and tests wait for the eventual records with bounded observers.
- Sandbox status re-probes every diagnose call - Fixed with a cached `SandboxCapabilityProbe` plus `_internals` seam for tests.
- Malformed JSONL skip visibility - Accepted as existing graceful behavior; malformed lines are skipped without crashing and covered by tests.
- `evaluateDestructiveDecision` config/duplication caveat - Documented in the PR body/release fragment; extraction remains a future refactor because it would change the hook hot path.
- `patchAgentName` null-check behavioral claim - Disproved for this feedback patch: current code blocks absent active-agent sessions before patch writes.
- `applySandboxExecution` signature claim - Disproved: function is local to `createToolBeforeHandler`; no exported caller surface.
- UNC path redaction - Fixed.
- `isPlainObject` duplication - Accepted as low-risk local JSONL guard duplication; no behavioral defect.

Independent critic passes:

- First critic found `pwsh` classifier drift and awaited audit I/O. Both were corrected.
- Second critic confirmed `pwsh` parity was closed and flagged pre-enforcement shell-audit await. Shell audit is now fire-and-forget; tests prove both legacy shell and typed destructive records with bounded waits.

zaxbysauce:
## 🤖 Tri-Model PR Review
**Pipeline:** Mellum2-12B (code) + gemma-4-12B-agentic (broad) → Qwen3.6-35B-APEX (verify + blind-spot)
**Commit reviewed:** `98c6c6485db6`

---

# PR Review — opencode-swarm / guardrail transparency suite (#1455)

## 🔍 PR Intent (reconstructed from issue #1274 + description):
Add three capabilities:
1. `/swarm guardrail explain` – dry-run shell/write analysis, reports allow/block without execution; supports `--agent`, `--scope`, `--write`.
2. Unified decision log (`guardrail_log`) – reads `.swarm/session/shell-audit.jsonl`; optional `--blocks-only` filter for block decisions only (legacy shells + sandbox wrap excluded). Redacted output, on-demand read — no hot path cost; audit failures are fail-open and do not affect enforcement semantics per the PR description ("fire-and-forget advisory").
3. Sandbox status in `/swarm diagnose` – reports executor mechanism (`bubblewrap`, `sandbox-exec`, `windows-runner`, or `none`), availability, whether commands would actually be sandboxed (sandboxing / silent pass-through). Advisory only — absence of a sandbox executor never causes hard failure; the probe is cached through an explicit DI seam so status remains testable.

## 📦 Implementation Summary:
The PR adds four new command handlers (`guardrail_explain`, `guardrail_log`), extends `.swarm/session/shell-audit.jsonl` with five decision types (file_write, scope_violation, destructive_block, sandbox_wrap, sandbox_skip), and wires the audit log into both shell execution paths via a DI seam (`src/hooks/guardrails/_internals`). The `diagnose` service is extended to probe capability through an injected interface so tests can mock executor behavior.

## ✅ / ⚠️ / ❌ Intended vs Actual:
| Obligation | Status | Evidence (file:line) | Notes |
|---|---|---|---|
| Dry-run explain reports allow/block without execution, supports flags | `SUPPORTED` | src/services/guardrail-explain-service.ts — handleGuardrailExplain() returns a markdown report; command is never executed in the service layer (`evaluateDestructiveDecision()` only analyzes) | Matches description exactly |
| Unified decision log reads shell_audit.jsonl, redacted output on demand, `--blocks-only` filter works as described (legacy shells + sandbox wrap excluded), audit failures are fail-open and do not affect enforcement semantics ("fire-and-forget advisory") | `SUPPORTED` | src/services/guardrail-log-service.ts — readAuditLog() parses JSONL line by line; formatSummary(entry) redacts paths via redactPath(); blocksOnly filter excludes non-block types (shell, sandbox_wrap); audit failures are caught and logged but never propagate as exceptions to the tool execution path | Matches description exactly |
| Sandbox status in `/swarm diagnose` reports mechanism + availability + whether commands would actually be sandboxed; advisory only — absence of executor does not cause hard failure. Probe is cached through DI seam so tests can mock it, preserving testability without live platform executors required for integration coverage (covered by TODO) | `SUPPORTED` with one caveat: sandbox probe/executor mocks are moved to `_internals`; this means the actual capability detection path remains untested in unit — covered only via a separate "CI cross-platform suite" that is not yet implemented. The advisory nature of diagnose status on failure (`⬜`) matches description ("advisory and testable"). | Matches description except for integration coverage gap noted above; no critical issue here, but the unverified risk below covers it as `NIT`. |

## 🚨 Confirmed Findings:
#### [HIGH] Title — Shared destructive-command evaluator duplication (task 1.2)
**Location:** src/services/guardrail-explain-service.ts evaluateDestructiveDecision() (~line 943, ~200 lines of nested ifs). Also present in the toolBefore hook at `src/hooks/guardrails/tool-before.ts` (lines around 1825) as checkDestructiveCommand().
**Why it matters:** The real guardrail evaluator is duplicated — evaluateDestructiveDecision() mirrors but does not share logic with checkDestructiveCommand(), so parity regressions for the feedback cases are covered by tests only, and any new destructive-command pattern added to one place won't be reflected in the other.
**Evidence:** `evaluateDestructiveDecision()` contains a TODO comment: "TODO(#1274): checkDestructiveCommand is a nested closure... shared evaluator remains; future refactor can extract top-level evaluateDestructiveDecision()". The same logic appears twice with different implementations (regexes, target extraction).
**Fix direction:** Extract the destructive-command evaluation into a single module (`src/hooks/guardrails/destructive.ts`) and import it from both places — this is exactly what the TODO points to as follow-up work; doing so now would close an existing gap rather than leaving duplication in place for another PR cycle, which aligns with "polish generated" discipline (the service has a `POLISH_GENERATED` finding).

#### [HIGH] Title — Redaction coverage is best-effort and does not cover arbitrary custom path layouts; callers remain responsible for domain-specific secrets embedded in paths.
**Location:** src/hooks/guardrails/audit-log.ts redactPath() (~line 154, ~20 lines). The function handles POSIX home (`/home/...`), macOS (`Users/...`), Windows profile (C:\Users\...), and UNC server prefixes — but it does not canonicalize or scrub arbitrary path segments that may contain secrets embedded in directory names; callers must redact those separately.
**Why it matters:** If an agent writes to a custom workspace where the user's home is mapped under `/custom/home/` (e.g., via bind mount), `redactPath()` will not recognize this as sensitive and leave raw paths like `/custom/home/alice/.swarm/...` in audit logs, leaking usernames across environments or even exposing secrets if a path contains them embedded mid-segment (`.../secret_key=abc123xyz`).
**Evidence:** redactPath() normalizes separators but does not compare against any configured sensitive paths; it only matches known home patterns. The comment in the function explicitly states: "This is intentionally minimal — callers remain responsible for stripping domain-specific secrets that may appear in path segments." This admission of partial coverage means a user could accidentally leak data via an unredacted audit entry if their workspace layout deviates from standard profiles, and there are no tests exercising this edge case.
**Fix direction:** Add configurable sensitive paths (e.g., `audit_log: { redactPaths: ['/custom/home/', '/mnt/secrets'] }`) or a shared path-redaction utility that callers invoke before constructing entries; the current approach is brittle for non-standard workspaces and should be hardened to cover arbitrary patterns via explicit configuration rather than hardcoding known profiles.

#### [MEDIUM] Title — Shell audit log entry shape changed from legacy (type discriminator stripped) but shell mode still writes 5 fields without type field, which could break downstream consumers that expect the old schema with `type: 'shell'`. The PR description says "legacy entries persist as exact 5-field shape" and this is correct for current behavior.
**Location:** src/hooks/guardrails/audit-log.ts appendGuardrailDecision() — shell case (~line 305, ~2 lines). `type` discriminator is stripped; entry has `{ ts, sessionID, agent, tool, command }`. The comment above the switch explicitly states this preserves original shape for legacy compatibility: "Legacy shell entries persist as exact 5-field shape with type discriminator stripped."
**Why it matters:** If any downstream consumer (e.g., an analytics pipeline or audit viewer) expects `type` to be present on all JSONL lines, they will see missing fields and may misclassify events; the PR description says this is intentional ("additive schema"), but consumers should verify their expectations match — especially since shell entries are now mixed with new-style typed entries in the same file.
**Evidence:** The switch case for `shell` does not include a type field at all (line ~305). A consumer that expects every entry to have `{ ts, sessionID, agent, tool, command, type }` will see missing keys and may fail parsing or misroute events; the PR description says this is additive schema behavior but doesn't explicitly warn downstream consumers.
**Fix direction:** Add a comment in `appendGuardrailDecision()` documenting that shell entries omit `type` (legacy shape) while other types include it, so any consumer should handle both cases gracefully — or add an optional flag to preserve type for backward compatibility if the project has strict schema expectations elsewhere; currently this is documented only implicitly.

#### [LOW] Title — Sandbox probe/executor mocks are moved to `_internals`; sandbox diagnose status remains advisory and testable but actual platform executor behavior (bwrap, nsjail) stays untested in unit tests because they're skipped when no executor exists per the design ("advisory"). The TODO on CI cross-platform suite is a real gap.
**Location:** src/services/diagnose-service.ts — _internals = { detectSandboxCapability: () => sandboxCapabilityProbe.detect(), getSandboxExecutor: getExecutor }, and `getDiagnoseData` uses these via DI seam so tests can mock them without requiring live executors (AGENTS.md invariant 7).
**Why it matters:** The unit test suite cannot verify that the actual platform executor is detected correctly because when no bwrap/nsjail exists, probe returns `{ mechanism: 'none', available: false }` and diagnose reports `⬜` with "Commands not sandboxed" — this matches description ("advisory only") but means a user on Linux could be misled into thinking sandboxing works if their executor is missing or broken. The TODO explicitly calls out that cross-platform suite CI needs to cover actual platform behavior, which isn't implemented yet; without it there's no way to verify the probe against real executors across OSes and environments (Windows runner on Windows vs Linux bwrap).
**Evidence:** `getDiagnoseData` uses `_internals.detectSandboxCapability()` — a function that can be mocked in tests via DI seam, so unit coverage is high but integration/platform-level correctness remains unverified; the TODO "CI cross-platform suite" (line ~958) explicitly flags this gap as not yet implemented and notes it's intentional ("advisory").
**Fix direction:** Implement a CI cross-platform test matrix that runs on Linux, macOS, Windows with actual sandbox executors available in each environment to verify probe/executor detection against real platform behavior; the current unit tests cover mock paths but cannot catch regressions where `bubblewrap` exists yet isn't detected correctly due to path resolution or capability checks.

## 🔬 Unverified but Plausible Risks:
#### [NIT] Title — Prompt injection risk in guardrail_explain's shell command parsing (low confidence). The service returns a markdown report of what the guardrails WOULD do without executing anything, so prompt-injection payloads passed to `guardrail explain` are not executed. However, if an agent passes malicious content that contains sensitive paths or secrets into `--write <path>` mode and those values get redacted via shell command parsing (redactShellCommand is used on output but input isn't sanitized), the audit log could leak raw strings containing tokens embedded in commands — though since no execution occurs this doesn't lead to code injection, it does risk leaking sensitive data from agent-provided arguments.
**Why suspicious:** The service uses `dcNormalizeCommand(rawCommand)` and other destructive-command helpers on input for analysis but never sanitizes the command string itself before passing through redactShellCommand() which is designed only for output; if an attacker crafts a shell argument containing secrets (e.g., `"echo 'TOKEN=abc123' > lib/output"`), those tokens appear in audit logs because they're part of `rawCommand` passed to dcNormalizeCommand, and redactShellCommand() doesn't scrub embedded paths or API keys — it only strips known token patterns at the start-of-line. Since no execution happens this is not a code injection risk but does leak agent input into persistent audit logs without sanitization; callers should be aware that guardrail_explain reports may contain raw command strings with secrets if they pass them directly, and redactShellCommand() only partially cleans output (it strips leading tokens like `TOKEN=` at the start of lines).
**What would verify it:** A test passing a shell argument containing an embedded API key into `--write` mode should confirm whether that key appears in audit logs; if yes then sanitization is needed for input, not just output. The PR description doesn't claim redaction applies to inputs — only outputs are explicitly mentioned as "redacted" via `redactShellCommand()` on the command field of shell entries and path fields elsewhere.

#### [NIT] Title — Test count arithmetic: if the PR claims 'N new tests', verify N (low confidence). The test plan lists 103 pass, 1 todo; no explicit claim about how many were added by this PR specifically vs existing ones or other PRs merged in same cycle. This is not critical but worth checking against commit diff to confirm actual count matches the description's implied "N new tests" language (the TODO on CI cross-platform suite also implies more work than currently implemented).
**Why suspicious:** The test plan says 103 pass, 1 todo — this includes all unit/integration tests for `guardrail_explain`, `guardrail_log`, and diagnose sandbox probe. Without checking the commit diff against prior state it's impossible to confirm how many were added by #1455 specifically versus existing ones or other PRs merged in same cycle; if someone later claims "N new tests" based on this PR, they'd need a precise count from `git log -p` rather than relying on the test plan summary.
**What would verify it:** Run `git diff --stat HEAD^..HEAD | grep '^ \d\+ files changed'` and sum lines added in unit/integration test directories; compare to "103 pass" claim (the 1 todo is likely existing). The PR description doesn't explicitly state N, so this isn't critical but worth flagging as low-confidence.

## 🧪 Test / Coverage Gaps:
#### [Gap] Sandbox diagnose tests use `_internals` for the probe/executor path — meaning actual platform executor behavior (bwrap on Linux, sandbox-exec on macOS) is not tested in unit because it's mocked; this matches design ("advisory and testable") but means integration coverage of real executors remains unverified.
**Evidence:** `src/services/diagnose-service.ts` uses `_internals = { detectSandboxCapability: () => ... , getSandboxExecutor: getExecutor }` so tests can mock these functions directly (function mocks restore cleanly in afterEach per AGENTS.md invariant 7). The TODO on CI cross-platform suite explicitly calls out this gap as not yet implemented and notes it's intentional ("advisory").
**Fix direction:** Implement the missing integration test matrix that runs diagnose against real platform executors across Linux, macOS, Windows to verify probe/executor detection works correctly in each environment; unit tests cover mock paths but cannot catch regressions where `bubblewrap` exists at runtime yet isn't detected due to path resolution or capability checks.

## 📋 Shipped-vs-Claimed Gaps:
#### [Gap] `/swarm guardrail explain` still duplicates part of the destructive-command evaluator because real validator is nested in hook; this PR adds parity regressions for feedback cases (the TODO on shared evaluator). Extracting a shared evaluator remains follow-up direction. The test plan doesn't explicitly verify that `guardrail_explain`'s analysis matches what would happen if command were executed — it only checks output format and flags, not semantic correctness of the decision logic against real shell execution behavior for edge cases (e.g., commands with embedded newlines or complex quoting).
**Evidence:** The PR description says "reuses existing detection functions directly" but `evaluateDestructiveDecision()` is a separate implementation from checkDestructiveCommand() in toolBefore; they share some helpers (`dc*`, isInDeclaredScope, DC_SAFE_TARGETS) via imports which means parity regressions are covered by tests only. The TODO on shared evaluator explicitly calls this out as follow-up work ("extract top-level evaluateDestructiveDecision()") and the PR description says "parity regressions for feedback cases; extracting a shared evaluator remains."
**Fix direction:** Extract destructive-command evaluation into `src/hooks/guardrails/destructive.ts` with one canonical implementation, import it from both guardrail_explain service and toolBefore hook — this would close an existing gap rather than leaving duplication in place for another PR cycle; the TODO points to exactly what needs doing now if we want immediate improvement over "covered by tests only".

## 📝 Merge Recommendation: **[APPROVE_WITH_FIXES]**
One high-confidence finding (shared evaluator duplication) and one unverified risk that is not critical — neither blocks merge. The PR delivers the requested features with correct behavior for all specified capabilities, test coverage adequate except where design intentionally defers to integration/platform testing ("advisory"), and no security issues at current confidence level given dry-run nature of explain service (no execution occurs).

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ / ❌ |
| No unresolved STEALTH_CHANGE with missing test coverage for parity regressions | ✅ / ❌ — the shared evaluator duplication is a `STEALTH_CHANGE` but tests cover it; no unverified gap here beyond what's already flagged as NIT/low-confidence risk. The TODO on CI cross-platform suite and sandbox probe mocks are design choices, not gaps requiring immediate action for merge approval since they're explicitly marked "advisory" in the PR description (user can address later). |
| No UNSUPPORTED obligations with no code evidence | ✅ / ❌ — all five capabilities have implementation; nothing claimed without backing. The TODO on shared evaluator is follow-up work, not an unsupported obligation for this merge cycle because tests cover parity regressions as described ("covered by existing PR tests"). |
| Test coverage adequate (bounded observers instead of fixed sleeps) | ✅ / ❌ — yes, bounded JSONL observer pattern used in guardrail_log service; no other major test gaps. The sandbox diagnose unit mocks are intentional per design and not a gap requiring immediate action for merge approval since they're explicitly marked "advisory" ("tests use _internals"). |
| No hardcoded secrets exposed (fail-open, redaction) | ✅ / ❌ — audit failures fail open as described; path/command fields in logs contain raw values but no secret exposure confirmed. The TODO on full canonical scrubber is a follow-up improvement not blocking merge since current behavior matches description ("best-effort and covers standard POSIX..."). |
| All async errors handled (guardrail_log, diagnose) | ✅ / ❌ — both services catch audit failures internally; guardrail_explain returns markdown without throwing on analysis failure. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken agent role boundaries (prompt injection risk low confidence) | ✅ / ❌ — guardrail_explain returns markdown without executing; prompt payloads are advisory only. The NIT on input sanitization is not critical given dry-run nature and would be addressed by follow-up hardening if desired, but does not block merge at current confidence level since no execution occurs (prompt injection risk applies to output redaction of inputs which doesn't lead to code execution). |
| No broken tool allowlist drift (commands/exports covered) | ✅ / ❌ — command registry updated in src/commands/index.ts; tests cover existing PR's test_runner usage and sandbox diagnose path via _internals mocks as described ("tests use _internals"). The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description (user can address later). |
| No broken release/cache versioning (docs updated) | ✅ / ❌ — docs/releases/pending added; no version files edited. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| Lockfile consistent (no drift) | ✅ / ❌ — verified with `bun run build` and node --input-type=module -e. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken plan durability (not touched) | ✅ / ❌ — invariant audit says untouched; no changes to src/plan. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken test_runner safety (not touched) | ✅ / ❌ — invariant audit says untouched; validation used shell commands, not broad `test_runner`. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken session state (not touched) | ✅ / ❌ — invariant audit says untouched; no changes to src/session. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken guardrails/retry (touched, but retry semantics unchanged) | ✅ / ❌ — touched; new audit telemetry fire-and-forget does not change retry/circuit breaker per invariant. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken chat/system msg (not touched) | ✅ / ❌ — invariant audit says untouched; no changes to src/chat. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken tool registration (touched, covered by existing tests) | ✅ / ❌ — touched; command registry/export surfaces already covered by existing PR tests. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken release cache (touched, docs updated) | ✅ / ❌ — touched; `docs/releases/pending` added. The TODO on CI cross-platform suite is not a gap requiring immediate action for merge approval since it's explicitly marked "advisory" in the PR description ("tests use _internals"). |
| No broken agent role boundaries (prompt injection risk low confidence) — reiterating NIT above, but this time as final check. The guardrail_explain service returns markdown without executing anything; prompt payloads are advisory only and do not lead to code execution at current confidence level given dry-run nature of explain service ("fires fire-and-forget telemetry so audit I/O cannot delay enforcement"). | ✅ / ❌ — yes, no critical issue here. The NIT on input sanitization is a follow-up improvement that would be addressed by hardening if desired but does not block merge at current confidence level since prompt payloads are advisory only and do not execute (prompt injection risk applies to output redaction of inputs which doesn't lead to code execution). |

---
<sub>🔒 Reviewed locally by three on-prem models (two reviewers + a verifier) for blind-spot coverage. Findings are advisory — verify before acting.</sub>

PR #1478, opened 2026-06-22
fix(knowledge): unblock skill maturity + accrue outcomes (#1477)

### LINKED ISSUE #1477: Skill maturation pipeline is stalled (0/102 entries eligible for compilation)
# Knowledge → Skill maturation pipeline is stalled (0/102 entries eligible)

## Summary

The skill maturation pipeline is producing **0 mature uncompiled clusters** despite having **102 swarm knowledge entries** and **38 draft skill proposals** in `.swarm/skills/proposals/`. The skill_improver run on 2026-06-21 surfaced this as the dominant issue: 8 of 9 active skills are stale (missing `source_knowledge_ids`) and no new skills are being compiled because no entries clear the maturity gate.

This is not a pipeline bug — it's the deterministic consequence of three compounding issues that need to be diagnosed and fixed.

## Problem statement

The skill generator at `src/services/skill-generator.ts:154-185` requires a knowledge entry to pass **all** of these gates before it can be considered for compilation:

```typescript
// Line 174: confidence gate
if (entry.confidence < opts.minConfidence && !strongOutcomes) return false;

// Line 184: confirmation gate
return distinctPhases >= opts.minConfirmations || strongOutcomes;
```

Where:
- `minConfidence` defaults to 0.7 (`DEFAULT_SKILL_MIN_CONFIDENCE`)
- `minConfirmations` defaults to 2 distinct phases (`DEFAULT_SKILL_MIN_CONFIRMATIONS`)
- `strongOutcomes` means `applied_explicit_count >= 3 || succeeded_after_shown_count >= 3`

### Confidence distribution — 0/102 entries clear the 0.7 floor

| Confidence bucket | Count |
|---|---|
| 0.0–0.4 | 11 |
| 0.4–0.5 | 1 |
| 0.5–0.6 | 1 |
| 0.6–0.7 | **89** |
| 0.7–0.8 | **0** |
| 0.8–0.9 | **0** |
| 0.9–1.0 | **0** |

89 of 102 entries are clustered **just below** 0.7 — they look like they were created at default confidence (~0.5–0.6) and then **never received the confidence boost** that the system is supposed to apply on successful retrieval + application.

### Retrieval outcomes — none recorded

```
No outcomes:           102/102
applied > 0:            0/102
succeeded > 0:          0/102
STRONG (>=3):           0/102
```

The positive-gate escape hatch at line 170 (`if (outcomeSignal > 0 && strongOutcomes) return true`) would let low-confidence entries through — but **not a single entry has any outcome recorded**.

### Confirmations — 89 entries only confirmed in 1 phase

```
Has any confirmations:  89/102
By distinct phase count:
  0 distinct phases:  13
  1 distinct phase:   89
```

The maturity gate requires **2 distinct phases** (line 184). Entries created mid-plan and only confirmed in one phase have no path to maturity within the same plan.

### Status — all 102 are `candidate`, none are `established` or `promoted`

```
status: { candidate: 102 }
```

### Directive priority distribution

```
critical: 6
high: 30
medium: 45
low: 4
none: 17
```

36 entries have `critical` or `high` directive priority. The `isHighPriorityDirective()` function at line 147 is used **only** in cluster formation (line 310, allowing single-entry clusters). It does **not** bypass the maturity gate at line 174.

### Simulated eligibility with current defaults

```
ELIGIBLE:                                  0
BLOCKED by confidence:                  102
BLOCKED by confirmations (<2 phases):     0
```

**100% of entries are blocked at the confidence gate.**

### #1 — Confidence never gets boosted past 0.6

The flat distribution at 0.6–0.7 strongly suggests entries are created at default confidence and then never receive the boost that the system is designed to give on successful application. Most likely cause:

- The `knowledge-application` hook at `src/hooks/knowledge-application.ts` is the code path that should boost confidence on successful retrieval + application.
- If outcomes are 0, the boost never fires, and confidence never climbs.
- This is a **bootstrap problem**: outcomes are needed to boost confidence, but the retrieval hooks may gate on confidence floors or the application gate may not be invoked without an active retrieval hit.

### #2 — Outcomes are not being recorded at all

Either:
- The knowledge-application hook is not running, OR
- It's running but writing to a different path than the JSONL the simulator reads, OR
- It's running but the confidence-update side effect is being filtered out.

This is the **highest-leverage single fix** — once outcomes start flowing, the entire chain unsticks.

### #3 — 2-phase confirmation is a high bar for new knowledge

A knowledge entry that originates during a multi-phase plan must persist across at least 2 phase boundaries to mature. Many entries are created in the middle or end of a plan and never get that second confirmation.

### #4 — `directive_priority` bypass doesn't actually bypass the gate

The `isHighPriorityDirective()` function bypasses the cluster-size floor (allowing single-entry clusters) but does NOT bypass the confidence floor. So 36 high/critical entries are blocked by the same 0.7 confidence requirement as everything else.

### Fix 1: Diagnose the outcomes pipeline (highest priority)

Trace why `applied_explicit_count` and `succeeded_after_shown_count` are 0 across all entries.

**Investigation steps:**
1. Verify the knowledge-application hook is registered and running. Check `src/hooks/knowledge-application.ts` and the hook registration in `src/hooks/`.
2. Check if `knowledge-application` is wired into the tool execution flow or the message transform path. Compare with `src/hooks/knowledge-events.ts` (which has `readKnowledgeCounterRollups` and `effectiveRetrievalOutcomes`).
3. Add a single test that exercises the full outcomes pipeline: create an entry → trigger retrieval → confirm it was applied → assert `applied_explicit_count` increments.
4. If the hook IS running but outcomes aren't landing, check the write path — maybe outcomes are being written to a different JSONL or the counter rollup is stale.

**Expected impact:** Should unblock the entire maturation chain once outcomes start flowing.

### Fix 2: Add a 1-phase confirmation path (low-risk, high-value)

Allow entries with single-phase confirmation AND high/critical directive_priority AND confidence ≥ 0.6 to mature. This is a 5-line change at `src/services/skill-generator.ts:184`.

```typescript
// Current:
return distinctPhases >= opts.minConfirmations || strongOutcomes;

// Proposed:
const hasHighPrioritySinglePhase = isHighPriorityDirective(entry) && distinctPhases >= 1 && entry.confidence >= 0.6;
return distinctPhases >= opts.minConfirmations || strongOutcomes || hasHighPrioritySinglePhase;
```

**Expected impact:** Unblocks 36 high/critical entries currently stuck.

### Fix 3: Lower the confidence floor for high-priority directives

Allow `isHighPriorityDirective()` to also bypass the confidence gate, not just the cluster-size gate. One-line change at `src/services/skill-generator.ts:174`.

```typescript
// Current:
if (entry.confidence < opts.minConfidence && !strongOutcomes) return false;

// Proposed:
if (entry.confidence < opts.minConfidence && !strongOutcomes && !isHighPriorityDirective(entry)) return false;
```

**Expected impact:** Same 36 entries, different mechanism. Pairs well with Fix 2.

### Fix 4 (preventive): Boost initial confidence for high-priority directives

New entries created with `directive_priority=critical` or `high` should start at 0.7, not 0.5. This requires tracing the entry-creation code path. Most likely location: where `swarm_memory_propose` constructs the entry, or wherever knowledge is added with directive metadata.

**Expected impact:** Preventive — helps future entries skip the bootstrap problem.

## Acceptance criteria

- [ ] At least 50 of the 102 current entries pass the maturity gate (with the 38-draft proposal pipeline unblocked) — measured by `skill_generate` reporting `mature_uncompiled_clusters > 0`.
- [ ] `applied_explicit_count` or `succeeded_after_shown_count` is non-zero for at least 10 entries (proving the outcomes pipeline is alive).
- [ ] The 8 stale active skills in `.opencode/skills/generated/` either get their `source_knowledge_ids` repaired or are explicitly retired.
- [ ] A regression test asserts that creating an entry, triggering retrieval, and confirming application increments `applied_explicit_count` and boosts `confidence` to ≥ 0.7.

## Files involved

- `src/services/skill-generator.ts` — maturity gate (lines 154–185), clustering (215+)
- `src/services/skill-improver.ts` — proposal generator, diagnostic
- `src/hooks/knowledge-application.ts` — application hook (line 410+ has `min_confidence: 0.85`)
- `src/hooks/knowledge-events.ts` — `readKnowledgeCounterRollups`, `effectiveRetrievalOutcomes`
- `src/hooks/knowledge-store.ts` — `readKnowledge`, `computeOutcomeSignal`
- `src/hooks/knowledge-types.ts` — `KnowledgeEntryBase` shape
- `.swarm/knowledge.jsonl` — the data being filtered
- `.opencode/skills/generated/` — 9 active skills (8 stale, 1 healthy)
- `.swarm/skills/proposals/` — 38 draft proposals

## Related

- skill_improver proposal at `.swarm/skill-improver/proposals/2026-06-21T22-04-17-850Z.md`
- skill_generate run on 2026-06-21T22:04 (0 drafts written)
- Post-mortem for "lowtier-Memory_System_Phase_1__Correctness___Hygiene" (this run): 100 knowledge entries reviewed, 100 never-applied entries flagged, 39 pending proposals, 5 quarantined entries

zaxbysauce:
Closes #1477

## Summary

- **High-priority directive maturity path**: `critical`/`high` directive entries confirmed in at least one phase and at or above a 0.6 confidence floor (`HIGH_PRIORITY_SKILL_MIN_CONFIDENCE`) now satisfy the skill-maturity gate. This decouples the confidence floor from the 2-distinct-phase confirmation requirement for the highest-value knowledge (RC-1/RC-4 bug).
- **Outcome accrual fixed (root cause RC-OUTCOMES)**: `effectiveRetrievalOutcomes` merged stored entry counts with the event rollup using `{ ...base, ...rollup }`, so the always-zero production rollup silently overwrote real outcome counts. The merge is now **additive** for outcome-count fields (preserves history, no migration, no double-count). `updateRetrievalOutcome` now emits immutable `'outcome'` events as the single source of truth instead of mutating entries.
- **Outcome events wired**: the previously-missing `'outcome'` producer is now supplied by `updateRetrievalOutcome`, feeding the existing test-covered fold consumed by learning-metrics.
- **Stale-skill self-healing (Part C)**: closed unparseable-frontmatter detection blind spot, added deterministic regenerate-or-retire reconciliation wired into skill-improver, marked shims with `skill_origin: shim` for skip, converted committed `guardrail-patterns` to proper generated frontmatter, marked `pr-review-fix` as a shim.
- **Corrupt-baseline resilience**: `readCounterBaseline` now fails open scoped to the baseline — a corrupted `knowledge-counter-baseline.json` no longer propagates to the outer rollup catch and discards the live event log. The folded baseline is dropped while live events are replayed, so outcome accrual survives corruption.

## Invariant audit

- 1 (plugin init): not touched — changes are to runtime service code (`knowledge-events`, `knowledge-reader`, `skill-generator`, `skill-improver`), not init path; `src/index.ts` unchanged
- 2 (runtime portability): not touched — no top-level `bun:` imports, no `Bun.*` calls outside `bun-compat.ts`, no plugin shape changes; tsc 0 errors
- 3 (subprocesses): not touched — no new subprocess calls; `git diff --name-only origin/main..HEAD | xargs grep -nE 'bunSpawn\(|spawn\(|spawnSync\(' || true` returns nothing
- 4 (.swarm containment): not touched — events appended to existing `.swarm/knowledge-events.jsonl` pattern, `recordKnowledgeEvent` already wrapped in try/catch per AGENTS.md §1; `readCounterBaseline` error handling scoped and logged non-fatally
- 5 (plan durability): not touched — no changes to `.swarm/plan-*` or ledger
- 6 (test_runner safety): not touched — only direct `bun:test` use, no `test_runner` calls
- 7 (test writing): touched — new tests (24 in knowledge-events, 7 in skill-improver-reconcile) use bun:test, proper mocking via `_internals` seam, temp dirs with `mkdtemp`, all 102 tests pass in isolation; test-file isolation verified per Invariant 7 (bun:test handles `mock.module` isolation via _internals seam)
- 8 (session state): not touched — no session-scoped behavior added
- 9 (guardrails/retry): touched — `readCounterBaseline` error handling demonstrates scoped fail-open (parse failure caught, logged, returns empty Map); outer catch in `readKnowledgeCounterRollups` still wraps for defence-in-depth, but scoped error prevents silent loss of live event log
- 10 (chat/system msg): not touched — no message handling changes
- 11 (tool registration): not touched — no new tools, no registration changes
- 12 (release/cache): not touched — release fragment at `docs/releases/pending/1477-skill-maturity-outcome-accrual.md` created correctly; `package.json` version, `CHANGELOG.md`, and `.release-please-manifest.json` untouched

## Test plan

- [x] `bun --smol test tests/unit/hooks/knowledge-events.test.ts --timeout 30000` — 24 pass (including new corrupt-baseline test)
- [x] `bun --smol test tests/unit/hooks/knowledge-reader.test.ts --timeout 30000` — 20 pass (migrated outcome tests)
- [x] `bun --smol test tests/unit/hooks/knowledge-reader-key-normalization.test.ts --timeout 30000` — all pass
- [x] `bun --smol test tests/unit/services/skill-generator.test.ts --timeout 30000` — 51 pass (high-priority path + Part A predicate tests)
- [x] `bun --smol test tests/unit/services/skill-improver-reconcile.test.ts --timeout 30000` — 7 pass (reconciliation, shim-skip, dry-run, stale detection)
- [x] `bunx tsc --noEmit` — 0 errors
- [x] `bunx biome ci .` — 0 errors (all formatting/import issues fixed)
- [x] All tests in isolation (per file): 102 pass, 0 fail
- [x] Independent implementation review returned APPROVE (no blockers); pipeline sweep ruled out additional blockers

## Known caveats

- **Cross-file test isolation (pre-existing, not introduced):** Running all four affected files together via `bun test` (not in isolation) produces 23 spurious failures due to file-scoped `mock.module` leaks from knowledge-reader.test.ts. Each file passes alone (as verified above). CI runs per-file, so it stays green. This is a pre-existing issue tracked separately — not a regression from this PR.
- **AC1 literal "≥50" not force-met:** Per the evidence-based decision in Q1 of the planning phase, we ship ~36 high/critical unblocks (targeted fix) plus whatever had outcomes the merge was discarding (RC-OUTCOMES fix), rather than loosening the global gate for all 102. This is an intentional, quality-first deviation documented in the plan.
- **Baseline recovery:** Corrupt `knowledge-counter-baseline.json` self-heals on the next FIFO eviction (when the event log exceeds MAX_EVENT_LOG_ENTRIES and `foldEvictedEventsIntoBaseline` rewrites the baseline). Until then, the corrupt file sits on disk but causes no functional stall (live events still replay, maturation works).

---
_Generated by [Claude Code](https://claude.ai/code/session_018SvPUTRy2nZUDVuqVfmc57)_


zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `92050ad19a6d`

---

---

### 🔍 PR Intent

> Reconstructed obligation list from PR description, issue #1477, commit messages, and changed tests.

- **O-001**: Fix outcome accrual root cause — `effectiveRetrievalOutcomes` must preserve historical entry counts instead of letting zero-valued event rollup overwrite them
- **O-002**: Wire outcome events — `updateRetrievalOutcome` must emit immutable `'outcome'` KnowledgeEvents as single source of truth (replacing in-place entry mutation)
- **O-003**: Add high-priority directive maturity path — `critical`/`high` directives confirmed in ≥1 phase at ≥0.6 confidence should satisfy the maturity gate
- **O-004**: Fix corrupt baseline resilience — `readCounterBaseline` must fail open scoped to the baseline without discarding live event log
- **O-005**: Close unparseable-frontmatter blind spot — skills with no parseable YAML frontmatter must be surfaced as stale (not silently skipped)
- **O-006**: Add stale-skill self-healing — `reconcileStaleActiveSkills` pass must regenerate-or-retire stale active skills
- **O-007**: Mark shims for skip — hand-authored compatibility shims marked `skill_origin: shim` must not be flagged or reconciled
- **O-008**: Convert committed `guardrail-patterns` skill to proper generated frontmatter; mark `pr-review-fix` as shim

---

### 📦 Implementation Summary

**Outcome accrual fix:** `updateRetrievalOutcome` no longer reads/mutates swarm/hive knowledge files. Instead it emits one immutable `'outcome'` KnowledgeEvent per shown id via `recordKnowledgeEvent`. The event log is the single source of truth; `recomputeCounters` folds these events into a rollup; `effectiveRetrievalOutcomes` adds historical entry counts to the rollup (additive merge for `succeeded_after_shown_count` and `failed_after_shown_count` only).

**High-priority path:** New constant `HIGH_PRIORITY_SKILL_MIN_CONFIDENCE = 0.6`. `isSkillMaturityEligible` evaluates negative outcome signal first (unchanged), then checks strong outcomes, then the high-priority path (`isHighPriorityDirective && distinctPhases ≥ 1 && confidence ≥ 0.6`), then legacy gates. The high-priority path bypasses both the 0.7 confidence floor and the 2-distinct-phase requirement.

**Corrupt baseline resilience:** `readCounterBaseline` wraps `JSON.parse` in try/catch; on failure it logs a warning and returns empty Map, allowing `readKnowledgeCounterRollups` to fall through to live event replay instead of discarding everything.

**Stale-skill reconciliation:** New function `reconcileStaleActiveSkills` iterates active skills, detects unparseable frontmatter or missing `source_knowledge_ids`, then either calls `regenerateSkill` or `retireSkill`. Wired into `runSkillImprover` with `dryRun: writeMode !== 'draft_skills'`.

**Frontmatter/shim changes:** `guardrail-patterns/SKILL.md` gets proper YAML frontmatter with `skill_origin: generated`; `pr-review-fix/SKILL.md` gets `skill_origin: shim`. `gatherInventory` skips shim-marked skills.

---

|---|---|---|
| O-001 Outcome accrual fix | `SUPPORTED` | `src/hooks/knowledge-events.ts:638-652` — additive merge for two outcome fields |
| O-002 Outcome event wiring | `SUPPORTED` | `src/hooks/knowledge-reader.ts:594-607` — emits `'outcome'` events per id |
| O-003 High-priority path | `SUPPORTED` | `src/services/skill-generator.ts:186-203` — new path with 0.6 floor and ≥1 phase |
| O-004 Corrupt baseline | `SUPPORTED` | `src/hooks/knowledge-events.ts:626-640` — scoped try/catch returns empty Map |
| O-005 Unparseable frontmatter | `SUPPORTED` | `src/services/skill-improver.ts:189-197` — surfaces `unparseable_frontmatter` reason |
| O-006 Stale-skill reconciliation | `SUPPORTED` | `src/services/skill-improver.ts:396-480` — full reconcile function; wired at line 736 |
| O-007 Shim skip | `SUPPORTED` | `src/services/skill-improver.ts:198-200` and line 435 — both paths check `skillOrigin === 'shim'` |
| O-008 Frontmatter conversion | `SUPPORTED` | `.opencode/skills/generated/guardrail-patterns/SKILL.md:1-17` and `.opencode/skills/generated/pr-review-fix/SKILL.md:7` |

---

### 🚨 Confirmed Findings

#### [CRITICAL] Outcome merge incomplete — future field additions silently lost

- **Location**: `src/hooks/knowledge-events.ts:638-652`
- **Why it matters**: The additive merge only covers `succeeded_after_shown_count` and `failed_after_shown_count`. Any future outcome-count fields (e.g., `partial_after_shown_count`, `retry_count`, or new metrics) will be silently overwritten by the `{ ...base, ...rollup }` spread. This is a maintenance-timebomb: the comment claims "Only these two counts were ever written to the entry" but that is a historical claim, not a structural guarantee.
- **Evidence**: Lines 638-652 explicitly list only two fields for additive merging. The remaining rollup fields (including `partial_after_shown_count`) ride the plain spread at line 639 (`...rollup`), meaning any future addition to rollup that isn't in base gets zeroed out by the spread precedence when base has a value.
- **Fix direction**: Use an explicit merge strategy or document the invariant as a runtime assertion; consider a helper function that validates both base and rollup have matching outcome-count field sets.

#### [CRITICAL] `updateRetrievalOutcome` no longer verifies entry existence before emitting events

- **Location**: `src/hooks/knowledge-reader.ts:594-607`
- **Why it matters**: The old code used `transactKnowledge` which would read the swarm/hive files, find matching entries, and only mutate those found. The new code emits outcome events for every id in `shownIds` regardless of whether the entry exists in any store. If an id is no longer present (e.g., archived or deleted), this creates orphaned outcome events that will accumulate in the event log forever, inflating rollup counts and potentially causing memory pressure on the event log.
- **Evidence**: Lines 594-607 iterate `shownIds` directly without any lookup to verify entry existence. The old code at lines 580-612 (before deletion) had explicit entry matching: `if (shownIds!.includes(entry.id))`.
- **Fix direction**: Either skip ids not found in swarm/hive (reintroduce the lookup), or accept orphaned events but document this behavior and add cleanup for stale outcome events.

#### [HIGH] Stale-skill reconciliation dryRun semantics are inverted from the comment

- **Location**: `src/services/skill-improver.ts:742`
- **Why it matters**: The comment says "Only mutates skills when this run is allowed to write skills (draft_skills mode); otherwise reports retire-candidates via a dry run." But `dryRun: writeMode !== 'draft_skills'` means: when `writeMode === 'draft_skills'`, `dryRun = false` (mutates — correct); when `writeMode !== 'draft_skills'`, `dryRun = true` (reports only — correct). Wait, this is actually **correct**. Let me re-read...

Actually, I need to re-examine this. The reviewer claim was that it's inverted. Let me verify:
- `writeMode === 'draft_skills'` → `dryRun = false` → mutations happen ✓
- `writeMode !== 'draft_skills'` → `dryRun = true` → dry run, no mutations ✓

The comment says "proposal-only run will actually mutate skills" but that contradicts the code. The reviewer's claim appears to be **incorrect**. However, I should check if there's a semantic mismatch between "proposal-only" and `writeMode !== 'draft_skills'`.

- **Evidence**: Line 742 shows `dryRun: writeMode !== 'draft_skills'`. At line 637, the comment says "In proposal-only runs this is a dry run."
- **Verdict**: The reviewer's claim about inversion is **refuted** — the code is correct. The condition matches the intended semantics.

#### [HIGH] `reconcileStaleActiveSkills` silently swallows all errors

- **Location**: `src/services/skill-improver.ts:468-471`
- **Why it matters**: The entire reconciliation loop for a single skill is wrapped in try/catch that catches everything and does nothing. If `regenerateSkill` fails due to a real error (e.g., knowledge store corruption, disk full, permission denied), the skill is neither regenerated nor retired — it remains active but broken. This creates a silent failure mode where stale skills persist indefinitely.
- **Evidence**: Lines 468-471: `} catch { /* fail-open: skip this skill, continue the sweep. */ }`. No logging, no error tracking, no fallback behavior.
- **Fix direction**: At minimum log the error with context (slug, operation failed). Consider whether "fail-open" should mean "retire on failure" rather than "do nothing."

#### [MEDIUM] Event-sourced outcomes are applied at `generateSkills` time, not persisted

- **Location**: `src/services/skill-generator.ts:547-560`
- **Why it matters**: The PR adds a `.map()` call to apply `effectiveRetrievalOutcomes` onto the pool entries before passing them to downstream logic. This is a **runtime-only** transformation — the underlying knowledge entries are not updated. Any subsequent code path that reads entries directly (not through this function) will see stale outcome counts. This creates a consistency gap: two consumers of the same data see different outcome values depending on whether they go through `generateSkills`.
- **Evidence**: Lines 547-560 create new objects with spread `{ ...e, retrieval_outcomes: effectiveRetrievalOutcomes(...) }` but never persist this back to the knowledge store.
- **Fix direction**: Either persist the merged outcomes back to entries (risking double-count if both paths are used), or document that `effectiveRetrievalOutcomes` must be the universal accessor for outcome data.

#### [MEDIUM] Unparseable frontmatter detection creates a false positive for skills with empty description

- **Location**: `src/services/skill-improver.ts:189-197`
- **Why it matters**: `parseDraftFrontmatter` returning `null` is treated as "unparseable" and the skill is surfaced as stale. However, a skill with valid YAML frontmatter but missing required fields (e.g., no `description` field) might also return `null` depending on implementation. This conflates "no frontmatter at all" with "invalid frontmatter structure."
- **Evidence**: Line 189: `if (!fm) { ... staleActiveSkills.push({ slug, reasons: ['unparseable_frontmatter'] }); }`. The reason string is hardcoded as `unparseable_frontmatter` regardless of whether the issue is missing YAML entirely or malformed structure.
- **Fix direction**: Distinguish between "no frontmatter" and "invalid frontmatter" with different reason codes.

---

- **Risk**: Concurrent event emission from multiple `updateRetrievalOutcome` calls could cause duplicate outcome events in the JSONL log.
  - *Why suspicious*: `recordKnowledgeEvent` is called per id without any deduplication check. If two phases complete simultaneously for the same knowledge entry, two `'outcome'` events are emitted.
  - *What would verify it*: Check if `recordKnowledgeEvent` has any idempotency mechanism (e.g., event_id uniqueness, deduplication on read).

- **Risk**: The additive merge in `effectiveRetrievalOutcomes` could double-count if `updateRetrievalOutcome` is called multiple times for the same phase.
  - *Why suspicious*: Each call appends a new event; `recomputeCounters` folds all events; if the same outcome is emitted twice (e.g., due to retry logic), the rollup counts it twice, and the additive merge adds it again.
  - *What would verify it*: Check for any retry/de-duplication logic in the phase completion path.

- **Risk**: `partial_after_shown_count` handling assumes it's always event-only, but if a future migration writes it to entries, it will be silently dropped by the spread.
  - *Why suspicious*: The comment at line 649 says "partial is event-only (never written to the entry)" — this is an assumption about past behavior, not a guarantee about future behavior.
  - *What would verify it*: Audit all write paths for `partial_after_shown_count`.

---

### 🧪 Test / Coverage Gaps

- **Gap**: No test for `updateRetrievalOutcome` emitting events when the knowledge entry no longer exists in swarm/hive stores.
  - *Evidence*: All tests mock `readKnowledge` to return entries; none test the case where entries are missing (orphaned ids scenario).

- **Gap**: No test for `reconcileStaleActiveSkills` with a skill that has parseable frontmatter but invalid/missing required fields (e.g., no `name` field).
  - *Evidence*: Tests cover "no frontmatter" (`UNPARSEABLE`) and "valid frontmatter" (`HEALTHY`, `SHIM`), but not "malformed frontmatter."

- **Gap**: No test for the high-priority path when `outcomeSignal < 0` (negative track record).
  - *Evidence*: The test at `tests/unit/services/skill-generator.test.ts:710-718` tests negative outcomes, but only for the general case. There's no specific test verifying that the high-priority path cannot resurrect a negative-record entry.

---

### 📋 Shipped-vs-Claimed Gaps

- **Gap**: PR description claims "no migration, no double-count" for the additive merge. However, there is no mechanism to prevent double-count if `updateRetrievalOutcome` is ever called twice for the same outcome (e.g., due to retry). The claim of "no double-count" relies on operational discipline, not structural enforcement.
  - *Evidence*: Line 645-647 in `knowledge-events.ts`: "each outcome lives in exactly one of the two sources — summing counts each exactly once." This is an assumption about call patterns, not a guarantee.

---

### 🔁 Validation Provenance

**CONFIRMED findings that SURVIVED my challenge:**
- CRITICAL: Outcome merge incomplete — supported by exact code at `knowledge-events.ts:638-652`; only two fields are additive, future additions are lost.
- HIGH: Stale-skill reconciliation swallows errors — supported by `skill-improver.ts:468-471`; empty catch block with no logging.
- MEDIUM: Event-sourced outcomes runtime-only — supported by `skill-generator.ts:547-560`; entries are mapped but not persisted.

**CONFIRMED findings that I DROPPED:**
- HIGH (reviewer): DryRun semantics inverted — **refuted**. The code is correct: `writeMode === 'draft_skills'` → `dryRun = false` (mutates), `writeMode !== 'draft_skills'` → `dryRun = true` (reports only). The comment and code align.

- CRITICAL: `updateRetrievalOutcome` emits events for non-existent entries (orphaned outcome events)
- MEDIUM: Unparseable frontmatter detection conflates "no frontmatter" with "invalid structure"

---

### 📝 Merge Recommendation

**[APPROVE_WITH_FIXES]**

The PR successfully addresses the core issues in #1477 (outcome accrual, high-priority path, corrupt baseline, stale-skill detection), but requires fixes for: (1) orphaned outcome event accumulation when entries don't exist, (2) silent error swallowing in reconciliation, and (3) runtime-only outcome transformation creating consistency gaps.

| Check | Result |
|---|---|
| No CRITICAL findings | ❌ (2 critical findings remain) |
| Test coverage adequate | ⚠️ (gaps in orphaned-id and malformed-frontmatter scenarios) |
| No hardcoded secrets | ✅ |
| All async errors handled | ❌ (reconciliation swallows all errors silently) |
| Lockfile consistent | ✅ |

---

zaxbysauce:
## PR Feedback Closure Ledger — Round 1

Commit: `8c37e55` — all 11 confirmed findings addressed.

### Fixed

| ID | Location | Status | Evidence |
|----|----------|--------|----------|
| **F-001** | `skill-improver.ts:473` | **fixed** | Added `import { warn }` from logger; catch block now logs error and pushes slug to `result.skipped` |
| **F-002** | `knowledge-reader.ts:587` | **fixed** | Reordered: `transactShownFile` runs before event emission — no double-emit window on retry |
| **F-003** | `skill-improver.ts:446` | **fixed** | Predicate extended to `!fm \|\| fm.sourceKnowledgeIds.length === 0 \|\| !fm.generatedAt` |
| **F-004** | `knowledge-events.ts:960` | **fixed** | `n_a_count` explicitly excluded via destructure; `partial_after_shown_count` added to `RetrievalOutcome` interface |
| **F-005** | `docs/releases/pending/...` | **fixed** | Added `## Migration steps` and `## Known caveats` sections |
| **F-006** | `docs/knowledge.md`, `docs/skills.md` | **fixed** | Added high-priority maturity path section to knowledge.md; updated maturity gate tiers + `skill_origin: shim` semantics to skills.md |
| **F-007a** | `skill-improver-reconcile.test.ts` | **fixed** | Added `regenerateSkill` to `_internals`; new test stubs it to `{ regenerated: true }` and asserts `result.regenerated` contains the slug |
| **F-007b** | `skill-generator.test.ts` | **fixed** | New integration test seeds a high-priority entry, emits 3 `appendKnowledgeEvent(outcome:'failure')` calls, asserts `selectCandidateEntries` excludes it end-to-end |
| **F-007c** | `knowledge-events.test.ts` | **fixed** | Corrupt-baseline test now also emits a `retrieved` event and asserts `shown_count === 1` survives alongside `succeeded_after_shown_count` |
| **F-011** | `scripts/check-cross-contamination.sh` | **fixed** | Added `knowledge-reader.test.ts\|skill-generator.test.ts\|20\|52\|63` pair — regression alert triggers if co-run pass count drops below 63 |

### Needs user decision (no ship blocker — confirming scope)

| ID | Location | Status | Rationale |
|----|----------|--------|-----------|
| **F-008** | `knowledge-reader.ts:587` | **needs user decision** | Batching all outcome events in a single lock acquisition requires a new `recordKnowledgeEvents()` API. Practical impact is LOW (≤5 IDs typical, cap 50). Worth doing but is an API design change — flagging rather than silently adding. |
| **F-009** | `skill-generator.ts:533` | **confirmed low / accepted** | `readKnowledgeCounterRollups` called twice in the explicit-`sourceKnowledgeIds` path; stat-based in-memory cache makes the second call a near-zero-cost map lookup. No user decision required — documented. |
| **F-010** | `skill-improver.ts:415` | **needs user decision** | `reconcileStaleActiveSkills` has no batch limit. Stale skills are expected to be few in practice. If a batch limit is desired (e.g. 5 per run to match `AUTO_APPLY_BATCH_LIMIT`), that's a policy call. |

### Validation

- `bunx tsc --noEmit` — 0 errors ✓
- `bunx biome ci .` — 0 errors ✓
- `bun --smol test tests/unit/hooks/knowledge-events.test.ts` — 24 pass ✓
- `bun --smol test tests/unit/hooks/knowledge-reader.test.ts` — 20 pass ✓
- `bun --smol test tests/unit/services/skill-generator.test.ts` — 52 pass ✓
- `bun --smol test tests/unit/services/skill-improver-reconcile.test.ts` — 8 pass ✓ (was 7)


---

zaxbysauce:
## Round-2 Review Response

The review was made against `92050ad` (pre-fix commit). `8c37e55` landed at 13:58 with F-001..F-011 fixes. Addressing the four surviving findings:

---

### HIGH: Silent catch block — **already fixed in `8c37e55`**

`skill-improver.ts:479-487` now logs the error with context and pushes the slug to `result.skipped`. Covered by the closure ledger comment.

---

### CRITICAL 1: Outcome merge incomplete — **documented invariant, not a current bug**

`knowledge-events.ts:976-980` already has the explicit documentation:

> "Only these two counts were ever written to the entry; `partial_after_shown_count` is event-only and correctly rides the spread above."

This is a forward maintenance concern. The invariant is: outcome-count fields that were historically written to entries get additive treatment; event-only fields ride the spread. That's currently correct for all three outcome-count fields (`succeeded_after_shown_count`, `failed_after_shown_count`, `partial_after_shown_count`). If a fourth is added, the code needs updating — but that's true of every explicit merge, not a structural gap unique to this code.

---

### CRITICAL 2: Orphaned outcome events — **by design, cannot inflate existing entry counts**

`shownIds` are populated from the shownFile when entries are shown to the LLM. Those IDs were live in the store at show time. The only way to get a truly orphaned event is if an entry is deleted between show and phase-close — a window of one LLM response, essentially impossible in practice.

Even if it did happen: orphaned events accumulate in the JSONL log (bounded by FIFO trim at `MAX_EVENT_LOG_ENTRIES`) and get folded into the baseline under that knowledge ID. But `effectiveRetrievalOutcomes(entry, rollup)` is only ever called when evaluating an existing entry — the orphaned rollup key is never read. The reviewer's claim that this "inflates rollup counts" doesn't hold: the inflated counts are for an ID that no entry possesses, so no maturation decision is affected.

Reintroducing a read-before-emit to filter orphaned IDs would re-add the O(n×stores) read overhead that the event-sourced rewrite was specifically designed to remove, for a risk with zero known real-world impact.

---

### MEDIUM: Runtime-only outcome transformation — **by design**

`skill-generator.ts:547-560` applies `effectiveRetrievalOutcomes` at read time without persisting back. That is the intentional architecture: the knowledge store entries are append-only ground truth; outcome rollups are folded at read time. Persisting merged counts back to entries would create a circular write that double-counts on the next fold. The correct contract is that callers needing merged outcome data call `effectiveRetrievalOutcomes` — the same pattern used consistently throughout `skill-generator.ts` and `knowledge-reader.ts`.

---

CI unit jobs still in progress; will update once they complete.

---

zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `8c37e557b04e`

---

## 🔍 PR Intent

> Reconstructed obligation list (from PR text, issue, commits, changed tests, changed docs, changed interfaces):

- **O-001**: Fix outcome accrual root cause — `effectiveRetrievalOutcomes` must additively merge stored entry counts with event-derived rollup counts instead of overwriting
- **O-002**: Wire up outcome events — `updateRetrievalOutcome` must emit immutable `'outcome'` KnowledgeEvents as the single source of truth instead of mutating entries in-place
- **O-003**: Add high-priority directive maturity path — `critical`/`high` directives confirmed in ≥1 phase with confidence ≥0.6 should satisfy the maturity gate
- **O-004**: Add corrupt-baseline resilience — `readCounterBaseline` must fail open scoped to the baseline without discarding the live event log
- **O-005**: Close stale-skill detection blind spot — unparseable frontmatter SKILL.md files must be surfaced as stale, not silently skipped
- **O-006**: Add stale-skill reconciliation — `reconcileStaleActiveSkills` must regenerate or retire stale generated skills
- **O-007**: Mark hand-authored shims with `skill_origin: shim` so they are skipped by reconciliation
- **O-008**: Convert committed `guardrail-patterns` to proper generated frontmatter with `source_knowledge_ids`
- **O-009**: Update documentation to reflect new maturity path, outcome merge behavior, and shim convention
- **O-010**: Add/update tests covering all changed behavior (outcome events, additive merge, high-priority path, reconciliation)

---

## 📦 Implementation Summary

The PR performs a paradigm shift in outcome attribution: instead of mutating `entry.retrieval_outcomes` in-place via `transactKnowledge`, the new code emits immutable `'outcome'` events to `.swarm/knowledge-events.jsonl`. These events are folded by `recomputeCounters` into per-entry rollups, which are then merged with historical entry counts using an additive strategy (preventing the zero-overwrite bug).

The skill maturity gate (`isSkillMaturityEligible`) gains a new high-priority path for `critical`/`high` directives that requires only 1 distinct phase confirmation and 0.6 confidence instead of the standard 2-phase/0.7 requirement.

A new `reconcileStaleActiveSkills` function is added to detect and repair stale SKILL.md files (unparseable frontmatter, missing source IDs, missing generated_at). It regenerates skills when possible or retires them otherwise. Hand-authored shims marked with `skill_origin: shim` are explicitly skipped.

The `readCounterBaseline` function gains a try/catch that fails open on JSON parse errors instead of propagating to the outer catch that would discard the live event log.

---

|---|---|---|
| O-001 | `SUPPORTED` | `src/hooks/knowledge-events.ts:957-982` — additive merge with explicit `(base + rollup)` for outcome counts |
| O-002 | `SUPPORTED` | `src/hooks/knowledge-reader.ts:594-601` — emits `'outcome'` events via `_internals.recordKnowledgeEvent` |
| O-003 | `SUPPORTED` | `src/services/skill-generator.ts:185-210` — high-priority path checks `isHighPriorityDirective`, `distinctPhases >= 1`, `confidence >= HIGH_PRIORITY_SKILL_MIN_CONFIDENCE` |
| O-004 | `SUPPORTED` | `src/hooks/knowledge-events.ts:626-643` — try/catch around JSON.parse, returns empty Map on error with warn() |
| O-005 | `SUPPORTED` | `src/services/skill-improver.ts:192-197` — catches `!fm` and pushes `{ reasons: ['unparseable_frontmatter'] }` |
| O-006 | `SUPPORTED` | `src/services/skill-improver.ts:397-583` — full `reconcileStaleActiveSkills` implementation with regenerate/retire logic |
| O-007 | `SUPPORTED` | `src/services/skill-improver.ts:198-200, 468-470` — both `gatherInventory` and `reconcileStaleActiveSkills` skip `skillOrigin === 'shim'` |
| O-008 | `SUPPORTED` | `.opencode/skills/generated/guardrail-patterns/SKILL.md:1-17` — proper frontmatter with all 4 source_knowledge_ids |
| O-009 | `SUPPORTED` | `docs/knowledge.md:432-461`, `docs/skills.md:112-152`, `docs/releases/pending/1477-skill-maturity-outcome-accrual.md` — all updated |
| O-010 | `SUPPORTED` | `tests/unit/hooks/knowledge-events.test.ts:245-289, 481-567` (24 tests), `tests/unit/services/skill-generator.test.ts:667-764` (high-priority path), `tests/unit/services/skill-improver-reconcile.test.ts` (7 tests) |

---

## 🚨 Confirmed Findings

### [HIGH] `_internals.recordKnowledgeEvent` import is not through the DI seam in knowledge-reader.ts

- **Location**: `src/hooks/knowledge-reader.ts:11`
- **Why it matters**: The `_internals` export in `knowledge-reader.ts` includes `recordKnowledgeEvent` (line 683), but the function itself imports `recordKnowledgeEvent` directly from `./knowledge-events.js` (line 11). This means if `knowledge-events.ts` is mocked or replaced, the `_internals` seam becomes inconsistent — the exported function would use the real implementation while test mocks might target a different path.
- **Evidence**: Line 11 imports `recordKnowledgeEvent` from `'./knowledge-events.js'`, then line 683 exports it via `_internals`. The test at `tests/unit/hooks/knowledge-reader.test.ts:666-668` spies on `readerInternals.recordKnowledgeEvent`, but if the module mock doesn't cover the direct import path, the spy won't intercept calls.
- **Fix direction**: Either import through `_internals` within knowledge-reader itself, or ensure all tests mock `knowledge-events.js` directly (which they currently do not — see `knowledge-reader-key-normalization.test.ts:78` which only mocks `knowledge-store.js`).

### [HIGH] `effectiveRetrievalOutcomes` called with potentially undefined `retrieval_outcomes` in generateSkills

- **Location**: `src/services/skill-generator.ts:553-561`
- **Why it matters**: The mapping calls `effectiveRetrievalOutcomes(e.retrieval_outcomes, rollups.get(e.id))`. If any entry `e` has a missing or undefined `retrieval_outcomes` field (possible for legacy entries or corrupted data), the function receives `undefined` as the first argument. Inside `effectiveRetrievalOutcomes`, it destructures `base.succeeded_after_shown_count` which will throw a TypeError if `base` is undefined.
- **Evidence**: `skill-generator.ts:558` — `.map((e) => ({ ...e, retrieval_outcomes: effectiveRetrievalOutcomes(e.retrieval_outcomes, rollups.get(e.id)) }))`. No null-check on `e.retrieval_outcomes` before passing it.
- **Fix direction**: Add defensive check: `effectiveRetrievalOutcomes(e.retrieval_outcomes ?? {}, rollups.get(e.id))` or guard inside `effectiveRetrievalOutcomes` with `if (!base) return rollup ?? defaultBase`.

### [MEDIUM] Dry-run result naming is misleading — `result.retired` used for non-mutated candidates

- **Location**: `src/services/skill-improver.ts:510-512`
- **Why it matters**: In dry-run mode (line 510), stale skills are pushed to `result.retired` even though nothing is actually retired. The name `retired` implies mutation happened. This is intentional per the comment at line 748-750, but the naming creates confusion for callers who inspect the result.
- **Evidence**: `skill-improver.ts:510-512` — `if (options.dryRun) { result.retired.push(skill.slug); continue; }`. The comment at line 748 says "reports retire-candidates via a dry run" but doesn't clarify the naming convention.
- **Fix direction**: Either rename to `result.wouldRetire` in dry-run mode, or add a separate field like `result.dryRunRetireCandidates`. Alternatively, document this behavior explicitly in the interface.

### [MEDIUM] Test file `knowledge-reader-key-normalization.test.ts` does not mock `knowledge-events.js`

- **Location**: `tests/unit/hooks/knowledge-reader-key-normalization.test.ts:78`
- **Why it matters**: The test mocks `knowledge-store.js` (line 78) but does NOT mock `knowledge-events.js`. Since `updateRetrievalOutcome` now emits outcome events via `_internals.recordKnowledgeEvent` which calls the real `recordKnowledgeEvent` from `knowledge-events.js`, the test relies on the actual event file being created and read. This is fragile — if the `.swarm/knowledge-events.jsonl` file doesn't exist or has stale content, the test will fail unpredictably.
- **Evidence**: The mock block at line 78 only covers `knowledge-store.js`. The helper function `readOutcomeEvents()` at line 124 reads from the real filesystem path `path.join(tmpDir, '.swarm', 'knowledge-events.jsonl')`. No mock ensures this file exists or is clean between tests.
- **Fix direction**: Either mock `_internals.recordKnowledgeEvent` via the DI seam (like `knowledge-reader.test.ts:666-668` does), or ensure the test creates and cleans up the events file in beforeEach/afterEach.

### [LOW] `n_a_count` exclusion comment references "F-004" which is not a standard reference format

- **Location**: `src/hooks/knowledge-events.ts:962`
- **Why it matters**: The comment references "F-004" as if it's a tracked finding or requirement, but this format doesn't appear anywhere in the codebase or PR description. It may confuse future reviewers.
- **Evidence**: `knowledge-events.ts:962` — `// eslint-disable-next-line @typescript-eslint/no-unused-vars\n  const { n_a_count: _na, ...rollupWithoutNa } = rollup;` followed by comment referencing "F-004".
- **Fix direction**: Remove the F-004 reference or replace with a meaningful explanation of why `n_a_count` must be excluded.

---

- **Risk**: Cross-file test isolation leak from `knowledge-reader.test.ts` to `skill-generator.test.ts` may manifest differently under different Bun versions.
  - *Why suspicious*: The PR notes this is pre-existing, but the new `_internals.recordKnowledgeEvent` spy in `knowledge-reader.test.ts:666-668` uses a mutable global assignment that could leak if Bun's `mock.module` isolation changes.
  - *What would verify it*: Run both test files together on multiple Bun versions and observe failure patterns.

- **Risk**: `effectiveRetrievalOutcomes` additive merge may double-count if an entry has BOTH event-sourced outcomes AND legacy inline counts that were written by a concurrent writer before the migration.
  - *Why suspicious*: The PR claims "each outcome lives in exactly one of the two sources" but this assumes no overlap exists in production data. If a migration window had partial writes, both sources could contain the same outcome.
  - *What would verify it*: Audit production knowledge entries for entries that have non-zero values in BOTH `retrieval_outcomes.succeeded_after_shown_count` AND event log entries with type `'outcome'`.

---

## 🧪 Test / Coverage Gaps

- **Gap**: No test verifies that `effectiveRetrievalOutcomes` handles `undefined` base gracefully (when an entry lacks `retrieval_outcomes`).
  - *Evidence*: All tests pass a defined object as the first argument. The production code path in `generateSkills:558` does not guard against undefined.

- **Gap**: No test for `reconcileStaleActiveSkills` when `listSkills` throws an error.
  - *Evidence*: `skill-improver.ts:472-475` catches the error and returns empty result, but no test exercises this path.

- **Gap**: No test verifies that `readCounterBaseline` corrupt-baseline recovery doesn't affect subsequent calls (idempotency of the "ignore corrupt file" behavior).
  - *Evidence*: Only one test exists at `knowledge-events.test.ts:245-289`. It tests a single call but doesn't verify that subsequent reads also succeed with the same corrupt file.

---

## 📋 Shipped-vs-Claimed Gaps

- **Gap**: PR description claims "All 102 tests pass in isolation" but the test count in the diff shows approximately 96 tests across all files (not 102). The discrepancy may be due to pre-existing tests not counted, or the claim is inaccurate.
  - *Evidence*: Counting `it(` and `describe(` blocks across changed test files: knowledge-events.test.ts has ~24 new+modified, knowledge-reader.test.ts has ~15 modified, skill-generator.test.ts has ~30 new+modified, skill-improver-reconcile.test.ts has 7 new. Total ≈ 76 new/modified tests, not 102.

- **Gap**: The `updateRetrievalOutcome` function now calls `_internals.recordKnowledgeEvent` which is imported from `./knowledge-events.js`, but the test at `knowledge-reader-key-normalization.test.ts` does NOT mock this path. This means the test relies on real filesystem I/O for outcome event verification, which contradicts the claim that tests use "proper mocking via `_internals` seam."
  - *Evidence*: `knowledge-reader-key-normalization.test.ts:78` only mocks `knowledge-store.js`, not `knowledge-events.js`.

---

## 📝 Merge Recommendation

**[APPROVE_WITH_FIXES]**

The PR successfully addresses all core obligations from issue #1477: outcome accrual is fixed via additive merge, outcome events are properly wired, high-priority maturity path is implemented, corrupt-baseline resilience is added, and stale-skill reconciliation is functional. However, two HIGH-severity issues must be addressed before merge:

1. **`effectiveRetrievalOutcomes` must guard against undefined `base` argument** — a single corrupted entry would crash the entire skill generation pass.
2. **`knowledge-reader-key-normalization.test.ts` must mock `_internals.recordKnowledgeEvent`** — relying on real filesystem I/O for outcome event verification is fragile and inconsistent with the project's mocking pattern (as correctly done in `knowledge-reader.test.ts`).

The MEDIUM findings (naming convention, DI seam import) should be addressed in follow-up but are not blocking.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| Test coverage adequate | ⚠️ (see gaps above) |
| No hardcoded secrets | ✅ |
| Lockfile consistent | ✅ |

---

zaxbysauce:
## Round-2 Review Response (commit `8c37e55`)

All 16 checks are green and the PR is in the merge queue. Triaged the round-2 findings against the current code — **no code changes warranted**; both HIGH items are misreads.

### HIGH — `effectiveRetrievalOutcomes` crashes on undefined `base` → **refuted**
`knowledge-events.ts:954` is `const base = stored ?? { applied_count: 0, succeeded_after_count: 0, failed_after_count: 0 }`. `base` is never `undefined`, so the destructure cannot throw. The additive merge also uses `(base.succeeded_after_shown_count ?? 0)`, so even the missing-field case resolves to `0`. The cited crash path does not exist (the `??` default sits one line above the destructure).

### HIGH — `recordKnowledgeEvent` not routed through the DI seam → **refuted**
`updateRetrievalOutcome` calls `_internals.recordKnowledgeEvent` (`knowledge-reader.ts:595`) — that **is** the seam. The round-1 tests (`knowledge-reader.test.ts:666-668`) override exactly that property and intercept the calls. The module-level `import` only supplies the default seam value; overriding `_internals.recordKnowledgeEvent` redirects all call sites.

### MEDIUM — key-normalization test doesn't mock `knowledge-events.js` → **by design**
The test uses a per-test `mkdtemp` tmpdir and reads the real event log from that isolated path — the repo's established pattern (`writing-tests` skill: temp dirs, no cross-test bleed). Verified locally: passes in isolation (13/13).

### MEDIUM — dry-run pushes to `result.retired` → **by design / documented**
Intentional and noted in the code comment ("Report-only: surface the slug without mutating anything"). Cosmetic naming only; no mutation occurs in dry-run.

### LOW — `F-004` comment reference is "non-standard" → **factually incorrect**
The `F-0xx` comment convention is used in ~18 places across the repo (`phase-complete.ts`, `semgrep.ts`, `knowledge-application.ts`, `delegation-gate.ts`, `lang/detector.ts`, …). It's an established convention, not a one-off.

### Validation (local reproduction of the exact per-file CI invocation)
- 5 affected files in isolation: 24 + 20 + 13 + 8 + 52 = **117 pass, 0 fail**
- `tsc --noEmit` — 0 errors · `biome ci .` — 0 errors
- cross-contamination guard — exit 0, both pairs at tracked baselines (33, 63), no new regression
- CI: all 16 checks green, `mergeable_state: clean`

---

PR #1542, opened 2026-06-27
feat(tools): add bounded external cli wrappers

zaxbysauce:
## Summary
- Harden `search` through a bounded shared external-tool runner and safer fallback traversal.
- Add read-only `ast_grep`, `actionlint_scan`, `osv_scan`, and `gh_evidence` wrappers with lazy binary resolution, structured missing-binary results, and bounded execution.
- Wire tool metadata/manifest/exports/docs/release notes and focused regression coverage for subprocess contracts, truncation, malformed output, registration coherence, and PR-feedback timeout cleanup.

## Invariant audit
- 1 (plugin init): touched - no new init-time probes; external binaries resolve only inside tool execution. Verified `node scripts\repro-704.mjs` PASS after sandbox escalation: T1 225.2ms/400ms, T2 85.6ms, T3 94.7ms. The harness printed all PASS results, then required manual cleanup of the lingering validation process.
- 2 (runtime portability): touched - external runner uses the existing Bun compatibility layer and array-form subprocess calls; verified `bun run build` PASS and `node --input-type=module -e "await import('./dist/index.js')"` PASS.
- 3 (subprocesses): touched - `src/utils/external-tool-runner.ts` centralizes explicit absolute `cwd`, `stdin: 'ignore'`, timeout, bounded stdout/stderr, Windows-safe timeout kill signal, and best-effort kill. Covered by `tests/unit/utils/external-tool-runner.test.ts`, focused tool tests, and subprocess grep of the only runner spawn call.
- 4 (.swarm containment): not touched - tools read under injected workspace directories; PR evidence files are runtime artifacts under `.swarm/` and are not committed.
- 5 (plan durability): not touched - no ledger, projection, checkpoint, or plan schema changes.
- 6 (test_runner safety): not touched - validation used shell commands, not broad OpenCode `test_runner` scopes.
- 7 (test writing): touched - tests use `bun:test`, temp directories, and `_internals` seams restored after each test; verified focused tool/runner tests PASS.
- 8 (session state): not touched - no session/global state changes.
- 9 (guardrails/retry): not touched - no guardrail or retry logic changes.
- 10 (chat/system msg): not touched - no chat or system-message hook changes.
- 11 (tool registration): touched - updated metadata, manifest, exports, and docs; verified `bun run scripts\check-tool-registration.ts` PASS with 104 coherent tools, registration test set PASS, drift check PASS, and config map subset PASS.
- 12 (release/cache): touched - added `docs/releases/pending/search-ripgrep-runner-hardening.md`; no version, changelog, release manifest, or cache deletion changes.

## Test plan
- PASS `bunx @biomejs/biome@2.3.14 ci .`
- PASS `bun run typecheck`
- PASS `bun --smol test tests\unit\utils\external-tool-runner.test.ts src\tools\ast-grep.test.ts src\tools\external-cli-tools.test.ts src\tools\search.test.ts src\tools\search.adversarial.test.ts --timeout 60000` (155 pass)
- PASS `bun --smol test src\tools\search.test.ts src\tools\search.adversarial.test.ts src\tools\ast-grep.test.ts src\tools\external-cli-tools.test.ts tests\unit\utils\external-tool-runner.test.ts --timeout 60000` (153 pass, prior head before feedback tests were added)
- PASS `bun --smol test tests\unit\tools\manifest-parity.test.ts tests\unit\tools\registration-smoke.test.ts tests\unit\tools\tool-registration-conformance.test.ts tests\unit\tools\tool-names.test.ts src\tools\plugin-registration-adversarial.test.ts --timeout 60000` (50 pass)
- PASS `bun run build`
- PASS `node --input-type=module -e "await import('./dist/index.js'); console.log('dist import OK')"`
- PASS `node scripts\repro-704.mjs` after sandbox escalation (T1 225.2ms/400ms, T2 85.6ms, T3 94.7ms; harness printed PASS and then needed manual cleanup of the lingering validation process)
- PASS `git diff --check`
- PASS `Select-String -Path src\utils\external-tool-runner.ts -Pattern 'bunSpawn\(|spawn\(|spawnSync\(' -Context 0,8` confirmed the runner spawn has explicit `cwd`, `stdin: 'ignore'`, bounded pipes, and timeout.
- PASS `bun run scripts\check-tool-registration.ts`
- PASS `bun run drift:check`
- PASS `bun --smol test tests\unit\config\agent-tool-map.test.ts tests\unit\config\critic-registration.test.ts tests\unit\config\constants.test.ts --timeout 60000` (48 pass)
- PASS `bun --smol test tests\unit\config\quiet-config.test.ts --timeout 60000` with sandbox escalation (19 pass)
- KNOWN BASELINE FAIL `bun --smol test tests\unit\config --timeout 120000` with sandbox escalation: 1464 pass, 1 fail in `ExecutionProfileSchema > defaults > parses empty object with all defaults` (`max_concurrent_tasks` expected 1, received 10). Reproduced on clean `origin/main` at `dc539021e34a1f8797abe7441ff6ea905a7f29e1` after `bun install --frozen-lockfile` in `C:\tmp\opencode-swarm-main-check`.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -565,18 +565,32 @@ Added as the default structured search path for workspace pattern lookup. Replac
 
 | Property | Type | Default | Description |
 |----------|------|---------|-------------|
-| `pattern` | `string` | required | Search pattern (literal or regex) |
-| `workspace` | `string` | required | Root directory to search |
+| `query` | `string` | required | Search query (literal or regex depending on `mode`) |
 | `mode` | `"literal" \| "regex"` | `"literal"` | Match mode |
-| `glob` | `string[]` | `[]` | Include globs (e.g. `["**/*.ts"]`) |
-| `exclude` | `string[]` | `[]` | Exclude globs |
-| `maxResults` | `number` | `100` | Hard cap on returned matches |
-| `maxLines` | `number` | `10` | Max lines per match snippet |
-| `context` | `number` | `0` | Surrounding lines per match (only when > 0) |
+| `include` | `string` | unset | Comma-separated include globs (e.g. `"src/**/*.ts"`) |
+| `exclude` | `string` | unset | Comma-separated exclude globs |
+| `max_results` | `number` | `100` | Hard cap on returned matches |
+| `max_lines` | `number` | `200` | Max characters per matched line |
 
-Returns structured JSON with `file`, `line`, `text`, and `truncated` fields per hit. Falls back to a graceful error when ripgrep is unavailable. **Not a structural AST search** — combine with `symbols` and `imports` for full module analysis.
+Returns structured JSON with normalized match metadata. **Not a structural AST search** — use `ast_grep` for syntax-aware pattern matching, or combine with `symbols` and `imports` for full module analysis.
 
 **Registered for**: architect, coder, reviewer, explorer, test_engineer.
+
+Current output includes `matches[].file`, `matches[].lineNumber`, `matches[].lineText`, `truncated`, `total`, `query`, `mode`, `maxResults`, and `engine`. The ripgrep path uses a bounded subprocess runner with explicit `cwd`, ignored stdin, timeout, capped stdout/stderr, and cleanup. Its argv places options and globs before `--`, then the query and `.` path operand, so dash-prefixed queries cannot become flags. If ripgrep is unavailable, the fallback engine performs bounded filesystem traversal, skips heavy runtime directories such as `.git`, `.swarm`, `node_modules`, build outputs, and coverage directories, and discloses that it does not fully emulate ripgrep gitignore semantics.
+
+### Structural AST Search Tool - `ast_grep`
+
+Read-only structural search using the ast-grep CLI. The tool resolves `ast-grep` or `sg` lazily at call time, never during plugin initialization, and runs `ast-grep run --pattern <pattern> --json=stream` through the bounded external-tool runner. It supports optional `language`, comma-separated `include` and `exclude` globs, and `max_results`. Output normalizes ast-grep's JSON stream into workspace-relative `matches[]` entries with 1-based line and column numbers. It does not rewrite files.
+
+### External QA CLI Tools
+
+Three read-only wrappers expose high-yield external project checks without adding plugin-init probes or package dependencies:
+
+- `actionlint_scan` runs `actionlint -format '{{json .}}'` against `.github/workflows/**/*.yml,yaml` or explicit workspace-relative workflow files, returning structured findings.
+- `osv_scan` runs `osv-scanner scan --format json <path>` against a workspace-relative path and normalizes OSV vulnerability records.
+- `gh_evidence` runs `gh pr view` or `gh issue view` with a bounded JSON field allowlist for review and CI evidence.
+
+All three tools resolve their executable lazily at call time and share the bounded external-tool runner: explicit `cwd`, ignored stdin, timeout, capped stdout/stderr, and best-effort cleanup. Missing binaries return structured guidance instead of failing plugin registration.
 
 ### Reviewer-Safe Patch Suggestion Tool — `suggest_patch` (v6.45.0)
 
--- a/docs/releases/pending/search-ripgrep-runner-hardening.md
+++ b/docs/releases/pending/search-ripgrep-runner-hardening.md
@@ -1,0 +1,7 @@
+# Search ripgrep runner hardening
+
+Hardened the `search` tool's ripgrep execution path so it uses an explicit working directory, ignored stdin, bounded stdout/stderr, a concrete timeout, and best-effort cleanup through a shared external-tool runner.
+
+The Node fallback search now uses bounded traversal, default skips for heavy runtime directories, realpath-based containment checks, and a warning that fallback behavior does not fully emulate ripgrep gitignore semantics.
+
+Added read-only `ast_grep`, `actionlint_scan`, `osv_scan`, and `gh_evidence` tools that run structural search, GitHub Actions linting, OSV dependency scanning, and bounded GitHub evidence collection through the same bounded external-tool runner, with lazy binary resolution, workspace-relative output normalization where applicable, and structured missing-binary guidance.
--- a/src/tools/actionlint-scan.ts
+++ b/src/tools/actionlint-scan.ts
@@ -1,0 +1,377 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import type { ToolDefinition } from '@opencode-ai/plugin/tool';
+import { z } from 'zod';
+import {
+	resolveExecutableFromPath,
+	runExternalTool,
+} from '../utils/external-tool-runner';
+import {
+	containsControlChars,
+	containsPathTraversal,
+} from '../utils/path-security';
+import { createSwarmTool } from './create-tool';
+
+const ACTIONLINT_TIMEOUT_MS = 15_000;
+const ACTIONLINT_MAX_STDOUT_BYTES = 4 * 1024 * 1024;
+const ACTIONLINT_MAX_STDERR_BYTES = 128 * 1024;
+const DEFAULT_MAX_RESULTS = 200;
+const HARD_CAP_RESULTS = 2_000;
+const MAX_WORKFLOW_FILES = 500;
+const MAX_WORKFLOW_DIRS = 200;
+const MAX_WORKFLOW_DEPTH = 20;
+
+interface ActionlintFinding {
+	file: string;
+	line: number;
+	column: number;
+	kind: string;
+	message: string;
+	endColumn?: number;
+}
+
+interface ActionlintResult {
+	clean: boolean;
+	findings: ActionlintFinding[];
+	total: number;
+	truncated: boolean;
+	files: string[];
+	command: string[];
+	outputTruncated?: boolean;
+	note?: string;
+}
+
+interface WorkflowDiscoveryResult {
+	files: string[];
+	truncated: boolean;
+}
+
+interface ActionlintError {
+	error: true;
+	type:
+		| 'actionlint-not-found'
+		| 'invalid-input'
+		| 'path-escape'
+		| 'timeout'
+		| 'unknown';
+	message: string;
+}
+
+function resolveActionlintBinary(): string | null {
+	return _internals.resolveExecutableFromPath(['actionlint']);
+}
+
+function normalizeRelativeWorkflowFile(
+	file: string,
+	workspace: string,
+): string | null {
+	if (!file || containsControlChars(file) || containsPathTraversal(file)) {
+		return null;
+	}
+	if (path.isAbsolute(file) || /^[A-Za-z]:[/\\]/.test(file)) {
+		return null;
+	}
+	const normalizedInput = file
+		.split(/[\\/]+/)
+		.filter(Boolean)
+		.join(path.sep);
+	if (!/\.(ya?ml)$/i.test(normalizedInput)) {
+		return null;
+	}
+	try {
+		const realWorkspace = fs.realpathSync(workspace);
+		const resolved = path.resolve(realWorkspace, normalizedInput);
+		const realFile = fs.realpathSync(resolved);
+		const relative = path.relative(realWorkspace, realFile);
+		if (!relative || relative.startsWith('..') || path.isAbsolute(relative)) {
+			return null;
+		}
+		return relative.split(path.sep).join('/');
+	} catch {
+		return null;
+	}
+}
+
+function discoverWorkflowFiles(workspace: string): WorkflowDiscoveryResult {
+	const root = path.join(workspace, '.github', 'workflows');
+	if (!fs.existsSync(root)) return { files: [], truncated: false };
+	const files: string[] = [];
+	let dirCount = 0;
+	let truncated = false;
+
+	const visit = (dir: string, depth = 0) => {
+		if (
+			files.length >= MAX_WORKFLOW_FILES ||
+			dirCount >= MAX_WORKFLOW_DIRS ||
+			depth > MAX_WORKFLOW_DEPTH
+		) {
+			truncated = true;
+			return;
+		}
+		dirCount++;
+		let entries: fs.Dirent[];
+		try {
+			entries = fs.readdirSync(dir, { withFileTypes: true });
+		} catch {
+			return;
+		}
+		for (const entry of entries) {
+			if (files.length >= MAX_WORKFLOW_FILES || dirCount >= MAX_WORKFLOW_DIRS) {
+				truncated = true;
+				break;
+			}
+			const full = path.join(dir, entry.name);
+			if (entry.isDirectory()) {
+				visit(full, depth + 1);
+			} else if (entry.isFile() && /\.(ya?ml)$/i.test(entry.name)) {
+				const rel = normalizeRelativeWorkflowFile(
+					path.relative(workspace, full),
+					workspace,
+				);
+				if (rel) files.push(rel);
+			}
+		}
+	};
+
+	visit(root);
+	return { files, truncated };
+}
+
+function parseActionlintOutput(
+	stdout: string,
+	workspace: string,
+	maxResults: number,
+): { findings: ActionlintFinding[]; total: number } {
+	const findings: ActionlintFinding[] = [];
+	let total = 0;
+
+	for (const line of stdout.split('\n')) {
+		if (!line.trim()) continue;
+		try {
+			const entry = JSON.parse(line);
+			const file =
+				typeof entry.filepath === 'string'
+					? normalizeRelativeWorkflowFile(
+							path.isAbsolute(entry.filepath)
+								? path.relative(workspace, entry.filepath)
+								: entry.filepath,
+							workspace,
+						)
+					: null;
+			if (!file) continue;
+			total++;
+			if (findings.length >= maxResults) continue;
+			findings.push({
+				file,
+				line: typeof entry.line === 'number' ? entry.line : 1,
+				column: typeof entry.column === 'number' ? entry.column : 1,
+				kind: typeof entry.kind === 'string' ? entry.kind : 'unknown',
+				message: typeof entry.message === 'string' ? entry.message : '',
+				endColumn:
+					typeof entry.end_column === 'number' ? entry.end_column : undefined,
+			});
+		} catch {
+			// Ignore malformed lines from mixed stdout.
+		}
+	}
+
+	return { findings, total };
+}
+
+function sanitizeMaxResults(value: unknown): number {
+	const numeric =
+		typeof value === 'number' && Number.isFinite(value)
+			? value
+			: DEFAULT_MAX_RESULTS;
+	return Math.min(Math.max(0, numeric), HARD_CAP_RESULTS);
+}
+
+export const actionlint_scan: ToolDefinition = createSwarmTool({
+	description:
+		'Run actionlint against GitHub Actions workflow YAML files and return structured findings. Read-only; resolves actionlint lazily.',
+	args: {
+		files: z
+			.array(z.string())
+			.optional()
+			.describe(
+				'Optional workspace-relative workflow YAML files. Defaults to .github/workflows/**/*.yml,yaml.',
+			),
+		max_results: z
+			.number()
+			.default(DEFAULT_MAX_RESULTS)
+			.describe('Maximum findings to return'),
+	},
+	execute: async (args: unknown, directory: string) => {
+		const obj = (
+			typeof args === 'object' && args !== null ? args : {}
+		) as Record<string, unknown>;
+		const maxResults = sanitizeMaxResults(obj.max_results);
+		const requestedFiles = Array.isArray(obj.files) ? obj.files : undefined;
+		if (requestedFiles?.some((f) => typeof f !== 'string')) {
+			return JSON.stringify(
+				{
+					error: true,
+					type: 'invalid-input',
+					message: 'files must be an array of strings',
+				} satisfies ActionlintError,
+				null,
+				2,
+			);
+		}
+
+		const discovery = requestedFiles
+			? {
+					files: requestedFiles
+						.map((f) => normalizeRelativeWorkflowFile(f, directory))
+						.filter((f): f is string => Boolean(f)),
+					truncated: false,
+				}
+			: _internals.discoverWorkflowFiles(directory);
+		let files = discovery.files;
+
+		if (requestedFiles && files.length !== requestedFiles.length) {
+			return JSON.stringify(
+				{
+					error: true,
+					type: 'path-escape',
+					message:
+						'files must be existing workspace-relative GitHub Actions YAML files',
+				} satisfies ActionlintError,
+				null,
+				2,
+			);
+		}
+		const uniqueFiles = Array.from(new Set(files));
+		const cappedFiles = uniqueFiles.slice(0, MAX_WORKFLOW_FILES);
+		const discoveryTruncated =
+			discovery.truncated || uniqueFiles.length > cappedFiles.length;
+		files = cappedFiles;
+
+		if (files.length === 0) {
+			return JSON.stringify(
+				{
+					clean: !discoveryTruncated,
+					findings: [],
+					total: 0,
+					truncated: discoveryTruncated,
+					files: [],
+					command: ['actionlint', '-format', '{{json .}}'],
+					note: discoveryTruncated
+						? `Workflow discovery was truncated at ${MAX_WORKFLOW_FILES} files, ${MAX_WORKFLOW_DIRS} directories, or depth ${MAX_WORKFLOW_DEPTH}; result is incomplete.`
+						: 'No GitHub Actions workflow YAML files found',
+				} satisfies ActionlintResult,
+				null,
+				2,
+			);
+		}
+
+		const executable = _internals.resolveActionlintBinary();
+		if (!executable) {
+			return JSON.stringify(
+				{
+					error: true,
+					type: 'actionlint-not-found',
+					message:
+						'actionlint executable not found. Install actionlint and ensure it is on PATH.',
+				} satisfies ActionlintError,
+				null,
+				2,
+			);
+		}
+
+		const lintArgs = [
+			'-format',
+			'{{json .}}',
+			...files.map((file) => `./${file}`),
+		];
+		const run = await _internals.runExternalTool({
+			executable,
+			args: lintArgs,
+			cwd: directory,
+			timeoutMs: ACTIONLINT_TIMEOUT_MS,
+			maxStdoutBytes: ACTIONLINT_MAX_STDOUT_BYTES,
+			maxStderrBytes: ACTIONLINT_MAX_STDERR_BYTES,
+		});
+
+		if (run.status === 'timeout') {
+			return JSON.stringify(
+				{
+					error: true,
+					type: 'timeout',
+					message: `actionlint timed out after ${ACTIONLINT_TIMEOUT_MS}ms`,
+				} satisfies ActionlintError,
+				null,
+				2,
+			);
+		}
+		if (run.status === 'spawn-error') {
+			return JSON.stringify(
+				{
+					error: true,
+					type: 'unknown',
+					message: run.message ?? 'actionlint failed to start',
+				} satisfies ActionlintError,
+				null,
+				2,
+			);
+		}
+
+		const { findings, total } = parseActionlintOutput(
+			run.stdout,
+			directory,
+			maxResults,
+		);
+		const hardError = run.exitCode !== 0 && findings.length === 0 && run.stderr;
+		if (hardError) {
+			return JSON.stringify(
+				{
+					error: true,
+					type: 'unknown',
+					message: run.stderr.split('\n')[0],
+				} satisfies ActionlintError,
+				null,
+				2,
+			);
+		}
+
+		return JSON.stringify(
+			{
+				findings,
+				total,
+				truncated:
+					discoveryTruncated ||
+					total > maxResults ||
+					run.stdoutTruncated ||
+					run.stderrTruncated,
+				files,
+				command: ['actionlint', ...lintArgs],
+				outputTruncated: run.stdoutTruncated || run.stderrTruncated,
+				clean:
+					findings.length === 0 &&
+					total === 0 &&
+					!discoveryTruncated &&
+					!run.stdoutTruncated &&
+					!run.stderrTruncated,
+				note: discoveryTruncated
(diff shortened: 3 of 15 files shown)


zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `c912372314cd`

---

# PR Review — opencode-swarm #1542: feat(tools): add bounded external cli wrappers

---

## 🔍 PR Intent

- **O-001** Harden `search` tool's ripgrep execution path with explicit working directory, ignored stdin, bounded stdout/stderr, concrete timeout, and best-effort cleanup via a shared external-tool runner.
- **O-002** Add Node fallback search with bounded traversal, default skips for heavy runtime directories, realpath-based containment checks, and a warning about incomplete gitignore emulation.
- **O-003** Add read-only `ast_grep`, `actionlint_scan`, `osv_scan`, and `gh_evidence` tools that run structural search, GitHub Actions linting, OSV dependency scanning, and bounded GitHub evidence collection through the same bounded external-tool runner.
- **O-004** All new tools must resolve their executable lazily at call time (never during plugin initialization) and return structured missing-binary guidance instead of failing registration.
- **O-005** Wire tool metadata/manifest/exports/docs/release notes for the four new tools plus hardened search.
- **O-006** Add focused regression coverage for subprocess contracts, truncation, malformed output, and registration coherence.
- **O-007** Maintain plugin-init invariant: no new init-time probes; external binaries resolve only inside tool execution.

---

## 📦 Implementation Summary

The PR introduces a shared `external-tool-runner.ts` utility that centralizes bounded subprocess execution for all external CLI tools. It enforces:
- Explicit absolute `cwd` (validated before spawn)
- `stdin: 'ignore'` to prevent interactive hangs
- Configurable timeout with `SIGKILL` termination
- Bounded stdout/stderr byte limits with truncation tracking
- Best-effort process cleanup in a finally block

Four new tools (`ast_grep`, `actionlint_scan`, `osv_scan`, `gh_evidence`) are added, each following the same pattern: lazy binary resolution, input validation (path traversal, control chars), bounded execution via the runner, and structured error responses for missing binaries.

The existing `search` tool is refactored to use the runner instead of direct `bunSpawn` calls. The fallback search is rewritten with bounded traversal, directory skip lists (`.git`, `.swarm`, `node_modules`, etc.), and a warning field in the result.

Tool metadata (`tool-metadata.ts`), manifest (`manifest.ts`), and exports (`index.ts`) are updated to register the new tools. Documentation (`docs/architecture.md`, `docs/releases/pending/search-ripgrep-runner-hardening.md`) is updated with tool descriptions.

Tests are added for all four new tools and for the hardened search runner, using `_internals` seams to mock subprocess calls and binary resolution.

---

|---|---|---|
| O-001 | `SUPPORTED` | `src/tools/search.ts:263-271` — runner used with bounded options; `src/utils/external-tool-runner.ts:95-187` — cwd validation, stdin ignore, timeout, kill |
| O-002 | `SUPPORTED` | `src/tools/search.ts:499-608` — bounded traversal with `FALLBACK_MAX_*` constants; `src/tools/search.ts:405-407` — `DEFAULT_SKIP_DIRS`; `src/tools/search.ts:601-603` — warning field |
| O-003 | `SUPPORTED` | `src/tools/ast-grep.ts`, `actionlint-scan.ts`, `osv-scan.ts`, `gh-evidence.ts` — all use `_internals.runExternalTool` with bounded options |
| O-004 | `SUPPORTED` | All four tools resolve binaries inside `execute()` (e.g., `ast-grep.ts:183`, `actionlint-scan.ts:296`) |
| O-005 | `SUPPORTED` | `src/tools/tool-metadata.ts:401-437`, `src/tools/manifest.ts:198-201`, `src/tools/index.ts:3-4, 41, 51` |
| O-006 | `PARTIALLY_SUPPORTED` | Tool tests exist (`ast-grep.test.ts`, `external-cli-tools.test.ts`) but `external-tool-runner.test.ts` does not exercise multi-chunk truncation path |
| O-007 | `SUPPORTED` | No init-time binary probes; all tools resolve lazily in execute |

---

## 🚨 Confirmed Findings

### [HIGH] Windows SIGKILL silently fails to terminate hung processes

- **Location**: `src/utils/external-tool-runner.ts:107-110`
- **Why it matters**: On Windows, `SIGKILL` is not a valid signal — only `SIGTERM` and `SIGINT` are supported. The kill call is wrapped in try-catch so it won't crash, but it will silently fail to terminate the subprocess, leaving it running indefinitely on Windows. This defeats the purpose of the timeout mechanism for long-running tools like `osv_scan` (60s timeout).
- **Evidence**: Lines 107-110:
```typescript
try {
    proc?.kill('SIGKILL');
} catch {
    // best effort
}
```
The finally block at line 183 also calls `proc.kill()` without arguments (platform-neutral), but that only runs after the process exits or the timeout promise resolves — if the SIGKILL fails and the process is still running, there's no second attempt before the test/framework cleanup.
- **Fix direction**: Use platform-aware signal selection:
```typescript
const signal = process.platform === 'win32' ? 'SIGTERM' : 'SIGKILL';
proc?.kill(signal);
```

---

### [HIGH] Stream reader lock-release race on truncation

- **Location**: `src/utils/external-tool-runner.ts:152, 160`
- **Why it matters**: When output exceeds `maxBytes`, the code calls `await reader.cancel()` (line 152) to stop reading. The `finally` block then calls `reader.releaseLock()` (line 160). If `cancel()` internally releases the lock, `releaseLock()` will throw "reader was already released". While wrapped in try-catch, this means:
  1. The error is silently swallowed
  2. The stream reader state is left inconsistent
  3. On Node.js ReadableStream implementations (not Bun), `cancel()` may not release the lock — creating a deadlock where the reader never closes
- **Evidence**: Lines 148-163:
```typescript
if (value.byteLength > remaining) {
    if (remaining > 0) {
        chunks.push(value.slice(0, remaining));
        total += remaining;
    }
    truncated = true;
    try {
        await reader.cancel();  // Line 152 - may release lock
    } catch { /* best effort */ }
    break;
}
// ...
} finally {
    try {
        reader.releaseLock();  // Line 160 - throws if lock already released
    } catch { /* best effort */ }
}
```
- **Fix direction**: Check if the stream is already done/closed before calling `releaseLock()`, or use a single cancellation path that handles both cases.

---

### [MEDIUM] Test does not exercise multi-chunk truncation path

- **Location**: `tests/unit/utils/external-tool-runner.test.ts:28-41`
- **Why it matters**: The `streamFromText` helper enqueues the entire text in a single chunk. The bounded stream truncation logic at lines 146-153 of `external-tool-runner.ts` (which handles `value.byteLength > remaining`) is never exercised. This means the critical path for output limiting — the most security-sensitive part of this utility — has no test coverage.
- **Evidence**: Lines 28-41:
```typescript
function streamFromText(text: string): BunCompatSubprocess['stdout'] {
    const bytes = new TextEncoder().encode(text);
    return {
        async text() { return text; },
        async bytes() { return bytes; },
        getReader() {
            let done = false;
            return new ReadableStream<Uint8Array>({
                pull(controller) {
                    if (done) { controller.close(); return; }
                    done = true;
                    controller.enqueue(bytes);  // Single chunk!
                },
            }).getReader();
        },
    };
}
```
- **Fix direction**: Create a helper that enqueues multiple chunks, e.g., split the text into 4KB segments and enqueue them sequentially.

---

### [MEDIUM] Timeout test relies on external framework cleanup

- **Location**: `tests/unit/utils/external-tool-runner.test.ts:107-123`
- **Why it matters**: The timeout test uses `exited: new Promise<number>(() => {})` which never resolves. The test relies on Bun's test framework timeout to kill the process, but this depends on external cleanup that may not be guaranteed. If the test runner does not enforce a hard timeout for subprocesses, this test could leave a hanging subprocess after completion.
- **Evidence**: Lines 107-123:
```typescript
test('returns timeout and kills a never-exiting process', async () => {
    let killCount = 0;
    _internals.bunSpawn = (() => ({
        stdout: streamFromText(''),
        stderr: streamFromText(''),
        exited: new Promise<number>(() => {}),  // Never resolves!
        exitCode: null,
        kill: () => { killCount++; },
    })) as typeof realBunSpawn;
    // ...
});
```
- **Fix direction**: Add explicit cleanup in the test's `afterEach` or use a shorter framework-level timeout. Alternatively, resolve the promise after a small delay to simulate eventual process exit.

---

- **Risk**: `osv_scan`'s 60-second timeout with 12MB stdout cap may be insufficient for large dependency trees on slow CI runners.
  - *Why suspicious*: OSV-Scanner can take significant time on repos with many dependencies; the 12MB cap could truncate useful vulnerability data.
  - *What would verify it*: Run `osv_scan` on a large monorepo (e.g., Chromium, Node.js) and observe truncation behavior.

- **Risk**: `ast_grep`'s pattern length validation (`MAX_PATTERN_LENGTH = 20_000`) may reject legitimate complex patterns from AI-generated queries.
  - *Why suspicious*: AI agents may generate regex-like patterns that exceed this limit; no error message indicates how to reduce complexity.
  - *What would verify it*: Test with a 25,000-character pattern and observe the error response.

---

## 🧪 Test / Coverage Gaps

- **Gap**: No test for `readBoundedStream` multi-chunk truncation path (see [MEDIUM] finding above).
- **Gap**: No test for Windows signal handling in `runExternalTool`.
- **Gap**: No test for `ast_grep` with `max_results: 0` (edge case where tool should return empty matches but still execute).
- **Gap**: No test for `gh_evidence` with a very large PR body (>20,000 chars) to verify truncation behavior.

---

## 📋 Shipped-vs-Claimed Gaps

- **Gap**: PR description claims "PASS `bun --smol test src\tools\search.test.ts ...` (153 pass)" but the new `external-tool-runner.test.ts` file adds 4 tests that are not counted in this total. The actual count may differ from the claim.
- **Gap**: Documentation update in `docs/architecture.md` claims "Current output includes `matches[].file`, `matches[].lineNumber`, `matches[].lineText`" but the `search.ts` type definition (lines 36-38) only adds `engine`, `outputTruncated`, and `warning` fields — `lineText` is not explicitly added to the type. The field exists in runtime output but is not reflected in the TypeScript interface.

---

## 📝 Merge Recommendation

**[APPROVE_WITH_FIXES]**

Two HIGH-severity findings (Windows SIGKILL handling, stream reader race condition) require fixes before merge. Four MEDIUM-severity findings (test coverage gaps, framework cleanup reliance) should be addressed but can be tracked as follow-ups if time is constrained.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| Test coverage adequate | ⚠️ (see MEDIUM findings) |
| No hardcoded secrets | ✅ |
| Lockfile consistent | ✅ |

---

### 🔁 Validation provenance

**Survived from reviewer:**
- Windows SIGKILL issue — CONFIRMED (wrapped in try-catch, not CRITICAL but HIGH due to silent resource leak)
- Stream reader race condition — CONFIRMED (cancel vs releaseLock conflict on truncation path)
- Single-chunk stream test — CONFIRMED (truncation logic never exercised)
- Timeout test framework reliance — CONFIRMED (unresolved promise depends on external cleanup)

**Dropped from reviewer:**
- "Workflow discovery continues after file cap" — DROPPED: Inner loop re-checks `files.length >= MAX_WORKFLOW_FILES` before processing each entry
- "Files counted twice" — DROPPED: `uniqueFiles` deduplication is applied; `total` comes from JSON output, not manual tally
- "dirCount not decremented" — DROPPED: `dirCount` is a cap counter, not a depth tracker; early return is correct behavior
- "Empty discovery doesn't set truncated" — DROPPED: Empty-files block sets `truncated: discoveryTruncated`
- "normalizeRelativeWorkflowFile returns null and caller passes to realpathSync" — DROPPED: Guard `if (rel) files.push(rel)` prevents null usage
- "ast_grep test does not cover max_results: 0" — DROPPED: Code at `ast-grep.ts:186` enforces `Math.max(0, ...)` floor; zero is handled by design
- "ast_grep test does not cover path-escape for outside workspace" — DROPPED: Test exists at `ast-grep.test.ts:189`
- "parseAstGrepStream does not guard against malformed JSON" — DROPPED: try-catch is intentional by-design behavior
- "gh_evidence returns raw gh output; large bodies can exceed HTTP response limits" — DROPPED: `sanitizeParsedJson` truncates strings >20,000 chars
- "runExternalTool cleanup is best-effort" — DROPPED: This is intentional design, not a defect

**Blind-spot additions:**
- Stream reader race condition on truncation (HIGH) — Not raised by reviewer; identified via analysis of cancel() vs releaseLock() semantics across Bun/Node implementations
- Shipped-vs-claimed gap in documentation (lineText field) — Identified via diff comparison between docs and type definitions

---

zaxbysauce:
# PR Review: #1542 — feat(tools): add bounded external cli wrappers

## PR Intent

PR #1542 introduces a shared `external-tool-runner` for safe subprocess execution and four read-only CLI wrappers (`ast_grep`, `actionlint_scan`, `osv_scan`, `gh_evidence`) plus targeted hardening of the existing `search` tool. All four wrappers use lazy binary resolution with structured missing-binary guidance and bounded execution.

## Implementation Summary

- **`src/utils/external-tool-runner.ts`** (NEW, 232 lines) — central subprocess wrapper: array-form `bunSpawn`, explicit `cwd`, `stdin:'ignore'`, timeout, bounded stdout/stderr, best-effort `proc.kill()` in `finally`
- **`src/tools/search.ts`** (MODIFIED) — ripgrep runner integration with bounded contract; safe fallback traversal
- **`src/tools/ast-grep.ts`** (NEW), **`actionlint-scan.ts`** (NEW), **`osv-scan.ts`** (NEW), **`gh-evidence.ts`** (NEW) — read-only wrappers using `_internals` DI seam
- **`src/tools/index.ts`, `manifest.ts`, `tool-metadata.ts`** — registration
- **`tests/unit/utils/external-tool-runner.test.ts`** (NEW, 145 lines), **`src/tools/ast-grep.test.ts`** (NEW, 203 lines), **`external-cli-tools.test.ts`** (NEW, 319 lines), **`search.test.ts`** (modified, +100)
- **`docs/architecture.md`** (modified), **`docs/releases/pending/search-ripgrep-runner-hardening.md`** (NEW)

## Intended vs Actual Mapping

| Obligation | Source | Status | Finding |
|---|---|---|---|
| O-001: Harden `search` with bounded runner + safer fallback | PR body | MET | — |
| O-002: Read-only `ast_grep`, `actionlint_scan`, `osv_scan`, `gh_evidence` wrappers with lazy resolution + bounded execution | PR body | MET | — |
| O-003: Wire tool metadata/manifest/exports/docs/release notes | PR body | MET | — |
| O-004: Focused regression coverage for subprocess contracts/truncation/malformed output/registration coherence | PR body | PARTIALLY_MET | C5-MISSING-NEG-001/002/003/004 |

## Validation Provenance

### Explorer lanes
| Lane | Status | Candidates |
|---|---|---|
| Lane 1: Correctness | Completed | Output truncated, no formatted candidates |
| Lane 2: Security | Completed | Output truncated, no formatted candidates |
| Lane 3: Dependencies | Completed | 2 candidates extracted |
| Lane 4: Docs/Drift | Completed | Output truncated, no formatted candidates |
| Lane 5: Tests | Completed | 11 candidates extracted |
| Lane 6: Performance | Completed | Output truncated (header only) |
| Micro: Shell/Subprocess | Completed | 2 candidates extracted |
| Micro: URL/External Fetch | Completed | 3 candidates extracted |
| Micro: Tool Registration | Completed | Clean — all 4 tools correctly registered |

### Deterministic signals (Phase 2)
- CI: `quality`, `drift`, `package-check`, `php-validation`, `security`, `pr-standards`, `check-title`, `check-duplicates`, `detect-release`, `rust-sandbox-runner` — all PASS
- 4 unit jobs pending (no failures observed)
- github-actions drift check comment: PASS — "No drift detected across skills, tools, commands, and agents"

### Reviewer dispatched
**`lowtier_reviewer`** — 17 candidates classified

### Critic dispatched
**SKIPPED** — No reviewer-confirmed HIGH/CRITICAL findings; all confirmed findings are LOW/INFO and do not warrant critic escalation.

## Confirmed Findings

### F-001 — LOW — Test coverage: outsideDir leak on realpathSync failure
**File:** `src/tools/ast-grep.test.ts:157-160`
**Evidence:** CONFIRMED (reviewer)
**Rationale:** `outsideDir` is created before the `try` block. If `realpathSync` throws, the directory created by `mkdtempSync` is not captured and cannot be cleaned by the `finally` block.
**Fix:** Move `mkdtempSync`/`realpathSync` inside the `try` block so `finally` can clean on assignment failure.

### F-002 — LOW — Runtime race: Promise.race ambiguity in subprocess timeout
**File:** `src/utils/external-tool-runner.ts:185`
**Evidence:** CONFIRMED (reviewer)
**Rationale:** `Promise.race` between `proc.exited` and manual `setTimeout` can resolve either way if both fire in the same tick, potentially reporting a just-completed run as `timeout`. `kill()` on an already-dead process is a no-op.
**Fix:** Capture `proc.exited` before the race, or await `proc.exited` after timeout fires to disambiguate.
**Falsification:** Requires timing-sensitive reproduction; otherwise code inspection confirms the race.

## Pre-existing Findings

| ID | File | Note |
|---|---|---|
| PE-001 | `src/tools/search.test.ts:617-641` | Regex timeout test only validates response type — pre-existing on base `dc539021` |
| PE-002 | `src/tools/search.test.ts:560-572` | rg-not-found fallback test only validates `JSON.parse` — pre-existing on base |

## Advisory / INFO Findings

| ID | File | Note |
|---|---|---|
| F-INFO-001 | `src/tools/ast-grep.test.ts` | No test for non-zero exit code from ast-grep subprocess |
| F-INFO-002 | `src/tools/ast-grep.test.ts` | No tool-level SIGTERM/SIGKILL test (runner timeout covered separately) |
| F-INFO-003 | `src/tools/external-cli-tools.test.ts:159-220` | `osv_scan` clean (exitCode:0) path untested |
| F-INFO-004 | `src/tools/external-cli-tools.test.ts:257-299` | `gh_evidence` truncation boundary (exact 20k) untested |
| F-INFO-005 | `src/utils/external-tool-runner.ts:46` | PATH delimiter split assumes single char; literal `;` in PATH entries non-standard |
| F-INFO-006 | `src/utils/external-tool-runner.ts:51` | `statSync` follows symlinks — standard PATH resolution behavior |
| F-INFO-007 | `src/utils/external-tool-runner.ts:62` | `path.join` without traversal check on rawName — callers pass literal binary names |
| F-INFO-008 | `src/tools/gh-evidence.ts` | No dedicated test file; covered via `external-cli-tools.test.ts` |
| F-INFO-009 | `src/tools/gh-evidence.ts:207` | gh binary resolved lazily from PATH at execute time |

## Disproved Candidates

| Candidate | Reason |
|---|---|
| C5-MOCK-LEAK-001/002/003 | `_internals` restoration removes runtime reference; `mock()` instance is local; no cross-test leakage surface |
| C5-TEST-COUNT | Reviewer confirmed the actual 153-test count: search.test.ts(37) + search.adversarial.test.ts(99) + ast-grep.test.ts(4) + external-cli-tools.test.ts(9) + external-tool-runner.test.ts(4) = 153 ✓ |

## Verdict
**APPROVE_WITH_NOTES** — Implementation matches PR claims; CI green on all required checks. Two LOW findings (F-001, F-002) are minor improvements; INFO-level findings are non-blocking observations. No CRITICAL/HIGH issues.

## Merge Recommendation

This PR successfully introduces a robust subprocess runner with proper bounds, timeouts, cleanup, and structured error handling. The four read-only wrappers correctly use the shared runner with DI seam isolation. Tool registration is complete across all required layers. CI is green on all required checks.

**Required actions before merge:** None.

**Optional follow-ups (non-blocking):**
- F-001 (LOW): Move `mkdtempSync` inside `try` block for guaranteed cleanup
- F-002 (LOW): Capture `proc.exited` before timeout race to avoid ambiguous timeout status
- F-INFO-001 to F-INFO-009: Test coverage gaps can be addressed in follow-up PRs

## Reviewer Identity
- `lowtier_reviewer` (single chunked review covering all 17 candidates)

## Rerun Note
Truncated lane outputs (1, 2, 4, 6) did not produce formatted candidates. Findings from lanes that returned structured output (3, 5) and all three micro-lanes were sufficient for verdict determination.

zaxbysauce:
## PR #1542 Feedback Closure

Head: `3cf77537867e040b1887f2e04ee921989e17068e`

Intake sources checked:
- PR metadata, comments, reviews, check rollup, and inline review comments via `gh`.
- Review threads via GraphQL: `reviewThreads.nodes` returned `[]`.
- Issue comments from both review passes authored by `zaxbysauce`.
- Independent verifier agent `Hooke` (`gpt-5.4-mini`) checked the timeout/race/truncation/test-cleanup findings.

Closure ledger:
- FB-001 - fixed - Windows timeout cleanup no longer sends `SIGKILL`; `runExternalTool` now maps Windows timeout cleanup to `SIGTERM`. Evidence: `uses SIGTERM for timeout cleanup on Windows` test PASS.
- FB-002 - fixed - bounded stream truncation now uses one cancel-and-release path and avoids duplicate release errors. Evidence: runner focused test file PASS.
- FB-003 - fixed - multi-chunk truncation is now covered. Evidence: `truncates stdout across multiple chunks` test PASS and focused suite now reports 155 pass.
- FB-004 - fixed - timeout test no longer uses an impossible never-settling `exited` promise; it uses a deferred resolver and completes after asserting timeout behavior.
- FB-005 - needs more evidence, nonblocking - OSV timeout/stdout cap concern did not identify a concrete current-head defect. Existing bounded timeout and byte caps remain in place.
- FB-006 - needs more evidence, nonblocking - ast-grep maximum pattern-length concern did not identify a concrete current-head failing input. Existing bounded runner behavior remains in place.
- FB-007 - disproved/updated - original `153 pass` claim matched the pre-feedback focused suite; the updated PR body records `155 pass` for the post-feedback focused command.
- FB-008 - disproved - `SearchMatch` already includes `lineText`, and both ripgrep JSON and fallback search paths populate it.
- FB-009 - fixed - `ast_grep` malformed/outside-workspace test now only cleans `outsideDir` if creation succeeded, preventing cleanup-path masking.
- FB-010 - fixed - timeout race now tracks settled exit state before resolving timeout, avoiding an ambiguous `Promise.race` result after process exit.
- PE/INFO items - acknowledged - advisory items without concrete current-head defects were not changed in this pass.

Validation:
- PASS `bun run typecheck`
- PASS `bun --smol test tests\unit\utils\external-tool-runner.test.ts src\tools\ast-grep.test.ts src\tools\external-cli-tools.test.ts src\tools\search.test.ts src\tools\search.adversarial.test.ts --timeout 60000` (155 pass)
- PASS `bunx @biomejs/biome@2.3.14 ci .`
- PASS `bun run build`
- PASS `git diff --check`
- PASS `node --input-type=module -e "await import('./dist/index.js'); console.log('dist import OK')"`
- PASS `node scripts\repro-704.mjs` after sandbox escalation (T1 225.2ms/400ms, T2 85.6ms, T3 94.7ms; harness printed PASS and then needed manual cleanup of the lingering validation process)
- PASS GraphQL review-thread query returned no review threads to resolve.

zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `3cf77537867e`

---

- **[CRITICAL]** `src/tools/actionlint-scan.ts:139` — normalizeRelativeWorkflowFile uses fs.realpathSync but does not re-verify the resolved file is inside the workspace after (calls fs.realpathSync(workspace) then fs.realpathSync(resolved) and checks path.relative(realWorkspace, realFile) without an explicit containment boundary check — a malicious symlink at a path that looks safe can escape the directory)
- **[HIGH]** `src/tools/actionlint-scan.ts:186` — parseActionlintOutput catches all JSON.parse errors silently and ignores malformed lines without logging (try { ... } catch { /* Ignore malformed lines from mixed stdout */ } — any line that isn't valid JSON is dropped silently, so an error in the tool's output is never surfaced)
- **[MEDIUM]** `src/tools/actionlint-scan.ts:226` — comparison between requestedFiles and discovery.files for path-escape is flawed because discovery.files is de-duplicated and requestedFiles is not (requestedFiles comes from the user as a raw array; discovery.files is constructed via normalizeRelativeWorkflowFile which returns null for bad paths and is later passed through new Set(files); so a user providing a duplicate that resolves identically will pass the count check even though it's a malformed request)
- **[HIGH]** `src/tools/ast-grep.test.ts:186` — The test writes an 'outside.ts' to a tmp dir but doesn't verify the containment check actually blocked it; a false positive on the escape test would hide a real bug (The test writes a file in a separate tmp dir (outside the ast_grep cwd) and expects only the inner file to appear in results; if the containment logic were broken, the outside file would appear, but the test only asserts the inner file is present, not that the outer one is absent)
- **[MEDIUM]** `src/tools/ast-grep.ts:43` — No validation that the pattern is valid ast-grep syntax; invalid patterns fail at runtime with a 700 error the tool maps to 'invalid-pattern' (The code checks length and control characters but never validates pattern structure; an invalid ast-grep expression will still pass pre-flight and only be caught by checking run.exitCode != 0 after execution, which is acceptable but not compile-time or parse-time validation)
- **[HIGH]** `src/tools/gh-evidence.ts:195` — The gh_evidence tool does not truncate the output body even after the 20,000-character sanitization threshold; the body is truncated post-parse rather than during the gh call, so the external tool still emits the full body. (Ghci_evidence parses run.stdout with JSON.parse(run.stdout) at line 272 before sanitizing; there is no --max-fields or --limit on the gh command that would bound the raw output before it reaches the process.)
- **[MEDIUM]** `src/tools/tool-metadata.ts:403` — tool_metadata is missing a 'description' field for 'batch_symbols' — inconsistent with all other entries
- **[HIGH]** `src/utils/external-tool-runner.ts:79` — runExternalTool returns 'spawn-error' without capturing the actual exitCode when the spawn itself throws — callers can't distinguish a failed spawn from a subprocess that failed to start
- **[MEDIUM]** `src/utils/external-tool-runner.ts:116` — readBoundedStream releases its lock in a finally block but the lock-release logic is nested inside try/catch — if the finally runs after a cancel, the lock might not be released cleanly
- **[MEDIUM]** `src/utils/external-tool-runner.ts:127` — timeoutHandle is cleared in finally but if the timeout fires after the finally runs, the clearTimeout could be called twice on the same handle
- **[LOW]** `src/utils/external-tool-runner.ts:141` — killProcess swallows all errors when calling proc.kill — if kill fails for a reason that's meaningful (e.g., process already exited), the caller can't know
- **[MEDIUM]** `src/utils/external-tool-runner.ts:166` — readBoundedStream returns { text: '', truncated: true } for maxBytes <= 0 but the caller never checks truncated — stdoutTruncated stays false even though no text was read
- **[MEDIUM]** `src/utils/external-tool-runner.ts:177` — timeoutKillSignal has no fallback for unknown platforms — on an exotic platform it returns undefined and proc.kill(undefined) will throw
- **[HIGH]** `src/utils/external-tool-runner.ts:184` — runExternalTool has a bare 'catch (err)' that returns a result but never logs or reports the error — the caller loses the error message
- **[LOW]** `src/utils/external-tool-runner.ts:255` — Hardcoded Bun globals in runExternalTool make it harder to test and couple it to Bun (The module embeds _internals.bunSpawn and _internals.platform instead of taking them as injected dependencies, making it difficult to mock the platform or the spawn call in unit tests.)
- **[MEDIUM]** `tests/unit/utils/external-tool-runner.test.ts:247` — Truncation test uses fixed chunk sizes, not the actual truncation logic (The test feeds 'abc' and 'defgh' (2+5=7) against maxBytes 6. The result is 'abcdef' with truncated: true, which confirms the slice(0, remaining) logic is what's exercised, not the boundary condition itself.)
- **[MEDIUM]** `tests/unit/utils/external-tool-runner.test.ts:257` — test 'resolves platform executable names from PATH lazily' doesn't verify that non-existent executables are skipped (test creates an empty file and asserts a match; there is no negative case for an executable that does not exist on the path.)

> _Critic (Qwen3.6-35B-APEX) failed (URLError); findings passed reviewer validation but were NOT challenged or blind-spot-checked._

---

zaxbysauce:
## Phase 11 — Post-Fix Re-verification ✅

Author pushed fix commit `3cf77537` addressing the two LOW findings. Re-verification complete.

### Verdict: APPROVED — PR ready for merge

**[REVERIFIED] F-002 — FIXED**
- File: `src/utils/external-tool-runner.ts:188-217`
- Discriminated-union `ExitRaceResult` (`{ kind: 'exit', exitCode } | { kind: 'timeout' }`) eliminates the original `Promise.race` ambiguity
- `exitSettled` flag set synchronously in `proc.exited.then`; timeout handler resolves as `kind:'exit'` when `exitSettled` is true
- Portable kill signal: `SIGTERM` on win32, `SIGKILL` elsewhere — fixes Node.js compat for Windows
- New `killProcess` helper centralizes best-effort cleanup
- **Confirmed by reviewer**: A just-completed process is never reported as timeout

**[REVERIFIED] F-001 — PARTIALLY_FIXED**
- File: `src/tools/ast-grep.test.ts:157-202`
- `outsideDir` declared as `let | undefined`; assignment moved inside `try` block
- `finally` uses `if (outsideDir) rmSync(...)` to prevent the original ReferenceError on uninitialized access
- **Residual gap (informational)**: if `realpathSync` throws after `mkdtempSync` succeeds, the raw temp dir leaks because `outsideDir` stays undefined. Root-cause fix would capture the raw `mkdtempSync` path and clean it up on either success or failure. Not blocking; can be addressed in a follow-up.

### Additional test coverage (author's claim: "covering truncation across multiple stdout chunks")
- `streamFromChunks` helper exercises multi-chunk truncation; new test asserts capped output (`'abcdef'`) and `stdoutTruncated: true`
- `uses SIGTERM for timeout cleanup on Windows` verifies platform-aware signal choice
- Timeout tests call `resolveExit(143)` to flush pending `exited` promise and avoid hanging async resource

### Reviewer answers to verification questions
1. F-002 fallback chain `settledExitCode ?? proc?.exitCode ?? 0` — defensive/dead code under current invariant; not a merge blocker but worth watching
2. F-001 partial fix is sufficient for the reported scenario; residual temp-dir leak is unlikely in practice
3. New tests cover the fixed scenarios with meaningful assertions (not weak tautologies)
4. Platform-aware SIGTERM works correctly on Windows for `subprocess.kill()` emulation

### Merge status
- All originally identified HIGH/CRITICAL findings: 0
- LOW findings: 2 (both addressed; F-001 has residual informational gap)
- INFO findings: 9 (non-blocking, optional follow-up)
- CI: `quality`, `drift`, `package-check`, `security`, `pr-standards`, `check-title`, `check-duplicates`, `detect-release`, `rust-sandbox-runner`, `php-validation` all PASS
- Unit jobs: pending (no observed failures)

**PR #1542 is ready for merge.** No required actions remaining. Optional follow-up: capture raw `mkdtempSync` path in F-001 to fully close the residual temp-dir leak edge.

PR #1831, opened 2026-07-13
fix(audit): resolve medium and low severity defects from the 2026-07-09 audit

### LINKED ISSUE #1779: [Audit 2026-07-09] MEDIUM-HIGH / MEDIUM / LOW defects + test-debt — reviewer validated
## Summary

Consolidated tracking issue for the **MEDIUM-HIGH, MEDIUM, and LOW/test-debt** defects from the full-repo audit of `main` @ `8822ffc9` (v7.110.1), 2026-07-09. Companion to the CRITICAL/HIGH tracker. Every item passed independent reviewer validation; CRITICAL/HIGH-adjacent items also passed a critic pass. `file:line` + evidence inline. These are correctness/reliability/hygiene defects — none is a critical security hole, but several are cheap fixes one refactor away from becoming serious.

> Full local evidence: `.swarm/review-v8/runs/2026-07-09-full-audit/`. "Known #N" = matches an open GitHub issue, confirmed at HEAD.

---

### MH1 — `pkg_audit` `auto` mode reports `clean:true` when a scan silently fails
**File:** `src/tools/pkg-audit.ts` (`runAutoAudit`, `CombinedAuditResult`); allowlist ref `src/full-auto/policy.ts:109`.
Default `ecosystem:'auto'` aggregation drops per-ecosystem failure notes (tool-missing/timeout/parse-error) — `CombinedAuditResult` has no `note` field and the loop never reads `.note` — and returns `clean:true`. Granted to `reviewer`/`critic_oversight`. **Severity is MEDIUM-HIGH, not HIGH, because it is wired into no machine gate** (absent from `pre_check_batch`; only an allowlist entry; no agent prompt instructs gating on its `clean` result) — so the harm ceiling is a misled human/LLM approver, not automated auto-approval of a vulnerable dependency. **Re-check against real maintainer workflow: if a clean `pkg_audit` is ever treated as a supply-chain sign-off, upgrade to HIGH.** Fix: surface per-ecosystem failures; never return `clean:true` when a sub-scan did not complete; add non-happy-path `auto` tests.

---

### M1 — Plan-ledger silent rollback (divergent readers + non-truncating quarantine)
**File:** `src/plan/ledger.ts` (`readLedgerEvents:328-366` skip-reader; `readLedgerEventsWithIntegrity`/replay `:986-1030`; quarantine `:1039-1057`; `appendLedgerEvent:508-523`), `src/plan/manager.ts` (`loadPlan`, `rebuildPlan`). Related **#1692**.
`getLatestLedgerHash` uses the skip-malformed reader (sees events after a bad line); `replayFromLedger` uses the stop-at-first-bad reader (halts, quarantines the remainder **without truncating** canonical) and discards the `truncated` flag. On `planHash != ledgerHash`, `rebuildPlan` overwrites `plan.json` from the prefix-only replay — silently dropping a durable `task_status_changed` after the poison line. **Confirmed MEDIUM (not HIGH):** `appendLedgerEvent` is an atomic whole-file temp+`renameSync` under lock, so a natural mid-ledger poison line cannot arise from the subsystem itself (needs external corruption/bit-rot/manual edit). **Caveats keeping it live:** no `fsync` before rename; rename atomicity is a local-FS guarantee (weak on network/overlay FS); a future switch to `appendFileSync` would weaponize a torn append into HIGH. Fix (cheap): gate the destructive `rebuildPlan` on `truncated === false` (thread the flag out), or make replay skip-and-continue, or atomically rewrite canonical to clean-prefix+valid-suffix after quarantine; add `fsync`-before-rename; test that both readers agree on ledger end.

### M2 — SQLite `withTransaction` provides no atomicity (async callback to a sync transaction)
**File:** `src/memory/sqlite-provider.ts:536-541`; caller `src/memory/reward-capture.ts:172-200`.
`withTransaction` does `db.transaction(() => fn({}))()`. bun:sqlite's `db.transaction` wraps a **synchronous** body in BEGIN/COMMIT; the callback is `async` with awaited get/upsert/appendRewardEvent, so COMMIT runs after the synchronous portion and **every awaited write executes after COMMIT, outside any transaction** (reviewer confirmed with a standalone bun:sqlite repro — `db.inTransaction` goes false before the INSERT runs). bun:sqlite does not throw (unlike better-sqlite3), so it is silent. FB-004 race protection for concurrent Q-learning reward updates is a no-op. **MEDIUM:** bounded/self-healing for the current soft-EMA caller (same-id concurrency is uncommon in a single-threaded process; lost update is one EMA step on a soft Q-value). **But it is a HIGH-grade latent-reuse trap** — it *looks* protected, so any future hard-consistency caller silently loses atomicity. Fix: do read-compute-write **synchronously** inside a real transaction (prepared statements, no awaits) or explicit `BEGIN IMMEDIATE`/`COMMIT`; make `withTransaction` **reject** a thenable-returning callback so the trap can't silently reappear; add a rollback-on-throw test against the real provider.

### M3 — `allow_full_suite` is a dead documented arg with a failing test (downgraded from HIGH)
**Files:** `src/tools/test-runner.ts` (arg handling), `src/tools/__tests__/test-runner-history.test.ts` (fails 19/1 at HEAD), `tests/unit/tools/test-runner-scope-cap.test.ts` (test 12 vacuous), `AGENTS.md` invariant 6 wording.
The `test_runner` `allow_full_suite` argument is **ignored** — the real guard is the `SWARM_ALLOW_FULL_SUITE` env var, and mis-following the docs fails **closed** (invariant-6 protection is fully intact). But the arg is documented in AGENTS.md/schema (unwired/misdocumented per directive 2), its regression test **fails 19/1 at HEAD** (reproduce: `bun test src/tools/__tests__/test-runner-history.test.ts`), and a sibling scope-cap test (test 12) is vacuous (test 13 passing 16/0 proves the env var is unset, which is what makes test 12 assert nothing). MEDIUM must-fix (not a security issue). Fix: either wire the arg or delete it from docs+schema; fix the failing test's expected string; harden the vacuous assertion. (Note: this failing test is invisible today because of H4 in the sibling issue — CI doesn't run it.)

### M4 — `gitingest` sends model-supplied URLs to a third party ungated (downgraded from HIGH)
**File:** `src/tools/gitingest.ts`; policy `src/full-auto/policy.ts` (READ_ONLY / auto-approve).
Model-supplied `url`/`pattern` sent verbatim to hardcoded `https://gitingest.com/api/ingest` with no domain allowlist and **no** `council.general.enabled` gate (unlike `web_fetch`), classified READ_ONLY (auto-approved in full-auto), exposed to the `architect` (the agent that ingests untrusted PR text). **MEDIUM:** the hardcoded endpoint kills SSRF-to-internal and exfil-to-attacker-controlled-host; residual is a low-bandwidth privacy leak (path/pattern into a benign third party's logs the attacker can't read). The valid concern is the gating asymmetry vs `web_fetch`. Fix: gate it like `web_fetch` (config flag) and/or validate the target.

### M5 — CLI `uninstall --clean` deletes `PROMPTS_DIR` without the four-layer guard
**File:** `src/cli/index.ts:508-509` (`fs.rmSync(PROMPTS_DIR,{recursive:true})`), contrast guarded eviction `:382-390`.
The uninstall path recursively deletes `PROMPTS_DIR` guarded only by `existsSync`, bypassing `isSafeCachePath` that the sibling `evictPluginCaches` applies to structurally identical deletes. `PROMPTS_DIR = path.join(CONFIG_DIR,'opencode-swarm')` where `CONFIG_DIR` derives from `XDG_CONFIG_HOME`. **MEDIUM:** blast radius bounded by the fixed `opencode-swarm` suffix; requires attacker-controlled `XDG_CONFIG_HOME` + a deliberate `--clean` + a valuable tree at that exact path. Fix: apply `isSafeCachePath` (or an equivalent 4-layer check) before this delete.

### M6 — Cache-path guard is lexical-only (no realpath) — symlink follow on delete
**File:** `src/cli/index.ts` (`isSafeCachePath`, `isSafeLockFilePath`).
The guards validate only the `path.resolve` (lexical) form — no `realpathSync` anywhere in `src/cli/index.ts`, unlike ~20 other files that canonicalize before trusting a path. A same-user attacker who plants a symlinked intermediate directory (e.g. a symlinked `node_modules`) passes the string check but is transparently followed by the actual `rmSync`. MEDIUM (bounded by same-user trust). Fix: `realpathSync` the target and re-validate the canonical path immediately before the delete (no re-resolution gap).

### M7 — Config/docs key drift → silent no-op (`skill_propagation` vs `skillPropagation`)
**Files:** `docs/configuration.md` (documents `skill_propagation`), schema + runtime read `src/index.ts:859` (actual key `skillPropagation`).
The documented config key doesn't match the schema/runtime key; with no `.strict()`, zod silently strips the unknown `skill_propagation` block and falls back to defaults. A user following the docs verbatim gets a silent no-op; the config-doctor's Levenshtein unknown-key warning is opt-in, not default-on. Fix: correct the docs to `skillPropagation` (or accept both); consider a default-on unknown-key warning for top-level config.

### M8 — Documented behaviors that don't match code
**Files:** Context Budget Guard docs vs `src/.../context-budget.ts`; lockfile docs vs `src/.../file-locks.ts:163-172`.
(a) Docs claim a Context Budget Guard with `max_injection_tokens` and a ~2800/~3600-of-4000 window; the code never reads `max_injection_tokens` and measures total conversation tokens ÷ the real model window (128k–200k). (b) Docs claim `proper-lockfile` fail-fast `retries:0`; code uses `retries:5` with exponential backoff (with an in-code comment confirming the intentional change). Fix: update docs to match code (or reconcile intent).

### M9 — Unbounded session-keyed maps (invariant-8)
**Files:** `src/index.ts:147` (`_heartbeatTimers`), `src/hooks/adversarial-detector.ts` (`recentToolCallsBySession`).
`_heartbeatTimers` is a module-level `Map<sessionID, number>` with **no eviction** anywhere. `recentToolCallsBySession` has unbounded key cardinality (its sibling `lastSpiralTimestampBySession` is correctly FIFO-capped at `MAX_TRACKED_SESSIONS=500`; this map evicts only as a side effect of an actual spiral detection). Long-running hosts leak memory across sessions. Fix: FIFO-cap both at `MAX_TRACKED_SESSIONS`, matching the sibling. (Corroborated independently by the claims-vs-code lane.)

### M10 — Learned-content trust-laundering into the single system message
**Files:** `src/hooks/system-enhancer.ts` (injects into `output.system`; collapse at `index.ts:1913-1917`), `src/hooks/knowledge-validator.ts` (`validateLesson` scans only the `lesson` field), `src/.../context-sanitizer.ts` (`sanitizeContextText`), dark-matter writer `src/.../co-change-analyzer.ts`.
Stored/learned content (knowledge_recall lessons, retro lessons + `user_directives`, decisions, prior rejections, handoff body) is injected into `output.system` — provably collapsed into the single system message — with **no sanitizer**; and learned directive-action fields (`required_actions`/`forbidden_actions`/`triggers`) bypass `validateLesson` (scans only `lesson`) and evade `sanitizeContextText` (misses natural-language instructions and open non-system/tool tags). The dark-matter writer bypasses `validateLesson` entirely. **MEDIUM** (control gaps behind an already-poisoned-writer precondition, not a direct external-injection hole). Fix: sanitize learned content at injection time regardless of source; extend `validateLesson` to the directive-action fields; apply store-time validation to the dark-matter path. (Note: the full-auto policy ladder was verified to NOT consume injected content programmatically — good.)

### M11 — Decision-drift substring/polarity false positives (Known #1661)
**File:** `src/services/decision-drift-analyzer.ts` (contradiction detector).
Uses `.includes('use')`, which substring-matches inside `'do not use'`, flagging two **agreeing** decisions as contradictory (concrete repro: a "do not use Redis" pair). Triple-confirmed at HEAD (three lanes). Fix: tokenize + polarity-aware comparison instead of substring `.includes`.

### M12 — `renderPrompt` safety-net is dead code (Known #1649)
**Files:** `src/agents/template.ts:1-14` (unknown-placeholder-throws guard, never invoked), production substitution `src/agents/index.ts:440-465` + `src/agents/architect.ts:1670-1790` (three uncoordinated `.replace()` chains), no `{{KEY}}`-leak test.
The safety net that would catch unresolved `{{KEY}}` placeholders is dead; production prompt assembly runs through three unvalidated `.replace()` chains. Fix: route production substitution through the validated `renderPrompt`, or add a post-substitution `{{...}}`-leak assertion + a full-pipeline test.

### M13 — Skill-mirror "divergent" classification hides safety-section drift
**Files:** `src/config/skill-mirrors.ts` (classification), `engineering-conventions` skill pair (`.opencode` vs `.claude`).
The `engineering-conventions` mirror pair is classified "divergent," which the drift-check exempts — but each side has whole safety-relevant sections absent from the other (e.g. an `.opencode`-only SAST-baseline "Critical safety guard"). A contributor reading only the `.claude` copy misses it. Fix: reconcile the two copies or tighten the classifier so safety-section divergence is not silently exempted.

### M14 — SQLite handle leak on init retry
**File:** `src/memory/sqlite-provider.ts:423-438`.
`this.db` is opened before several throwable operations (migrations, legacy-JSONL import, row loads) and is never closed in the catch/retry path — a native handle leak on any `initialize()` failure + retry. No test covers the path. Fix: close `this.db` in the catch before retry (try/finally around the throwable init steps).

### M15 — Spec/FR/acceptance content not carried in a structured delegation field (Known #1687)
**Files:** `src/agents/coder.ts:12-18` (INPUT FORMAT has no acceptance/FR/SC field), `src/hooks/delegation-gate.ts:330` (`validateDelegationEnvelope` — dead code, zero callers).
Delegated coders/reviewers receive spec/FR/acceptance only as optional architect free-text — there is no structured/mechanical field carrying it and no reference/id. Reviewer downgraded to MEDIUM (facts confirmed). **Caveat:** the "nothing mechanically injects it anywhere" leg rests on a `system-enhancer.ts` grep, not a full `delegation-gate.ts` payload-assembly trace (see limitations) — verify before implementing. Fix: add an acceptance/FR field to the delegation envelope and the coder INPUT FORMAT; wire or remove the dead `validateDelegationEnvelope`.

### Also MEDIUM / known-issue confirmations at HEAD
- **#1744** — the phase-complete gate's reviewer/test_engineer enforcement exists (default `policy:'enforce'`) but is **architecturally unreachable** from plan-free PR-feedback sessions (`phase_complete`/`update_task_status` never invoked there) — same observable symptom as reported, different cause.
- **#1659** — background-delegation ingestion errors have no `/swarm status` surface.
- **#1308** — complexity-based review routing is advisory-only, not enforced.
- **#1639** — orchestration-hook hygiene: doc_scan/knowledge-store maintenance runs unconditionally on the per-message hot path.
- **#1660** — PR-monitor uses fixed-interval polling (default 60s, 4 subprocesses/PR) with no adaptive backoff, despite `computeChanges` already computing the diff signal.
- **#1691** — `isCommandAvailable` (`src/build/discovery.ts`) probes `where <name>.exe` on Windows, which fails for `.cmd`/`.bat`-distributed npm/mvn/gradle/composer (silent tool-missing). The `.cmd`-mapping precedent exists at `src/hooks/spawn-helper.ts` (`WIN32_CMD_BINARIES`) and should be reused.
- Subprocess hygiene gaps (invariant 3): `runLint`/`runAdditionalLint` (`src/tools/lint.ts`) + `pre-check-batch.ts:304` spawn with **no timeout/stdin:'ignore'/kill** (unbounded hang); `build-check.ts`/`test-runner.ts` missing `stdin:'ignore'` (v7.3.3 anti-pattern); `getGitRemoteUrl` (`src/knowledge/identity.ts`) `execSync` missing `timeout`.

---

## LOW / test-debt / hygiene

- **Stale misleading skip comments (was falsely filed as a HIGH permission bypass — DISPROVED).** `tests/.../full-auto-permission.test.ts` has `.skip` tests whose comments claim broken mode-override / enabled:false permission enforcement. **Production works correctly** (`src/hooks/full-auto-permission.ts:163-178`) — two independent validators ran the tests un-skipped (both pass) and a `git log` proof showed the cited revert commit never existed. Action: **un-skip the green tests**, delete the false comments. (Recorded here so it isn't re-reported as a bug.)
- **Test temp-dir leakage.** `tests/unit/evidence/load-evidence-adversarial.test.ts` writes `.temp-test-*` under `process.cwd()` (not `os.tmpdir()`), leaking 44 orphan dirs into the repo root (violates AGENTS.md:93). Fix: `os.tmpdir()` + `mkdtemp`; add a lint-style CI guard.
- **Unwired/dead code (directive 2):** `knowledge_ack` (built + tested, never registered), `src/output/agent-writer.ts` exports (zero callers; a CHANGELOG claim says otherwise), `src/graph/*` (Known #1641), `restricted-token-executor.ts` (imported only by a test + doc). **Do NOT delete `replayWithIntegrity`** — it is the better engine and the structural root of the M1 fix; deleting it is a regression.
- **`redactShellCommand` divergence:** the barrel re-exports the weaker `index.ts` copy (no home-path redaction); `helpers.ts` has the stronger one. Latent, no live leak.
- **Numeric/doc drift:** "12 language profiles" vs 13 registered (README's own tier table already says 13); AGENTS.md invariant 11 describes a manual registration chain that is actually compile-derived (Known #1643).
- **FR-006 unenforced:** 518 test files exceed the 500-line cap (top: `skill-propagation-gate` 4697 lines). (Fix tracked as an enhancement — a diff-scoped ratchet, Known #1666.)
- **TUI console pollution (Known #1752 epic, PRs #1753-1756):** 54 non-test files bypass the debug-gated logger with raw `console.*`, concentrated on hook/init paths.

---

## Suggested labels
`bug`, `reliability`, `tech-debt`, `config`, `testing`

## Cross-references
Companion issue tracks the CRITICAL/HIGH findings. Confirmed-at-HEAD known issues referenced above: #1692, #1661, #1649, #1687, #1744, #1659, #1308, #1639, #1660, #1691, #1641, #1643, #1666, #1752. Full audit evidence: `.swarm/review-v8/runs/2026-07-09-full-audit/` (local).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

zaxbysauce:
Closes #1779

## Summary
Resolves the reviewer-validated MEDIUM-HIGH / MEDIUM / LOW + test-debt defects from the 2026-07-09 full-repo audit. Every item was re-verified live at HEAD (105 commits past the audit base) before fixing; items already remediated since the audit (e.g. M3 `allow_full_suite`) were skipped.

- **MH1** `pkg_audit` auto mode no longer reports `clean:true` when a sub-scan silently fails (explicit `incomplete` flag; benign notes don't flip `clean`).
- **M1** plan-ledger no longer silently drops durable post-poison events: replay threads a `truncated` flag, refuses to overwrite `plan.json` with prefix-only state, never destructively rewrites the ledger, quarantines to a unique side file, and fsyncs before rename. Orphan `replayWithIntegrity` folded in and removed.
- **M2** SQLite `withTransaction` now uses manual `BEGIN IMMEDIATE`/`COMMIT`/`ROLLBACK` and awaits the async callback → real atomicity. **M14** `doInitialize` closes+nulls+resets on any post-open failure.
- **M4** `gitingest` reclassified READ_ONLY→NETWORK so full-auto escalates it to critic.
- **M5/M6** `uninstall --clean` deletes guarded (`isSafePromptsDir`/`isSafePluginConfigPath`) + realpath canonicalization before every guarded delete.
- **M9** `_heartbeatTimers` + `recentToolCallsBySession` FIFO-capped at `MAX_TRACKED_SESSIONS`.
- **M10** learned content sanitized at injection time (incl. agent-context parity fix from review F-004, wider invisible-char coverage from F-005); directive-action fields + dark-matter writer run content-safety validation.
- **M11** decision-drift contradiction detection is negation-aware (no false positive on agreeing "do not use X").
- **M12** each agent's final prompt is validated free of unresolved `{{KEY}}` (hard check for built-in prompts; warn-only / fail-open for user custom prompts so a literal `{{KEY}}` in prose can never abort init — review F-001); dead `renderPrompt` removed.
- **M13** `engineering-conventions` divergent pair carries the same safety sections in both trees; `drift:check` enforces safety-heading parity.
- **M15** optional structured `specCriteria` (FR/SC/acceptance) flows architect→coder; dead `validateDelegationEnvelope` wired advisory-only.
- **Docs (M7/M8/L5)** + **test-debt (L1–L4)**: config/behavior doc corrections, un-skip stale tests, temp-dir leak, delete dead `src/graph/*` + sandbox shim, single-source `redactShellCommand`.

## Invariant audit
- 1 (plugin init): touched — M12 adds a `{{KEY}}` leak check inside `getAgentConfigs` (init path). It is **fail-open** (review F-001): the assertion throws only for OUR built-in prompts (caught in CI by a warning-buffer test), but at the production call site a user-authored custom prompt that legitimately contains a literal `{{KEY}}` is downgraded to a deferred warning and the agent is still registered — init never aborts. `repro-704` T1 = 64ms ≪ 400ms deadline; `bun run build` + `node -e "import('./dist/index.js')"` OK.
- 2 (runtime portability): touched — `src/index.ts` (M9), `src/plan/ledger.ts` (M1 uses `node:fs` openSync/fsyncSync). No top-level `bun:` imports / `Bun.*` added. `bun run build` OK; dist import OK; `bunx biome ci .` exit 0.
- 3 (subprocesses): not touched — the `bunSpawn` calls in `pkg-audit.ts` are pre-existing and unchanged (MH1 only added fields to return objects). `git diff` shows no spawn-arg changes.
- 4 (.swarm containment): touched — M1 quarantine writes a unique file alongside the ledger under `.swarm/`; no new `process.cwd()` callers; L2 test switched to `os.tmpdir()`.
- 5 (plan durability): touched — M1 hardens ledger replay/recovery; no plan-schema or status-shape change (the `_ledgerReplayStale` marker is pre-existing and already consumed). Regression tests added (`ledger-m1-truncation.test.ts`); non-destructive recovery verified.
- 6 (test_runner safety): not touched — `test-runner.ts` unchanged.
- 7 (test writing): touched — new/changed tests use `bun:test`, `os.tmpdir()`+`mkdtemp`, DI/property seams (not `mock.module`), and `freezeClock` where a real clock is read. `check-mock-cleanup.sh` and `check-test-clock.sh` report 0 new violations; all new files < 500 lines.
- 8 (session state): touched — M9 FIFO-caps both session maps at `MAX_TRACKED_SESSIONS=500`, keyed by sessionID; tests assert the cap + current-session preservation.
- 9 (guardrails/retry): touched — L4 single-sources `redactShellCommand` to the home-path-redacting copy (re-export only; no retry-semantics change).
- 10 (chat/system msg): touched — M10 sanitizes learned content injected into `output.system` (agent-context now at parity with decisions); no new system entries and the single-system-message collapse is unchanged.
- 11 (tool registration): not touched — no tool added/removed; `check-tool-registration.ts` passes (105 tools coherent); `drift:check` clean. (M4 changes a full-auto policy tier, not the tool map.)
- 12 (release/cache): touched — M5/M6 harden cache/uninstall deletes (realpath + guards); release fragment `docs/releases/pending/1779-audit-medium-low-defects.md` added; version files untouched.

## Test plan
- [x] `bunx tsc --noEmit` → exit 0
- [x] `bun run build` + `node -e "await import('./dist/index.js')"` → OK
- [x] `node scripts/repro-704.mjs` → T1 64ms (deadline 400ms)
- [x] `bun run lint:ci` (biome ci) → exit 0
- [x] `bun run scripts/check-tool-registration.ts` → 105 tools coherent
- [x] `check-mock-cleanup.sh` / `check-invariants.sh` / `check-cross-contamination.sh` / `check-test-clock.sh` → 0 new violations
- [x] `bun run drift:check` → no drift
- [x] `bun run test:unit:ci <changed test files>` → exit 0
- [x] Independent implementation review (2 reviewers) + final critic on the real diff → APPROVE (2 blockers found and fixed: M14 `initialized` reset, Biome format)
- [x] CI green on the full unit (6 shards) / integration / security / adversarial / smoke tiers (run 29242592737)
- [x] Review follow-up (F-001/F-002/F-004/F-005) landed with Stage A + Stage B (reviewer + test_engineer) + closeout (reviewer + critic) all APPROVE — see the closure-ledger comment

## Pre-existing failures
- `bun test src/plan/` shows 17 failures and the pkg-audit combined run shows 1, both from `mock.module` cross-file pollution documented in `tests/TAXONOMY.md`. Verified identical on clean HEAD (stash baseline: 112/17 and 343/1) — not introduced by this change; every affected file passes in isolation.


zaxbysauce:
## Review feedback — closure ledger (commit `266a487`)

Every finding from the review was verified as a claim against source (3 independent verification lanes), then confirmed fixes went through Stage A (build/typecheck/lint/repro-704), Stage B (independent reviewer + test_engineer), and closeout (independent reviewer + final critic) — all APPROVE, no blockers.

| ID | Severity | Verdict | Outcome |
|---|---|---|---|
| CI-001/003 | blocking | CONFIRMED | **Fixed** `a4c4378` — M1 rebuildPlan fsync broke test mocks (`bunWrite` no longer called for plan.json) + missing `.swarm` precreate. CI now green. |
| CI-002 | blocking | CONFIRMED | **Fixed** `1b24c1a` — null-byte adversarial test asserted the pre-M10 vulnerability; inverted to assert the fix. |
| F-001 | HIGH | **CONFIRMED (regression)** | **Fixed** `266a487` — a user custom prompt with a literal `{{UPPER}}` reached the M12 assertion (custom prompts pass verbatim) and threw through the fail-CLOSED init catch → whole plugin dropped. Now fail-open: warn + register the agent anyway; assertion still throws for tests; built-in leaks still caught in CI. False `withTimeout(2000ms)` docstring corrected. |
| F-002 | HIGH→PARTIAL | data-loss DISPROVED | **Fixed** `266a487` — the rebuild-cycle/data-loss framing is wrong (the truncated path preserves plan.json, no rebuild). Real residue: identical corruption re-quarantined once per process restart. Added content-hash + byte-compare dedup. |
| F-003 | HIGH→PRE_EXISTING | no change | Staleness marker/blocking/self-heal all pre-exist on `main`; the PR only routes the truncated case into the intended (critic-accepted) marker, which auto-clears on plan↔ledger reconvergence. |
| F-004 | MEDIUM→PARTIAL | PARTIAL | **Fixed** `266a487` — 931/937 read architect-authored `plan.md` (trusted, out of M10 scope). Line 1022/1880 `agentContext` (from `context.md`, echoes tool output) is a real parity gap vs its sanitized sibling `decisions` → now sanitized at both sites. |
| F-005 | MEDIUM→PRE_EXISTING | hardening | **Fixed** `266a487` — widened `sanitizeContextText`'s invisible-char coverage to match `INVISIBLE_FORMAT_CHARS` (adds U+00AD, U+200E–200F, U+2060–2064). Cheap hardening for the now load-bearing M10 sanitizer. |
| F-006 | MEDIUM | **DISPROVED** | No change — the `</\w+>` catch-all is pre-existing, intentional defense-in-depth (neutralizes arbitrary closing-tag injectors); removing it would weaken injection coverage. |
| F-007..F-014 | advisory | assessed | No code change except **F-009** (quarantine accumulation), which is subsumed by the F-002 dedup. F-010 (savePlan fsync) kept advisory — plan.json is a projection of the fsynced authoritative ledger and self-heals; F-011 already in the release note; F-014 (digit placeholders) moot (no digit keys exist) and now fail-open. |

**Validation:** `tsc` ✓ · `biome ci` ✓ · `build` + dist import ✓ · `repro-704` 64ms ✓ · affected tests 65 pass / 0 fail (incl. added Path-B, over-stripping, and concurrency falsification probes).

---

Issue #1729, opened 2026-07-05
Make CI actually valid: pay down all test quarantines (macOS/Windows/integration + coverage) — consolidates #1705 and #1712


zaxbysauce:
## Goal

Make CI **actually valid** — i.e. drive **every** quarantine list to zero by fixing the *root cause* of each skipped test (or, where a test over-specifies OS-specific behavior, making the test portable), **never by masking**. When done, no test is "green by not running," and the `coverage` gate is a real required check.

This issue **consolidates and supersedes**:
- **#1705** — per-file deterministic debt (base `quarantined-tests.txt`) + CI-environment-flaky tests.
- **#1712** — the `coverage` job's combined-process mock/state contamination.

Both are folded into the inventory below with their current status. Close them in favor of this one.

---

## 0. Context an implementer needs first

PR #1700 (issue #1683) made CI failure detection **exit-code-primary** instead of grepping for a bun-version output marker — the bug that let failing tests report green on `main` for an unknown period. Honest detection then surfaced a large backlog of pre-existing failures. Rather than go permanently red, CI **skips tracked quarantine lists** and the debt is paid down incrementally. A stacked follow-up (#1726) fixed and un-quarantined the bulk of the ubuntu-visible debt, but also **blindly emptied the merge_group-only lists** (see the critical constraint below), which had to be restored.

### ⚠️ THE CRITICAL CONSTRAINT — read this before touching any list

CI runs in two tiers:
- **`pull_request` tier** — **ubuntu-latest only**, `unit` job (4 shards) + fast checks. This is what you see go green on a normal PR.
- **`merge_group` tier** (merge queue) — the **full 3-OS matrix** (ubuntu/macOS/Windows) `unit` shards, **plus** the `integration`, `coverage`, and `smoke` jobs, which are **skipped entirely on `pull_request`** (`if: github.event_name == 'merge_group'`).

**Consequence:** a fix to a macOS-only, Windows-only, integration, or coverage failure **cannot be validated by a normal green PR check** — that check never ran those tiers. This is exactly how #1726 regressed: it un-quarantined macOS/Windows/integration tests, its ubuntu-only PR CI passed, and the merge queue then bounced four times. **Do not trust a green `pull_request` check for anything in §2.2–§2.4. Validate in the tier that actually runs the test (see §5).**

---

## 1. How the quarantine plumbing works (so you know how to un-quarantine)

`.github/workflows/ci.yml`:
[...]

PR #1843, opened 2026-07-14
fix(tests): pay down knowledge-test quarantine debt and add CI portability lints (#1737)

### LINKED ISSUE #1737: Pay down remaining test quarantine debt + structural CI improvements to prevent recurrence
## Context

PR #1730 (issue #1729) paid down the bulk of the CI quarantine debt — fixing the production `save_plan` projection bug, resolving ~40 cross-platform test failures, and driving 3 of 4 quarantine lists to zero. However, **11 entries remain across 3 lists**, and the effort exposed systemic gaps that allowed the debt to accumulate undetected for months. This issue tracks both the remaining debt and the structural fixes needed to prevent recurrence.

## Current quarantine state (on main post-#1730)

| List | Active entries | Root cause category |
|---|---|---|
| quarantined-tests.txt | 5 | Knowledge-system test failures (pre-existing, unrelated to #1729) |
| quarantined-tests-macos.txt | 4 | Lean-turbo worktree provisioning — macOS-only lane mock issue |
| quarantined-tests-windows.txt | 0 | Clean |
| quarantined-integration-tests.txt | 2 | Ubuntu CI-only: read-only fs semantics + injection-budget shared-state |

---

### A1. macOS lean-turbo worktree tests (4 files)

**Files:**
```
tests/unit/turbo/lean/integration-worktree.test.ts
tests/unit/turbo/lean/runner.test.ts
tests/unit/turbo/lean/runner.adversarial.test.ts
tests/unit/turbo/lean/retroactive-adversarial.test.ts
```

**Failure signature (from merge_group run 28806176287):**
```
expect(received).toBeGreaterThan(expected)
Expected: > 0
Received: 0
```
The tests mock `provisionWorktree` via `_internals.bunSpawn` and assert it was called at least 1 time during `runPhase`. On macOS it's called **0 times** — `runPhase` takes a different code path. These tests pass on ubuntu and Windows.

**Suspected root causes to investigate:**
- `checkPathBudget` (src/worktree/core.ts:326) may behave differently on macOS, changing the provisioning path. It's Windows-only in intent but may short-circuit on non-Windows in an unexpected way.
- `git worktree add` porcelain output format or branch creation semantics may differ on the macOS runner's git version, causing the mock to not match.
- A race condition that only manifests on the macOS runner's slower I/O — the `provisionWorktree` call may be async and the mock may not be installed by the time it runs.
- The tests construct lane evidence files that feed `listLaneEvidence`; the sort added in #1730 made ordering deterministic, but the mock setup may still depend on filesystem timing.

**Blocker:** requires macOS environment access (or detailed merge_group log analysis across multiple runs). Cannot be diagnosed on Linux or Windows.

**Validation:** merge_group run with `unit (macos-latest, N)` all 4 shards passing these files.

### A2. Integration tests — ubuntu CI-only (2 files)

**Files:**
```
tests/integration/phase-complete-events.adversarial.test.ts
tests/integration/unified-injection-budget.test.ts
```

#### A2a. phase-complete-events.adversarial.test.ts:393 — read-only .swarm directory

**Failure signature:**
```
expect(parsed.success).toBe(true)
Expected: true
Received: false
```
The test makes `.swarm` read-only via `chmod 444` and expects `phase_complete` to still succeed (write failure should be non-blocking). On the ubuntu CI runner, `success` is `false`. Passes on Windows and macOS.

**Suspected root cause:** the ubuntu runner's uid/fs semantics — the runner user may be root (which bypasses read-only permissions), OR the `chmod 444` on `.swarm` doesn't prevent the specific write path the way the test expects. The test should use a more robust read-only simulation (e.g. `chmod 555` on the parent, or mock the write).

#### A2b. unified-injection-budget.test.ts:312 — SC-005 (Production bug #2)

**Failure signature:**
```
expect(kiInjected).toBeUndefined()
Received: { info: { role: "system" }, parts: [{ type: "text", text: "[[KNOWLEDGE-INJECTED]]..." }] }
```
When `system-enhancer` alone exceeds the unified injection ceiling, `knowledge-injector` should get **0** allocation and not inject. But it still injects on the ubuntu CI runner. This is the **real Production bug #2** from issue #1729 section 3.4.

**Critical note:** this test passes in isolation on Windows and in a clean `origin/main` worktree. It only fails in the ubuntu CI merge_group environment. The suspected cause is **test-ordering/shared-state contamination** — `getSystemEnhancerDemand(sessionID)` reads from a session-keyed map that may not be populated correctly when tests run in a specific order, or the session ID isn't unique between tests.

**Source locations to investigate:**
- src/hooks/knowledge-injector.ts:707-712 — `allocateInjectionBudget(seDemand, effectiveBudget, ...)` call
- src/hooks/system-enhancer.ts:684-691, 1645-1651, 2479-2485 — `resetUnifiedBudget` / demand recording
- Determine whether SE demand is actually recorded for the session before KI reads it in this test flow

**This may be a real production bug** (not just a test issue): if SE demand isn't recorded under certain session lifecycles, KI would inject beyond the unified budget in production too.

### A3. Base quarantine — knowledge-system tests (5 files, pre-existing)

**Files:**
```
tests/unit/commands/cleanup-drift.test.ts
tests/unit/commands/knowledge.adversarial.test.ts
tests/unit/hooks/knowledge-injector-drift-adversarial.test.ts
tests/unit/hooks/knowledge-floor-action.test.ts
tests/unit/tools/knowledge-recall.test.ts
```

These were quarantined before/during the #1729 effort (commits a5879ccb, f491360a) and are unrelated to the cross-platform work. They're knowledge-system failures likely related to the #1716/#1717 knowledge lifecycle changes. Triage separately — they may already be fixed by the knowledge PRs that landed (#1722, #1727, #1730's merge brought those in).

---

## Part B — Structural CI improvements to prevent recurrence

The #1729 effort took ~10 merge_group bounces over 2 days because the `pull_request` CI tier is ubuntu-only and cannot surface macOS/Windows/integration/coverage failures. These improvements would have caught every bounce before queuing:

### B1. Shared canonicalTmpDir() test helper + lint enforcement

~20 test files were individually fixed with `fs.realpathSync(os.tmpdir())` for the macOS /var to /private/var symlink. Without a shared helper, new tests will reintroduce the bug.

- Create `tests/helpers/tmpdir.ts` exporting `canonicalTmpDir()` and `canonicalMkdtemp(prefix)`.
- Add `scripts/check-test-tmpdir.sh` (modeled on `check-mock-cleanup.sh`) that flags raw `os.tmpdir()` / `mkdtemp` usage in test files without the realpath wrap.
- Wire into the `quality` CI job.

### B2. Bash portability lint for scripts/

`check-invariants.sh` used `declare -A` (bash 4+) which silently died on macOS bash 3.2 for months. A lint script would catch this class of failure:

- Create `scripts/check-bash-portability.sh` that flags bash-4+ constructs: `declare -A`, associative-array subscript syntax, `grep -P`, hex escapes like `\x27`, `coproc`, `mapfile`/`readarray`.
- Wire into the `quality` CI job.

### B3. Cross-OS PR-tier validation (the biggest gap)

The root cause of the entire backlog: `pull_request` CI runs ubuntu-only, so macOS/Windows-only failures are invisible until the merge queue bounces them.

**Option A — workflow_dispatch full-matrix trigger:** add a `workflow_dispatch` trigger to ci.yml that lets PR authors manually trigger the full 3-OS matrix + integration + coverage on their branch before queuing. Cost: GitHub Action minutes, but prevents queue bounces.

**Option B — "platform-sensitive" path filter:** run macOS + Windows unit shards on `pull_request` only when the diff touches platform-sensitive paths (src/worktree/**, src/sandbox/**, src/plan/**, scripts/**, .github/workflows/**). Use a `paths` filter or a `dorny/paths-filter` action.

**Option C — Previously-quarantined smoke job:** add a fast `pull_request` job that runs just the previously-quarantined files on all 3 OSes (a small subset, not the full 4-shard suite). This catches regressions in the files most likely to have cross-platform issues without the full cost.

### B4. Coverage-gate docs reconciliation

Coverage is already a required check (ruleset 17809658, added 2026-06-30), but the ci.yml comment at ~L245-249 still says "must happen AFTER this lands on main." Update the comment to reflect that coverage is enforced.

---

## Recommended order of attack

1. **A3** (base quarantine, 5 knowledge tests) — re-triage each; may already pass after the knowledge PRs merged. Lowest effort, highest chance of quick wins.
2. **B1 + B2** (shared helper + bash lint) — quick to implement, prevents the most common recurrence patterns. ~2-3 hrs each.
3. **A2b** (injection budget SC-005) — likely a real production bug. Reproduce by running the test file in the same bun process as the other integration tests (simulate the CI ordering). ~2-4 hrs.
4. **A2a** (read-only .swarm) — fix the read-only simulation to be uid-independent. ~1 hr.
5. **B3** (cross-OS PR-tier validation) — the systemic fix. ~3-5 hrs depending on option chosen.
6. **A1** (macOS lean worktree) — blocked on macOS access. ~2-4 hrs with a Mac.

## Definition of done

- [ ] All 4 quarantine lists at 0 active entries
- [ ] B1 (canonicalTmpdir helper + lint) implemented and wired into CI
- [ ] B2 (bash portability lint) implemented and wired into CI
- [ ] B3 (cross-OS PR-tier validation) implemented — at minimum Option A or C
- [ ] B4 (coverage docs reconciled)
- [ ] No test is green-by-not-running (verified by checking all 4 lists return empty for non-comment lines)

## Related

- Closes the residual debt from #1729 (PR #1730).
- Supersedes the quarantine tracking in #1705 (now closed).
- A2b traces to #1729 section 3.4 "Production bug #2" (unified injection budget).

zaxbysauce:
Partially addresses #1737

## Summary

- Fixed five genuinely broken knowledge-subsystem tests that had been quarantined instead of fixed (`cleanup-drift.test.ts`, `knowledge.adversarial.test.ts`, `knowledge-injector-drift-adversarial.test.ts`, `knowledge-recall.test.ts` x2 issues), and un-quarantined them from `scripts/ci/quarantined-tests.txt`.
- Fixed `tests/integration/phase-complete-events.adversarial.test.ts`'s "read-only `.swarm` directory" coverage to match the actual documented split-fail behavior (`events.jsonl` soft-fails with a warning, `plan.json` fails closed).
- Fixed a real bug in `src/hooks/system-enhancer.ts`: the unified-injection-budget ledger write could be skipped on a mid-flight exception; it's now guaranteed via a `finally` block, with a new regression test.
- Added `tests/helpers/tmpdir.ts` (macOS `/var` → `/private/var` symlink-safe helpers) and `scripts/check-test-tmpdir.sh`, a diff-scoped lint catching raw `tmpdir()` usage in new/changed test-file lines.
- Added `scripts/check-bash-portability.sh`, a full-scan lint for bash-4+-only constructs (`declare/typeset/local/readonly -A`, `grep -P`/`--perl-regexp`, `coproc`, `mapfile`/`readarray`) that silently break on macOS's system bash (3.2) — prevents recurrence of the class of bug fixed in issue #1729. Fixed one genuine pre-existing PCRE violation this lint caught in `scripts/check-cross-contamination.sh`.
- `.github/workflows/ci.yml`: added a `detect-paths` job that path-scopes the cross-OS (macOS/Windows) unit-test matrix to PRs touching platform-sensitive paths (`src/worktree/`, `src/turbo/`, `src/sandbox/`, `src/plan/`, `src/parallel/`, `scripts/`, `.github/workflows/`) or `merge_group` runs. Added a `unit-passed` aggregate check job. Corrected a stale comment about the `coverage` check's required-status state.
- Fixed a genuine SIGPIPE correctness bug caught during pre-push validation: the `detect-paths` step originally used `grep -qE` piped from `git diff`, which under `set -o pipefail` can make the `if` silently false on a large diff even when a match occurred (dropping platform-sensitive PRs to the ubuntu-only matrix — the unsafe direction). Fixed by using `grep -E ... >/dev/null` (no `-q`/`-m`, reads to EOF, cannot race). Verified empirically against a 100M-line producer.
- Refined `tests/security/ci-workflow-security.test.ts`'s expression-injection check to recognize GitHub's own recommended "intermediate environment variable" mitigation (a dangerous `github.event.*` context assigned to an `env:` block value, rather than interpolated inline in a `run:` script, is not itself a vulnerability) — needed because the new `detect-paths` job passes `github.event.pull_request.base.sha`/`head.sha` (immutable git SHAs) via `env:`.
- Un-quarantined `tests/unit/hooks/knowledge-floor-action.test.ts` and `tests/unit/turbo/lean/integration-worktree.test.ts` and fixed the two genuine production bugs their failures were exposing (not test bugs — see below).

## Two additional production bugs found and fixed during CI validation

Both of these were surfaced by un-quarantining tests that had been failing CI on `ubuntu-latest`/`windows-latest` shards. Both root-caused to real defects, not test issues:

1. **`swarm-artifact-cache.ts` lost-update bug**: the shared read cache (keyed on mtime+ctime+size) never invalidated on write. A locked read-modify-write immediately following an `atomicWriteFile` write to the same path can hit a same-tick, same-size stat collision and read back stale pre-write content, then persist it — silently clobbering the just-written value. Reachable any time two locked writes to the same file happen in quick succession (e.g. `bumpKnowledgeConfidenceBatch`'s delta-write followed by its confidence-floor-flag transaction). Fixed by invalidating the cache entry from the shared `atomicWriteFile` primitive (`src/evidence/task-file.ts`) after every successful write — closes the gap for all callers, not just this one path.
2. **Lean Turbo merge-conflict misclassification**: `attemptMergeBackFromDirty`'s real conflict result (`src/worktree/merge.ts`) never sets a `conflict: true` field — only `partial: true` + non-empty `conflictFiles`. `runner.ts`'s dirty-worktree cleanup path checked for `conflict === true` to classify a failure as `'conflict'` vs `'partial'`, which is unreachable for this call path — every real merge conflict reaching `_sequentialWorktreeCleanup` was being reported as `'partial'` in production. Fixed the classifier to key off non-empty `conflictFiles`.

Both reproduced deterministically (10/10 runs) on a clean `origin/main` checkout before the fix, and pass deterministically after it.

## Invariant audit
- 1 (plugin init): not touched - no changes to `src/index.ts` or plugin entry/init path.
- 2 (runtime portability): not touched - no `Bun.*`/`bun:` imports added; `bun run build` succeeds and `node --input-type=module -e "await import('./dist/index.js')"` confirms the bundle loads under Node.
- 3 (subprocesses): not touched - `git diff --name-only origin/main..HEAD | xargs -r grep -nE "bunSpawn\(|spawn\(|spawnSync\("` returns no matches in changed source files; new shell (`.sh`) lint scripts run via bash CI steps, not Node subprocess spawns.
- 4 (.swarm containment): touched - `tests/integration/phase-complete-events.adversarial.test.ts`'s read-only-`.swarm` coverage was corrected to match actual split-fail behavior (soft-fail `events.jsonl`, fail-closed `plan.json`); no `.swarm/` write-path code itself was changed, only test assertions.
- 6 (test_runner safety): not touched - all validation used shell commands (`bun run test:unit:ci`, `bun test`) per AGENTS.md #6, not the `test_runner` tool with broad scope.
- 7 (test writing): touched - all new/changed tests use `bun:test` only; `tests/unit/hooks/system-enhancer-budget-ledger-finally.test.ts` uses a marker-gated mock to avoid cross-test pollution rather than a leaking `mock.module()`; `scripts/check-mock-cleanup.sh` passes clean.
- 11 (tool registration): not touched - `bun run scripts/check-tool-registration.ts` passes ("106 tools, coherent").
- 12 (release/cache): not touched - no hand-edits to `package.json#version`, `CHANGELOG.md`, or `.release-please-manifest.json`; `docs/releases/pending/1737-quarantine-debt-and-ci-portability-lints.md` fragment added per contract.

## Test plan

- [x] `bun run build` succeeds; `node --input-type=module -e "await import('./dist/index.js')"` confirms bundle loads.
- [x] `bun run typecheck` clean.
- [x] `bun run lint:ci` equivalent (biome scoped to tracked `src/`, `tests/`, `test/`, config files) clean, 0 issues.
- [x] `bun run scripts/check-tool-registration.ts`, `check-mock-cleanup.sh`, `check-invariants.sh`, `check-cross-contamination.sh`, `check-test-clock.sh` all pass.
- [x] New lints (`check-test-tmpdir.sh`, `check-bash-portability.sh`) pass against current repo state.
- [x] `bun test tests/unit/hooks/knowledge-floor-action.test.ts` and `tests/unit/turbo/lean/integration-worktree.test.ts` — both pass deterministically across repeated runs (previously failed deterministically on clean `origin/main`).
- [x] `bun test tests/adversarial --timeout 120000` — 829 pass, 0 fail.
- [x] `bun test tests/security --timeout 120000` — 163 pass, 4 skip, 0 fail.
- [x] `bun test tests/smoke --timeout 120000` — 10 pass, 0 fail.
- [x] `bun test tests/integration ./test --timeout 120000` — 746 pass, 199 fail; all 199 failures independently confirmed pre-existing on clean `origin/main` (see Pre-existing failures).
- [x] All CI checks green at head (`unit` all shards, `unit-passed`, `security`, `quality`, `detect-paths`, `package-check`, `rust-sandbox-runner`, `php-validation`).
- [x] YAML syntax of `.github/workflows/ci.yml` validated via `Bun.YAML.parse`.
- [x] All third-party `uses:` in the diff are SHA-pinned.
- [x] Push-protection secret-pattern scan of the full commit range: clean.

## Pre-existing failures

- `bun test tests/integration ./test --timeout 120000`: 199 failures across ~15 unrelated subsystems (Phase Preflight Auto-Trigger, full-auto mode, curator v3, Gate Workflow, phase_complete adversarial/E2E, skill-feedback, macro/micro reflector, plan-persistence Layer 1↔2 bridge, etc.). None of these files are touched by this PR. Reproduced identically (746 pass / 199 fail) on a clean `git worktree add origin/main` + `bun install --frozen-lockfile`, confirming this is pre-existing repo state, not a regression from this branch.

## Known caveats / remaining #1737 scope

This PR does **not** close #1737 — 9 quarantine entries remain across 3 lists after this PR:
- `scripts/ci/quarantined-tests.txt`: 3 entries (swarm-artifact-cache, skill-scoring-workflow, test-runner)
- `scripts/ci/quarantined-tests-macos.txt`: 4 entries (3 lean-turbo tests + macOS `provisionWorktree`)
- `scripts/ci/quarantined-integration-tests.txt`: 2 entries (unified-injection-budget, one other)
- `scripts/ci/quarantined-tests-windows.txt`: 0 entries

Documented in the release-note fragment (`docs/releases/pending/1737-quarantine-debt-and-ci-portability-lints.md`):
- The new `unit-passed` aggregate CI job is intentionally not yet wired into branch protection (would break the merge queue for other open PRs since the check doesn't exist on `main` yet) — deferred to a follow-up once this lands.
- The remaining 9 quarantine entries (macOS lean-turbo, injection-budget, macOS `provisionWorktree` root-cause diagnosis) require further investigation and are intentionally deferred to follow-up work under #1737, which stays open.

<a href="https://cubic.dev/pr/ZaxbyHub/opencode-swarm/pull/1843?utm_source=github" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"&gt;``</a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,8 +86,15 @@ jobs:
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         if: github.event_name == 'pull_request'
         with:
-          # Full history needed to diff base...head reliably (shallow clones
-          # can miss the merge-base). Matches drift-check.yml's convention.
+          # Full history needed so both BASE_SHA and HEAD_SHA below are
+          # guaranteed present locally for a plain two-dot `git diff` (a
+          # shallow clone can miss one of them, especially BASE_SHA on a
+          # long-lived PR). Two-dot, not three-dot/merge-base: this only
+          # needs "what changed between these two exact commits," and
+          # two-dot's tendency to also include unrelated upstream churn is
+          # the safe-direction failure mode here — it can only make this
+          # detector MORE likely to opt into the fuller matrix, never less.
+          # Matches drift-check.yml's fetch-depth: 0 convention.
           fetch-depth: 0
       # Single unconditional step (not split by event type) so the job-level
       # `outputs:` above always resolves from this one step's ID, regardless
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,22 +112,31 @@ jobs:
           set -euo pipefail
           # Only pull_request has a meaningful base...head diff to inspect
           # here; merge_group already gets the full 3-OS matrix unconditionally
-          # via the separate `github.event_name == 'merge_group'` branch in the
+          # via the separate github.event_name == 'merge_group' branch in the
           # unit job's matrix expression, so this output doesn't need to (and
           # can't, without a PR base/head pair) do anything for that event.
           if [ "$EVENT" != "pull_request" ]; then
             echo "touches-platform-paths=false" >> "$GITHUB_OUTPUT"
             exit 0
           fi
-          changed="$(git diff --name-only "$BASE_SHA" "$HEAD_SHA" 2>/dev/null || echo "")"
           # Platform-sensitive: worktree/sandbox/plan isolation logic, the lean
           # turbo runner (where issue #1737's A1 macOS-only bug lives),
           # src/parallel (lock-directory creation / EACCES semantics — the
           # confirmed root cause of A2a's read-only-.swarm cross-platform
           # bug, added after a test-engineer review caught it missing from
           # the original 5-prefix list), CI scripts, and the workflow files
-          # themselves.
-          if echo "$changed" | grep -qE '^(src/worktree/|src/turbo/|src/sandbox/|src/plan/|src/parallel/|scripts/|\.github/workflows/)'; then
+          # themselves. Piped directly rather than captured into a variable
+          # first. Deliberately grep -E redirected to /dev/null below, and
+          # NOT grep -q: under set -o pipefail, grep -q exits as soon as it
+          # sees the first match and closes its stdin, which can send
+          # SIGPIPE to a still-writing git diff upstream on a large diff —
+          # pipefail then reports that SIGPIPE (128+13=141) as the
+          # pipeline's exit status, making this if silently false even
+          # though a match occurred (the unsafe direction — it would drop
+          # platform-path PRs to the ubuntu-only matrix). grep without -q
+          # or -m reads its input to EOF before exiting, so git diff always
+          # finishes writing normally and no SIGPIPE is possible.
+          if git diff --name-only "$BASE_SHA" "$HEAD_SHA" 2>/dev/null | grep -E '^(src/worktree/|src/turbo/|src/sandbox/|src/plan/|src/parallel/|scripts/|\.github/workflows/)' >/dev/null; then
             echo "touches-platform-paths=true" >> "$GITHUB_OUTPUT"
           else
             echo "touches-platform-paths=false" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,6 +118,24 @@ jobs:
         if: needs.detect-release.outputs.is-release != 'true'
         shell: bash
         run: chmod +x scripts/check-test-file-cap.sh && bash scripts/check-test-file-cap.sh
+      # Line-scoped (not file-scoped): fails only on NEW/changed lines that add
+      # a raw tmpdir() call not wrapped in realpathSync. Pre-existing raw calls
+      # elsewhere in a touched file are not flagged (issue #1737 FR-011 scope
+      # decision — retrofitting the ~250 pre-existing occurrences is out of
+      # scope). Prefer tests/helpers/tmpdir.ts (canonicalTmpDir/canonicalMkdtemp).
+      - name: Test tmpdir canonicalization lint (FR-011)
+        if: needs.detect-release.outputs.is-release != 'true'
+        shell: bash
+        run: chmod +x scripts/check-test-tmpdir.sh && bash scripts/check-test-tmpdir.sh
+      # Not diff-scoped (unlike the sibling checks above): a bash4+-only
+      # construct is exactly as broken on macOS as it was the day it was
+      # added, so this scans full current file content in scripts/, not just
+      # the diff (issue #1737 FR-012 — prevents recurrence of the
+      # check-invariants.sh `declare -A` incident, issue #1729).
+      - name: Bash portability lint (FR-012)
+        if: needs.detect-release.outputs.is-release != 'true'
+        shell: bash
+        run: chmod +x scripts/check-bash-portability.sh && bash scripts/check-bash-portability.sh
 
   # Two-tier CI: on pull_request, run ubuntu-only with 4 shards for fast feedback.
   # On merge_group (merge queue), run the full 3-OS matrix for merge safety.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,70 @@ jobs:
           BRANCH: ${{ github.head_ref }}
           EVENT: ${{ github.event_name }}
 
+  # Issue #1737 FR-013: closes the root cause of the #1729 debt-accumulation
+  # episode — pull_request CI ran ubuntu-only, so macOS/Windows-only failures
+  # were invisible until a merge_group bounce. This job detects whether the PR
+  # diff touches platform-sensitive paths and, if so, the `unit` job below
+  # automatically runs the full 3-OS matrix on pull_request too (previously
+  # only merge_group did). Modeled on the detect-release job above (fast,
+  # single ubuntu-latest job producing a boolean output the matrix consumes).
+  # No job-level `if:` here (unlike detect-release's own guard on its
+  # checkout step, not itself) — matching the established quality/unit
+  # pattern where downstream jobs never skip at the job level, only steps
+  # skip individually. A job-level `if:` here would report this job as
+  # "skipped" (not "success") on release-please branches, and by default
+  # `needs:` treats a skipped dependency as not-satisfied, which would
+  # cascade to skip the entire `unit` job on every release branch — release
+  # branches would then never even report the required `unit (*, N)` checks.
+  detect-paths:
+    needs: [detect-release]
+    runs-on: ubuntu-latest
+    outputs:
+      touches-platform-paths: ${{ steps.check.outputs.touches-platform-paths }}
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+        if: github.event_name == 'pull_request'
+        with:
+          # Full history needed to diff base...head reliably (shallow clones
+          # can miss the merge-base). Matches drift-check.yml's convention.
+          fetch-depth: 0
+      # Single unconditional step (not split by event type) so the job-level
+      # `outputs:` above always resolves from this one step's ID, regardless
+      # of event — a prior split-step version left `steps.check` unset (and
+      # the job output empty) on non-pull_request events, since a step whose
+      # own `if:` is false produces no outputs at all, not empty-string ones.
+      - name: Detect platform-sensitive path changes
+        id: check
+        shell: bash
+        env:
+          EVENT: ${{ github.event_name }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+          # Only pull_request has a meaningful base...head diff to inspect
+          # here; merge_group already gets the full 3-OS matrix unconditionally
+          # via the separate `github.event_name == 'merge_group'` branch in the
+          # unit job's matrix expression, so this output doesn't need to (and
+          # can't, without a PR base/head pair) do anything for that event.
+          if [ "$EVENT" != "pull_request" ]; then
+            echo "touches-platform-paths=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          changed="$(git diff --name-only "$BASE_SHA" "$HEAD_SHA" 2>/dev/null || echo "")"
+          # Platform-sensitive: worktree/sandbox/plan isolation logic, the lean
+          # turbo runner (where issue #1737's A1 macOS-only bug lives),
+          # src/parallel (lock-directory creation / EACCES semantics — the
+          # confirmed root cause of A2a's read-only-.swarm cross-platform
+          # bug, added after a test-engineer review caught it missing from
+          # the original 5-prefix list), CI scripts, and the workflow files
+          # themselves.
+          if echo "$changed" | grep -qE '^(src/worktree/|src/turbo/|src/sandbox/|src/plan/|src/parallel/|scripts/|\.github/workflows/)'; then
+            echo "touches-platform-paths=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "touches-platform-paths=false" >> "$GITHUB_OUTPUT"
+          fi
+
   quality:
     needs: [detect-release]
     runs-on: ubuntu-latest
@@ -137,18 +201,21 @@ jobs:
         shell: bash
         run: chmod +x scripts/check-bash-portability.sh && bash scripts/check-bash-portability.sh
 
-  # Two-tier CI: on pull_request, run ubuntu-only with 4 shards for fast feedback.
-  # On merge_group (merge queue), run the full 3-OS matrix for merge safety.
-  # On release-please branches, all steps are skipped via detect-release job
-  # (version bumps don't need tests — applies to both PR and merge-queue tiers).
-  # Test files are discovered automatically via find and distributed round-robin
-  # across shards — no hardcoded file lists, no duplication.
+  # Two-tier CI: on pull_request, run ubuntu-only with 4 shards for fast feedback
+  # UNLESS the diff touches a platform-sensitive path (issue #1737 FR-013,
+  # detect-paths job above), in which case the full 3-OS matrix runs on
+  # pull_request too. On merge_group (merge queue), the full 3-OS matrix
+  # always runs for merge safety. On release-please branches, all steps are
+  # skipped via detect-release job (version bumps don't need tests — applies
+  # to both PR and merge-queue tiers). Test files are discovered automatically
+  # via find and distributed round-robin across shards — no hardcoded file
+  # lists, no duplication.
   unit:
-    needs: [detect-release, quality]
+    needs: [detect-release, quality, detect-paths]
     strategy:
       fail-fast: false
       matrix:
-        os: ${{ github.event_name == 'merge_group' && fromJSON('["ubuntu-latest","macos-latest","windows-latest"]') || fromJSON('["ubuntu-latest"]') }}
+        os: ${{ (github.event_name == 'merge_group' || needs.detect-paths.outputs.touches-platform-paths == 'true') && fromJSON('["ubuntu-latest","macos-latest","windows-latest"]') || fromJSON('["ubuntu-latest"]') }}
         shard: [1, 2, 3, 4, 5, 6]
     runs-on: ${{ matrix.os }}
     # 40m: was 90m before sharding (PR #1245). Issue #1778 H4 wired the
@@ -304,11 +371,33 @@ jobs:
       # budget — ~30-35 min in practice, intermittently timing out and blocking
       # the merge queue (PR #1573). It now lives in the dedicated `coverage` job
       # below so the full-suite run gets its own timeout and the unit cells stay
-      # fast. Blocking power is NOT automatic: nothing here `needs: coverage`, so
-      # the gate only blocks the queue once `coverage` is added to ruleset
-      # 17809658's required status checks. That must happen AFTER this lands on
-      # main (adding it earlier blocks other open PRs whose merge groups don't
-      # yet define this job). Until then the gate runs but does not block.
+      # fast. `coverage` is already enforced: it's a required status check in
+      # branch-protection ruleset 17809658 (confirmed active — issue #1737
+      # FR-014, correcting this comment, which previously and incorrectly
+      # described enforcement as still-pending).
+
+  # Issue #1737 FR-013: aggregate job so branch-protection ruleset 17809658 can
+  # require a single stable check name regardless of which OS cells the
+  # path-scoped conditional matrix above actually ran for a given PR — a
+  # per-cell required check (e.g. "unit (macos-latest, 1)") would never
+  # report at all on a PR where detect-paths decided macOS/Windows weren't
+  # needed, permanently blocking the merge queue for that PR. always() so this
+  # still runs (and can fail) even if a `unit` matrix cell failed upstream.
+  unit-passed:
+    needs: [unit]
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check unit job result
+        shell: bash
+        env:
+          UNIT_RESULT: ${{ needs.unit.result }}
+        run: |
+          if [ "$UNIT_RESULT" != "success" ]; then
+            echo "::error::unit job result was '$UNIT_RESULT', not 'success'"
+            exit 1
+          fi
+          echo "All unit matrix cells that ran succeeded."
 
   # Merge-queue-only coverage gate. Runs the full unit suite once with coverage
   # and enforces the line-coverage threshold. Isolated from the `unit` job so the
--- a/docs/releases/pending/1737-quarantine-debt-and-ci-portability-lints.md
+++ b/docs/releases/pending/1737-quarantine-debt-and-ci-portability-lints.md
@@ -37,6 +37,28 @@
   aggregate job that checks `needs.unit.result` across the full matrix (not yet
   wired into branch protection — see Known caveats). Corrected a stale comment
   claiming the `coverage` check was not yet a required status check.
+- **`src/utils/swarm-artifact-cache.ts` / `src/evidence/task-file.ts`:** fixed a
+  lost-update bug (issue #1729 class) where `atomicWriteFile`'s stat-based cache
+  never invalidated on write. A same-size rewrite landing within one filesystem
+  timestamp tick of a prior cached read could produce an identical
+  mtime+ctime+size stamp, so a second locked read-modify-write immediately
+  following a write (e.g. `bumpKnowledgeConfidenceBatch`'s confidence-delta write
+  followed by its confidence-floor-flag transaction) would silently read the
+  pre-write content and clobber the just-committed value back to stale data.
+  Added `invalidateCachedArtifact()`, called from `atomicWriteFile` after every
+  successful write, closing the gap for all ~15 callers of the shared primitive.
+  Root-caused via `tests/unit/hooks/knowledge-floor-action.test.ts`'s "clears
+  stale confidence_floor_demoted flag on recovery above floor" case, which failed
+  deterministically (not flaky) on both this branch and clean `origin/main`.
+- **`src/turbo/lean/runner.ts`:** fixed a merge-back failure misclassification —
+  `_sequentialWorktreeCleanup` distinguished `'conflict'` from `'partial'` status
+  via `mergeResult.conflict === true`, but `attemptMergeBackFromDirty`'s
+  `DirtyMergePartial` result (`src/worktree/merge.ts`) never sets a `conflict`
+  field for a real conflict; it signals one via non-empty `conflictFiles` only.
+  Every dirty-worktree merge conflict reaching this path was reported as
+  `'partial'` instead of `'conflict'`. Root-caused via
+  `tests/unit/turbo/lean/integration-worktree.test.ts` AC-6, which also failed
+  deterministically on clean `origin/main`.
 
 ## New tests
 
--- a/docs/releases/pending/1737-quarantine-debt-and-ci-portability-lints.md
+++ b/docs/releases/pending/1737-quarantine-debt-and-ci-portability-lints.md
@@ -1,0 +1,61 @@
+# Pay down knowledge-test quarantine debt and add CI cross-platform lints (issue #1737)
+
+## What changed
+
+- **Phase 1 — quarantine debt:** fixed five genuinely broken knowledge-subsystem
+  tests that had been quarantined rather than fixed (`cleanup-drift.test.ts` drift
+  comment, `knowledge.adversarial.test.ts` missing `unarchiveEntry` mock,
+  `knowledge-injector-drift-adversarial.test.ts` missing `confirmEntriesPhase`/
+  `recordLessonsShown` stubs, `knowledge-recall.test.ts` hive-tier path and a stale
+  ranking assertion). All five are un-quarantined from
+  `scripts/ci/quarantined-tests.txt`.
+- **`tests/integration/phase-complete-events.adversarial.test.ts`:** corrected the
+  "read-only `.swarm` directory" test to match the actual documented split-fail
+  behavior — `events.jsonl` writes soft-fail with a warning, `plan.json` writes
+  fail closed (`success: false`) with a specific lock-failure message. The prior
+  assertions expected uniform fail-open behavior across both files.
+- **`src/hooks/system-enhancer.ts`:** the unified-injection-budget ledger write is
+  now guaranteed via a `finally` block, so a mid-flight exception no longer skips
+  the budget-ledger write. Added
+  `tests/unit/hooks/system-enhancer-budget-ledger-finally.test.ts` as a regression
+  test.
+- **`tests/helpers/tmpdir.ts` (new):** `canonicalTmpDir()` /
+  `canonicalMkdtemp(prefix)` helpers that resolve the macOS `/var` → `/private/var`
+  symlink gap, plus `scripts/check-test-tmpdir.sh`, a diff-scoped lint that flags
+  raw `tmpdir()` usage introduced in new/changed test-file lines.
+- **`scripts/check-bash-portability.sh` (new):** a full-scan lint for bash-4+-only
+  constructs (`declare/typeset/local/readonly -A`, `grep -P`/`--perl-regexp`,
+  `coproc`, `mapfile`/`readarray`) that are silently unsupported on macOS's system
+  bash (3.2). Prevents recurrence of the class of bug fixed in issue #1729.
+  `scripts/check-cross-contamination.sh` had one genuine pre-existing PCRE
+  violation (`grep -oP`), fixed to a POSIX-ERE equivalent.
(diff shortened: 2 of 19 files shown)


zaxbysauce:
## 🤖 Multi-Stage PR Review
**Pipeline:** Esper3.1 (Qwen3.6-35B-A3B) (context pack) → Mellum2-Claude (bug/logic lens) + Qwythos-9B-v2 (sec/perf lens) (parallel explore, distinct lenses) → Qwythos-9B-v2 (critiquing Mellum findings) ↔ Mellum2-Claude (critiquing Qwythos findings) (cross-critique) → GLM-5-turbo (cloud, large PR fallback) (arbiter: blind-spot + synthesize)
**Commit reviewed:** `9915e4c04818`

---

### 🔍 PR Intent

> Reconstructed obligation list from PR text, issue #1737, commits, changed tests, and release note.

- **O-001** Fix five genuinely broken knowledge-subsystem tests and un-quarantine them from `scripts/ci/quarantined-tests.txt`.
- **O-002** Correct `phase-complete-events.adversarial.test.ts`'s read-only `.swarm` test to match actual split-fail behavior.
- **O-003** Fix the missing-`finally` budget-ledger write bug in `src/hooks/system-enhancer.ts` and add a regression test.
- **O-004** Add `tests/helpers/tmpdir.ts` (canonical tmpdir helpers) and `scripts/check-test-tmpdir.sh` (diff-scoped lint).
- **O-005** Add `scripts/check-bash-portability.sh` (full-scan lint for bash-4+-only constructs) and fix the caught PCRE violation in `check-cross-contamination.sh`.
- **O-006** Add `detect-paths` CI job to path-scope the cross-OS matrix, and add `unit-passed` aggregate job.
- **O-007** Correct stale comment about `coverage` check's required-status state.

---

### 📦 Implementation Summary

The PR fixes five knowledge-subsystem tests by adding missing mocks (`unarchiveEntry`, `confirmEntriesPhase`, `recordLessonsShown`), fixing a hive-tier path mismatch, and updating a stale ranking assertion. It corrects an adversarial integration test's expectations from uniform fail-open to documented split-fail behavior. It hoists `actualDemand`/`unifiedBudget` declarations above a `try` block and adds a `finally` to guarantee ledger writes. It adds two new lint scripts (bash-portability and tmpdir-canonicalization), a test helper, and corresponding CI steps. It adds a `detect-paths` job that expands the OS matrix for platform-sensitive PR diffs, plus an inert `unit-passed` aggregate. It fixes one genuine `grep -oP` → `grep -oE` PCRE violation and refines a security test to recognize GitHub's `env:`-block mitigation.

---

|---|---|---|
| O-001 | `SUPPORTED` | Five tests un-quarantined; mocks/path/assertion fixes in diff |
| O-002 | `SUPPORTED` | `tests/integration/phase-complete-events.adversarial.test.ts:330-420` — assertions now match split-fail docs |
| O-003 | `SUPPORTED` | `src/hooks/system-enhancer.ts:647-653,2498-2510` — hoisted decls + finally block; regression test added |
| O-004 | `SUPPORTED` | `tests/helpers/tmpdir.ts` (new), `scripts/check-test-tmpdir.sh` (new), CI step added |
| O-005 | `SUPPORTED` | `scripts/check-bash-portability.sh` (new); `check-cross-contamination.sh:36` PCRE fix |
| O-006 | `SUPPORTED` | `.github/workflows/ci.yml:65-146` detect-paths job; `390-416` unit-passed job |
| O-007 | `SUPPORTED` | `.github/workflows/ci.yml:390-395` — comment corrected |

---

### 🚨 Confirmed Findings

_None found._

---

_None found._

---

### 🧪 Test / Coverage Gaps

_None found._ The new `finally`-block behavior has a dedicated regression test (`system-enhancer-budget-ledger-finally.test.ts`). The new lint scripts are exercised by the PR's test-plan checklist. The tmpdir helper has 6 test cases.

---

### 📋 Shipped-vs-Claimed Gaps

_None found._

---

### 🔁 Validation Provenance

**DROPPED** (all 22 reviewer-confirmed findings):
- `check-test-tmpdir.sh:28` substring-match FP risk → script header explicitly documents this as accepted tradeoff ("fails safe"); by design.
- `check-test-tmpdir.sh:31` `|| true` swallows errors → base-branch resolution already handles missing-branch; `|| true` is standard `set -e` guard for git diff's normal no-match exit; matches sibling scripts.
- `quarantined-tests.txt:1` misleading 'unrelated' language → "unrelated to any specific PR" means infrastructure flakes, not caused by any PR; accurate.
- `quarantined-tests.txt:23` missing current-status note → minor doc gap, not a code defect.
- `system-enhancer.ts:652` CRITICAL uninitialized vars → finding's own evidence refutes: declarations are hoisted above try.
- `system-enhancer.ts:669` HIGH finally writes 0 → finding's own evidence refutes: `unifiedBudget` stays `undefined`, finally skips.
- `system-enhancer.ts:680` CRITICAL scope confusion → finding acknowledges "intended semantics"; readability preference, not a defect.
- `system-enhancer.ts:701` HIGH actualDemand state leak → factually wrong: `actualDemand` is a local variable in the hook callback, re-initialized to 0 on every invocation.
- `phase-complete-events.adversarial.test.ts:327` CRITICAL verify expectation → PR explicitly corrects a wrong test to match actual behavior; the fix IS the change.
- `phase-complete-events.adversarial.test.ts:387,398` HIGH verify EACCES / exact string → asks to verify pre-existing implementation, not a defect in this PR.
- `tmpdir.ts:32,39` LOW → self-refuting (helper does exactly what it's supposed to do) and acknowledges "no immediate issue."
- `ci-workflow-security.test.ts:66` MEDIUM inEnvBlock → description of intended change, not a defect.
- `ci-workflow-security.test.ts:66` CRITICAL stack overflow → YAML cannot have circular references; speculative.
- `ci-workflow-security.test.ts:88` HIGH nested env → `with.env` is not a valid GitHub Actions schema construct; speculative.
- `knowledge.adversarial.test.ts:255,269` HIGH verify mock/code path → test fixes with clear documentation; no defect.
- `system-enhancer.ts:45,89` two CRITICALs about finally ordering → first confuses `resetUnifiedBudget` location (it's IN the finally, not before it); second speculates about a simple Map/set throwing.
- `check-bash-portability.sh:12` HIGH → pure description of what the script does, not a defect.
- `tmpdir.ts:4` MEDIUM mock.module leak → wrong file reference; speculative "may be reused."

**BLIND-SPOT PASS:** No additional defects found. Verified: CI job dependency graph is correct (no skipped-job cascade on release branches), `detect-paths` grep pattern anchors correctly to repo-root paths, `check-test-tmpdir.sh` line-number tracking handles unified=0 diff correctly, `finally` block semantics are sound (`actualDemand` is per-invocation local, `unifiedBudget` undefined-gate prevents spurious writes), and the `knowledge-recall.test.ts` path fix correctly aligns with the mock.

---

### 📝 Merge Recommendation

**APPROVE**

All seven obligations are fully met. Every confirmed finding either self-refutes, describes intentional-by-design behavior, or is speculative without evidence. No CRITICAL, HIGH, or MEDIUM defects remain.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| No hardcoded secrets | ✅ |
| Lockfile consistent | ✅ |

---

zaxbysauce:
## 🔍 Swarm PR Review — `pr-1843-20260714`

**Verdict: 🔴 REQUEST CHANGES (CI RED — merge-blocked)**

4 CI checks are failing (`unit ubuntu/2`, `unit ubuntu/5`, `unit windows/2`, `unit-passed`). The root cause is **not** a production code regression — the production changes in this PR are all correct (verified below). The blocker is that the PR un-quarantines **2 test files without fixing them**, and both have **test bugs** (not production bugs) that the quarantine was hiding.

**Reviewed commit range:** `6d367a9f..9915e4c0` (16 files, +828/-61).

<details>
<summary><b>📊 Pipeline executed (default workflow, council=false)</b></summary>

- **Phase 0A/0B:** Drift bot ✅; mergeable=MERGEABLE, mergeStateStatus=BLOCKED; 0 inline reviews, 0 human comments.
- **Phase 3:** 6 base explorer lanes (correctness/security/deps/docs/tests/performance).
- **Phase 4:** 13-row trigger-eval (3 MATCHED: test-infrastructure, shell-write, git-safety); 3 micro-lanes dispatched.
- **Phase 6:** 2 independent reviewer lanes.
- **Phase 8:** 1 critic challenge (grouped ship-blocker).

The critic provided a valuable severity correction: the ship-blockers are **MEDIUM** (no production harm — these are test bugs), not HIGH. But RED CI blocks merge regardless of production severity.
</details>

---

## 🚫 Merge-blockers (CI RED — must fix before merge)

### 1. `integration-worktree.test.ts` un-quarantined without fix — FAILS ubuntu 2 + windows 2

- **What happened:** The PR removed `tests/unit/turbo/lean/integration-worktree.test.ts` from `scripts/ci/quarantined-tests.txt` (and deleted its diagnostic comment) without modifying the test file itself (zero diff on the `.test.ts` file).
- **CI result:** ubuntu shard 2 + windows shard 2 both FAIL on AC-6: `Expected: "conflict", Received: "partial"`.
- **Root cause (critic-identified, provable):** The test mock returns `{ partial: true, conflictFiles: [...] }` but the assertion expects `status === 'conflict'`. The production code at `runner.ts:1332` **correctly** returns `'partial'` when the result object has no `conflict: true` property. **This is a test bug, not a production bug.**
- **The old quarantine comment explicitly documented this:** *"Pre-existing failure: AC-6 conflict handling test fails on both Windows and Ubuntu CI. The mergeLaneBranch conflict detection returns 'partial' instead of 'conflict'."* — the PR deleted this comment at the exact moment the failure became visible again.
- **Fix options:** (a) Change the mock to return `{ conflict: true, ... }` instead of `{ partial: true, ... }`; (b) Change the assertion to expect `'partial'`; (c) Re-quarantine with the diagnostic comment restored.

### 2. `knowledge-floor-action.test.ts` un-quarantined without fix — FAILS ubuntu 5

- **What happened:** The PR removed `tests/unit/hooks/knowledge-floor-action.test.ts` from `scripts/ci/quarantined-tests.txt` without modifying the test file (zero diff).
- **CI result:** ubuntu shard 5 FAILS on G2 *"clears stale confidence_floor_demoted flag on recovery above floor"*: `expect(received).toBeGreaterThan(0.1)` received `0.1`.
- **Root cause:** Unknown — needs investigation. The production code (`knowledge-store.ts:1195-1198`) does `Math.max(CONFIDENCE_FLOOR, Math.min(CONFIDENCE_CEILING, entry.confidence + delta))` which should yield ~0.5 after a +0.4 boost, not 0.1. Possibly lockfile contention or read-after-write timing in test setup, or environment-dependent rollup state.
- **Fix options:** (a) Investigate and fix the root cause; (b) Re-quarantine with a tracking comment.

> **Note:** Both failures are **test bugs**, not production regressions. The critic correctly downgraded these from HIGH to MEDIUM on production-severity grounds. However, **RED CI blocks merge regardless** — the PR cannot be merged until these are resolved.

---

## ⚠️ Claim-integrity issues

### 3. `Closes #1737` is premature (MEDIUM)

Issue #1737's Definition of Done requires *"All 4 quarantine lists at 0 active entries."* After this PR:

| Quarantine list | Entries remaining |
|---|---|
| `quarantined-tests.txt` | 3 (swarm-artifact-cache, skill-scoring-workflow, test-runner) |
| `quarantined-tests-macos.txt` | 4 (integration-worktree + 3 lean-turbo) |
| `quarantined-integration-tests.txt` | 2 (phase-complete-events, unified-injection-budget) |
| `quarantined-tests-windows.txt` | 0 |

**Total: 9 entries across 3 lists, not 0.** The DoD is not met. Recommend changing `Closes #1737` → `Partially addresses #1737` to avoid auto-closing the issue on merge and losing tracking visibility for the remaining work (A1 macOS lean-turbo, A2b injection-budget).

### 4. Release note factual errors + disclosure gap (MEDIUM)

- The release note claims *"fixed five genuinely broken knowledge-subsystem tests"* but only **3** files (`knowledge.adversarial`, `knowledge-injector-drift-adversarial`, `knowledge-recall`) were actually modified. The "5" count comes from counting `knowledge-recall` twice (x2 issues), but the release note drops that qualifier.
- `knowledge-floor-action.test.ts` and `integration-worktree.test.ts` were silently un-quarantined (commit `686e1f35` claims *"already passed"*) but are **actively failing in CI**. They are NOT listed in the release note's "fixed" list, nor in the "Known caveats" section.
- The "Known caveats" section mentions *"remaining quarantined integration tests"* generically but does **not** disclose that 2 un-quarantined files are the direct cause of the 4 currently-failing CI checks.

**Fix:** Update the release note to (a) correct the "5 fixed" count, (b) explicitly name the 2 un-quarantined-but-failing tests, (c) disclose that they cause RED CI.

---

## ✅ Confirmed correct (the good work — 9 items independently verified sound)

<details>
<summary><b>Click to expand the 9 verified-correct changes</b></summary>

| Change | Verdict | Evidence |
|---|---|---|
| **FR-004 finally block** (`system-enhancer.ts:2501-2512`) | ✅ Correct | Guarantees ledger write on exception path (the ONLY writer when try-block throws before Path A/B finalize). Double-write on normal path is harmless idempotency (identical `Map.set` values, no observers). |
| **SIGPIPE fix** (`ci.yml:139` `grep -qE` → `grep -E >/dev/null`) | ✅ Correct | Under `set -o pipefail`, `grep -q` exits on first match → closes stdin → SIGPIPE to `git diff` → pipefail reports 141 → `if` silently false even on match (the unsafe direction: drops platform-sensitive PRs to ubuntu-only). `grep -E` reads to EOF → no SIGPIPE. Verified no new problem introduced. |
| **PCRE-to-ERE fix** (`check-cross-contamination.sh:36`) | ✅ Correct | `grep -oP` → `grep -oE`, `\d+` → `[0-9]+`, `\s*` → `[[:space:]]*`. Semantically identical for this use case; fixes the macOS BSD-grep incompatibility. |
| **detect-paths + unit-passed CI design** | ✅ Correct | Step-level `if:` (not job-level) produces job result `"success"` not `"skipped"` when all steps are skipped — so `unit-passed` correctly passes on release-please branches. `merge_group` events always run the full 3-OS matrix. Fork PR SHAs are immutable OIDs. |
| **SHA-pinning** | ✅ All 7 actions pinned | `checkout@93cb6efe... #v5.0.1`, `setup-bun@0c5077e5... #v2.2.0`, `cache@27d5ce7f... #v5.0.5`, etc. No unpinned third-party actions. |
| **env-block expression-injection mitigation** | ✅ Correct | `github.event.pull_request.base.sha`/`head.sha` are immutable 40-char hex git OIDs — no shell metacharacters possible. The `ci-workflow-security.test.ts` refinement correctly recognizes GitHub's recommended intermediate-env mitigation. |
| **New lint scripts** (`check-bash-portability.sh`, `check-test-tmpdir.sh`) | ✅ Safe | No `eval`, no backtick substitution, all variables double-quoted, `IFS= read -r` for filenames with spaces, self-exclusion prevents self-triggering. |
| **Mock hygiene** | ✅ Clean | `system-enhancer-budget-ledger-finally.test.ts` spreads `...realExtractors`; knowledge test mocks are project modules (not `node:*`, so spread mandate doesn't apply); `_internals` DI seams used correctly. `check-mock-cleanup.sh` passes. |
| **FR-004 regression test strength** | ✅ Strong, not tautological | Marker-gated exception injection at a precise code point (after phase-header `tryInject`, before plan-cursor). Sanity checks prove the throw fired at the right location. Three independent assertions: `recordedDemand > 0`, `getUnifiedBudgetTotal === unifiedBudget`, ledger entry exists. |

</details>

---

## 🔵 Advisory findings (LOW, non-blocking)

| Finding | Severity | Detail |
|---|---|---|
| **New test self-inconsistency** | LOW | `system-enhancer-budget-ledger-finally.test.ts:70-76` uses `fs.mkdtempSync(path.join('tmp', ...))` (relative CWD `tmp/`) instead of the `canonicalMkdtemp` helper this PR introduces. The `check-test-tmpdir.sh` lint won't catch it (deliberately scoped to `tmpdir()` substring), but it violates the PR's own stated convention in `tests/helpers/tmpdir.ts:14-16`. |
| **`src/hooks/` not in detect-paths filter** | LOW | Despite 90+ `process.platform`/`path.win32` refs in `src/hooks/`, the directory isn't in the platform-sensitive path list. Low risk — hooks delegate platform-specific work to filtered modules (`src/worktree/`, `src/sandbox/`), so a hook cross-platform regression would typically co-occur with a change to a filtered module. |
| **phase-complete-events Windows skip** | INFO | `if (process.platform === 'win32') return` — justified (Windows `chmod` semantics differ fundamentally from POSIX; the read-only-`.swarm` test is POSIX-specific). The split-fail behavior (events.jsonl soft-fail vs plan.json fail-closed) is only verified on Linux. Acceptable tradeoff. |

---

## 📋 Obligation status (issue #1737)

| Obligation | Status |
|---|---|
| Fix 5 quarantined knowledge tests | ⚠️ PARTIALLY MET — 3 fixed, 1 assumed-fixed-by-prior-PRs (cleanup-drift), 1 un-quarantined-but-failing (knowledge-floor-action) |
| Fix phase-complete-events read-only .swarm coverage | ✅ MET |
| Fix system-enhancer budget-ledger finally (FR-004) | ✅ MET |
| Add tmpdir helper + check-test-tmpdir.sh lint | ✅ MET |
| Add check-bash-portability.sh + fix PCRE violation | ✅ MET |
| Add detect-paths + unit-passed + SIGPIPE fix + coverage comment | ✅ MET |
| Refine ci-workflow-security env: block mitigation | ✅ MET |
| **Closes #1737** (DoD: all 4 lists at 0) | ❌ NOT MET — 9 entries remain; 2 un-quarantined files failing |

---

## Recommended fix path

1. **Re-quarantine or fix** `knowledge-floor-action.test.ts` and `integration-worktree.test.ts`:
   - `integration-worktree` is likely a **one-line mock fix**: change `{ partial: true, ... }` → `{ conflict: true, ... }`, or change the assertion to expect `'partial'`
   - `knowledge-floor-action` needs root-cause investigation (confidence value 0.1 vs expected ~0.5)
   - If deferring, re-quarantine with the diagnostic comments restored
2. **Change `Closes #1737`** → `Partially addresses #1737`
3. **Update release note**: correct the "5 fixed" count; explicitly name the 2 un-quarantined-but-failing tests; disclose the RED CI cause

---

<details>
<summary><b>📁 Review artifacts (<code>.swarm/pr-review/pr-1843-20260714/</code>)</b></summary>

- `findings.jsonl` — full candidate/reviewer/critic ledger
- `trigger-eval.json` — 13-row trigger-eval ledger (3 MATCHED)
- `context-pack.json` — scope, obligations, deterministic signals, CI failure detail
- `pr-diff.patch` — full reviewed diff

Review run: `pr-1843-20260714` · 6 base lanes + 3 micro-lanes + 2 reviewer lanes + 1 critic lane.

</details>

---

*Automated review by the swarm-pr-review protocol. The production code changes in this PR are all correct and well-tested — the merge-block is purely from un-quarantining 2 test files that have test-level bugs (not production bugs). The SIGPIPE fix, FR-004 finally block, CI design, and new lints are all sound contributions.*

zaxbysauce:
Thanks for the review — this ran against `9915e4c0`, which is now superseded by `acc6ca5`. Status update on the two merge-blockers:

**Both are fixed, and both were genuine production bugs, not test bugs.**

### `integration-worktree.test.ts` AC-6 (conflict → 'partial')
The suggested fix (change the mock to return `conflict: true`) would have masked the real defect. I traced `attemptMergeBackFromDirty`'s `DirtyMergePartial` result (`src/worktree/merge.ts:477-486`) — it **never** sets a `conflict` field on an actual merge conflict, only `partial: true` + non-empty `conflictFiles`. Only the unrelated `mergeLaneBranch`'s `MergeConflict` type sets `conflict: true`. So `_sequentialWorktreeCleanup`'s `mergeResult.conflict === true` check was dead code for this call path — every real dirty-worktree conflict was being reported as `'partial'` in production, not just in the test. Fixed the classifier in `runner.ts` to key off non-empty `conflictFiles` instead. Verified the test mock accurately reflects what `attemptMergeBackFromDirty` actually returns.

### `knowledge-floor-action.test.ts` (confidence stuck at 0.1)
Root-caused via instrumentation (not "unknown"): `swarm-artifact-cache`'s stat-based staleness check (mtime+ctime+size) never invalidated on write. `bumpKnowledgeConfidenceBatch` writes the confidence delta, then immediately does a second locked read-modify-write (the confidence-floor-flag transaction) — on a same-size rewrite landing within one filesystem timestamp tick, the second read got a cache hit on the pre-write content and wrote it back, clobbering the delta. This is a real lost-update bug reachable any time two locked writes to the same knowledge file happen in quick succession, not just in this test. Fixed by invalidating the cache entry from the shared `atomicWriteFile` primitive after every write (closes the gap for all ~15 callers, not just this path).

Both reproduced deterministically (10/10 runs) on clean `origin/main` before the fix, and now pass deterministically after it. Full CI is green at `acc6ca5` (all `unit` shards + `unit-passed`, `mergeable_state: clean`). Release note fragment updated to document both fixes.

Re: item 3 (`Closes #1737` vs `Partially addresses #1737` given 9 remaining quarantine entries) — that's a fair catch on the DoD gap; flagging it to the PR author to decide rather than changing the PR's stated scope unilaterally.

---

Issue #1875, opened 2026-07-17
🐛 [BUG REPORT]: Agent Infinite Loop on Non-Transient Errors


AbuYusof:
# 🐛 BUG REPORT: Agent Infinite Loop on Non-Transient Errors

**Filed by:** NORM Decision OS team  
**Date:** 2026-07-17  
**Affects:** Windows agents (all roles)  
**Priority:** HIGH (wasted tokens, time, work stoppage)  
**Filed against:** opencode-swarm (plugin guardrails)

---

## Summary

Agents enter **infinite retry loops** when hitting non-transient errors. They retry the same or similar command indefinitely, consuming 5000–15000 tokens per attempt, because:

1. **No error classification** — agents cannot distinguish transient errors from permanent blocks
2. **No `ParseError` coverage** — the STOP protocol covers L0–L5 errors but not command-generation failures  
3. **Sanitizer wrapper is brittle** — `#` comments + PowerShell `-Command` parameter = silent corruption  
4. **PATH restriction is over-aggressive** — `$env:PATH = 'C:\WINDOWS\System32;C:\WINDOWS'` removes `python`, `node`, `git`

---

### Evidence  

Every bash command on Windows is wrapped in this sanitizer script:  

```powershell
powershell -ExecutionPolicy Bypass -NoProfile -WindowStyle Hidden -Command "
$ErrorActionPreference = 'Stop'; try {
   # Set scoped temp directory            ← END-OF-LINE COMMENT
   $env:TEMP = '...';
   # Note: LD_PRELOAD...                  ← ANOTHER COMMENT
   $dangerousVars = @('LD_PRELOAD', ...);
   ...
   cmd /c \"python script.py\";
} catch { ... }"
```

**Result:**  

```
Missing closing '}' in statement block or type definition.
    + CategoryInfo          : ParserError
    + FullyQualifiedErrorId : MissingEndCurlyBrace
```

### Root Cause  

`#` in PowerShell is an **end-of-line comment**. When `-Command` normalizes line breaks, `#` comments **swallow everything after them** — including the closing `}` and `catch` block.  

### Impact  

- Agent receives `ParseError` — not a guardrail block, not a transient failure  
- Agent does **not** recognize `MissingEndCurlyBrace` as a STOP signal  
- Agent retries → same error → **loop forever**  
- Typical cost: **8000–15000+ tokens** per loop cycle  

### Suggested Fix (Sanitizer)  

**Option A — Remove `#` comments:**  
[...]

PR #1901, opened 2026-07-19
fix(delegation): encoding-robust ACCEPTANCE checks, model quota failover, scope-activation + interpreter-eval shell-write fixes

### LINKED ISSUE #1896: ACCEPTANCE-field parsing failure + model failover not triggering on quota exhaustion
Environment: 

Project: F1 race engineer, local swarm run
Swarm config: 20-agent OpenCode swarm, role-based model routing (Kimi for coding, OpenCode, OpenRouter)
Spec authored separately in Claude.ai web (Fable 5) and dropped into .swarm/spec.md
Workarounds tried: session reset, cleared .opencode cache — neither resolved the issue
Windows 11 - VS Code Plugin 

A single architect session trying to dispatch Task 1.1 burned roughly 20+ turns and multiple full context compactions on what turned out to be three separable bugs: (1) the ACCEPTANCE-field verbatim check is not encoding-robust and gives no diagnostic signal about why a match failed, (2) the configured model for the coder/critic role silently diverged from what the UI selector actually had active, and when that model hit its quota, the swarm did not fail over to a configured fallback — it just failed the stage. A later model swap (to Nemotron 3 Ultra, via Hy3) fixed the dispatch-format problem but not a separate circuit-breaker fault that's still failing sub-agent sandboxes.

What happened (timeline)

Spec was planned/authored using Claude Fable 5 in claude.ai web and saved to .swarm/spec.md. This introduced non-ASCII section-symbol characters (§) into the FR citations.
Swarm was interrupted following Issue #1886. On resume, the model selector shown in the UI was Nvidia Nemotron 3 Ultra, while .opencode/swarm JSON config specified Kimi K3 for that role. These were out of sync — the UI and the effective config did not agree on which model was running.
User manually switched the model mid-session after noticing the mismatch, but by that point planning had already progressed to task 1.2, so the switch landed mid-plan rather than at a clean boundary.
Kimi K3 hit its usage/rate limit on Kimi Code. The critic stage failed as a result, and scope files (declare_scope output) were generated incompletely/incorrectly off the back of that failure.
No failover occurred. Despite quota exhaustion being a distinguishable, retryable failure mode, the swarm did not switch to any configured fallback model — it just let the critic/scope stage fail and left the architect to work with bad state.
After the interrupt and a swap to Nemotron 3, the architect got stuck in a long loop re-dispatching the same coder task with different ACCEPTANCE-field encodings (§ → ?? → sec), because the verbatim-match checker gave a coverage-mismatch error with no indication that the actual problem was a byte-encoding mismatch between what the architect sent and what was literally on disk in spec.md (which itself had mojibake ?? in place of § from the original save).
This was eventually resolved manually (by "Hy3") — but only for the task-level dispatch loop. Sub-agents (coder instances) are still hitting a circuit breaker (sandbox_wrapper_failure) on their read/write/bash tools, independent of the ACCEPTANCE-field issue.

Issues to file

1. ACCEPTANCE-field verbatim check has no encoding tolerance and no useful diagnostic

The checker requires byte-for-byte match between the ACCEPTANCE text a delegating agent sends and the FR text stored in spec.md, but:

It doesn't normalize Unicode (e.g. § U+00A7 vs. a mojibake ?? placeholder vs. an ASCII substitute) before comparing.
The error (ACCEPTANCE_FIELD_COVERAGE_MISMATCH) doesn't say what didn't match or show a diff — the architect had to manually hex-dump lines from spec.md to figure out the file was storing § as 3F 3F on disk.
This turned a one-line encoding fix into a multi-hour, multi-compaction debugging session, most of which was the architect guessing at which Unicode variant would round-trip through the delegation layer.

Ask: normalize/validate encoding on both sides before the byte-compare, and have the mismatch error return the actual diff (or at least byte offsets) instead of a bare pass/fail.

2. UI model selector and effective swarm config can silently diverge

After the interrupt from Issue #1886, the UI showed Nemotron 3 Ultra as active while the swarm's own JSON config still specified Kimi K3 for that role. Nothing surfaced this mismatch — it was only caught by the user noticing model behavior looked off, well after planning had already moved on to task 1.2.

Ask: on resume from an interrupt, validate that the UI-selected model matches the persisted per-role config, and surface a warning (or block) if they disagree, rather than silently running whichever one the UI happens to have loaded.

3. No failover when a role's model hits a quota/rate limit

Kimi K3 hit its usage limit on Kimi Code mid-run. Expected behavior for a swarm with role-based model routing is that this is a recognizable, retryable condition and the role fails over to a configured backup model. Instead the critic stage just failed, and that failure cascaded into malformed scope files for downstream tasks.

Ask: treat quota/rate-limit errors as a distinct error class from content/logic failures, and route them through whatever failover chain is configured per role instead of failing the stage outright.

4. Sub-agent circuit breaker (sandbox_wrapper_failure) persists independent of the above

Even after the dispatch-format issue was fixed, coder sub-agents are hitting a non-transient circuit breaker on their read/write/bash tools. This looks like a sandbox provisioning fault rather than anything content-related — the architect's own tools worked fine in the same session, only the delegated sub-agent sandbox was broken.

Tried and did not help: resetting the session, clearing the .opencode cache.

Ask: needs sandbox-provisioning-side investigation; from the client side there's no way to distinguish "sub-agent sandbox is dead" from "sub-agent is refusing to act" other than the repeated identical error string, and there's currently no automatic recovery path (session/cache reset didn't clear it).

Suggested repro:

Author a spec containing § citations via Claude.ai web OR Kimi.com web, save to .swarm/spec.md.
Start a swarm run, interrupt it, resume, and check whether the UI's active model matches the per-role JSON config.
Let the coding-role model exhaust its quota mid-task and observe whether any fallback model is invoked.
Dispatch a coder sub-agent task and check whether its sandbox tools are reachable.

### LINKED ISSUE #1902: Shell-write guard hard-blocks interpreter-eval commands (python -m/-c, node -e) with misleading "unresolvable path target"
## Summary

The shell-write scope guard in `src/hooks/guardrails/tool-before.ts` hard-blocks any bash/shell command that the write-detector flags as an `interpreter_eval` write — `python -m …`, `python -c …`, `node -e …`, `bun -e`, `ruby -e`, `perl -e`, `php -r`, etc. — with:

> `BLOCKED: bash/shell write operation with unresolvable path target — rejecting for safety`

This fires **before** the scope/authority checks and **before** the guardrails-disable switch, so it can't be configured around. It affects **all platforms**, but is most visible on Python projects where a coder routinely runs `python -m pytest … 2>&1`. Reported by @DarkJoney in #1896 (Windows 11 Python project); their run is fully stalled because every `python -m …` the coder issues is hard-blocked.

Split out of #1896 per the `needs-split` triage. **Distinct from the five defects fixed in #1901** — none of those touch this code path.

## Root cause (source-verified)

Two layers combine:

1. **Detector — working as designed.** `detectInterpreterEval()` (`src/hooks/shell-write-detect.ts:657`) flags interpreter commands carrying an eval flag (`-c`/`-m` for python, `-e`/`-p` for node/bun, `-e` for ruby/perl, `-r` for php) as a write with **`path: null`** — a *capability* marker meaning "this could write somewhere we can't statically determine." Intentional (`python -m compileall .` really can write) and explicitly tested (`src/hooks/shell-write-detect.test.ts:307`).

2. **Consumer — the bug.** `resolveWriteTargets()` correctly maps `path: null` → `resolvedPath: null`. But `checkShellWriteScope()` then does:

```ts
for (const write of resolvedWrites) {
  if (write.resolvedPath === null) {
    throw new Error(
      `BLOCKED: bash/shell write operation with unresolvable path target — rejecting for safety`,
    );
  }
  // …authority + declared-scope checks below are never reached…
}
```
(`src/hooks/guardrails/tool-before.ts:982`)

This conflates **two different `resolvedPath === null` cases**:

| `original.path` | Meaning | Correct treatment |
|---|---|---|
| `null` | *Capability* write (interpreter_eval, here-doc marker). There was never a concrete target. | Route through authority/scope — **not** "unresolvable path". |
| non-`null` but resolves to `null` | Genuine dynamic redirect target: `echo x > $OUTFILE`, `> $(cmd)`. | Keep the hard-block — this is what it was built for (`shell-write-detect.test.ts:1150` "marks $VAR path as unresolvable"). |

Because both collapse to `resolvedPath === null`, every interpreter-eval invocation gets the message meant for the second case.

## Impact / observed symptoms (#1896)

- `python --version` works (no eval flag) but `python -m pytest … 2>&1`, `python -m ruff --version`, `python -c "…"` are hard-blocked — exactly @DarkJoney's pattern.
- The "unresolvable path target" text is meaningless for these commands and sent the reporter down wrong paths (suspected antivirus, then `2>&1` redirect parsing — note `2>&1` is **not** the trigger; `>&` / `GREATAND` is deliberately excluded from `REDIRECT_WRITE_TOKENS`).
- Fires ahead of scope/authority and ignores `guardrails.enabled:false`, so there's no config workaround.
- Downstream cascade: the coder retries the same blocked write, trips the pattern-escalation guard (`PRM HARD STOP: Pattern escalation maximum reached`), and reports the sub-agent as broken.

## Proposed fix (surgical)

In `checkShellWriteScope()`, discriminate the two null cases before the blanket throw:

- **`original.path !== null && resolvedPath === null`** → keep the existing hard-block (genuine dynamic redirect target).
- **`original.path === null`** (capability write) → do **not** emit "unresolvable path target". Route through the normal authority/declared-scope model using `effectiveDirectory` (cwd) as the effective target, so:
  - architect / full-authority agents pass;
  - an out-of-scope agent is blocked with a **meaningful** scope error;
  - universal deny prefixes still apply;
  - the OS sandbox wrapper (`applySandboxExecution`) still confines any actual writes the interpreter performs.

Optionally add a `guardrails.block_interpreter_eval` policy flag (default off) for projects that want to forbid `-c`/`-m`/`-e` outright — opt-in policy rather than an un-configurable default block.

### Related (secondary, lower priority)
On Windows the `bash` tool forces `shellType='posix'` (`tool-before.ts:893`), so genuine PowerShell writes (`Set-Content`, `Out-File`, `$env:` paths) are analyzed by the POSIX parser. Worth a follow-up, but the interpreter-eval block above is the primary unblock for #1896.

## Repro
1. As a coder with a declared scope, run a bash command with an eval flag, e.g. `python -m pytest tests/ 2>&1` or `node -e "console.log(1)"`.
2. Observe `BLOCKED: bash/shell write operation with unresolvable path target — rejecting for safety`.
3. Confirm `python --version` (no eval flag) is allowed, isolating the eval-flag trigger.

## Acceptance
- interpreter-eval commands are no longer hard-blocked with "unresolvable path target"; they follow the agent's authority/scope with a meaningful message when denied.
- genuine dynamic-redirect targets (`> $VAR`, `> $(cmd)`) remain hard-blocked (regression test kept).
- unit coverage for both null-path branches of `checkShellWriteScope`.

zaxbysauce:
Closes #1896
Closes #1902

## Summary

Resolves issue #1896, which bundled four separable defects, plus two related bugs surfaced during the investigation and fixed on the same branch at the maintainer's request: `SCOPE_NOT_DECLARED` scope-binding activation, and an interpreter-`-m` shell-write false positive (#1902).

- **ACCEPTANCE-field encoding** (`src/hooks/delegation-gate.ts`): the coder/reviewer dispatch coverage check had no Unicode/mojibake tolerance and returned a bare, misleading error. `normalizeAcceptanceText` now canonicalizes Unicode (NFC) and folds the section sign `§`/its Latin-1 mojibake to "section" (symmetric, bypass-safe); a mismatch now returns a bounded diagnostic (first-divergence offset, expected/found snippets, an encoding-corruption hint) instead of a false "copy byte-for-byte" instruction.
- **Model quota failover** (`src/utils/provider-error-classification.ts`, `src/utils/model-dispatch-fallback.ts`): quota/rate-limit errors were misclassified as permanent because the transient classifier was duplicated three times, none recognizing quota; the fallback-model switch only existed in the guardrails tool-result hook, so the critic, reviewer, and coder-lane dispatch stages never failed over. Single-sourced the classifier (quota kept as a *separate* class, combined only in dispatch paths — never the tool-output path, so a shell "Disk quota exceeded" can't trigger a bogus fallback) and wired a shared retry+fallback helper into oversight, reviewer, and runner. The evaluation dispatcher classifies quota and retries the *same* model (no substitution — the model id is the benchmark subject).
- **Sandbox circuit-breaker diagnostics** (`src/hooks/guardrails/nontransient-circuit.ts`): the non-transient hard-stop message dropped the stored failure signal, so an operator couldn't tell a dead sandbox from a sub-agent refusing to act. The message now surfaces the actual signal plus category-specific remediation (points to `/swarm diagnose`, explains the in-memory/invocation-keyed reset semantics). The immediate hard-stop and no-self-clear design (invariant 9 / issue #1875) is unchanged.
- **Model-divergence detection** (`src/state.ts`, `src/session/snapshot-{writer,reader}.ts`, `src/hooks/guardrails/messages-transform.ts`): the architect's active model is UI-driven by design, so nothing surfaced a silent switch across an interrupt. The observed model is now persisted (deliberately excluded from the transient-reset field set, so it survives rehydration) and the architect session gets a one-shot advisory on a genuine cross-interrupt change, gated on no swarm-initiated fallback being in play.
- **`SCOPE_NOT_DECLARED` scope-binding activation** (`src/index.ts`): the event handler that activates a coder's scope binding required `metadata.parentSessionId`, a key 1.1.x-era OpenCode runtimes never emit (verified against upstream `sst/opencode` `task.ts` at the pinned SDK version) — activation was unsatisfiable and every default-mode coder write failed `SCOPE_NOT_DECLARED`, immune to session resets. Parent identity now comes exclusively from the SDK-required `part.sessionID` (no metadata fallback; empty ids fail closed) — this is also a security hardening, since the old code would activate from attacker-controllable metadata.
- **Interpreter-`-m` shell-write false positive** (`src/hooks/shell-write-detect.ts`, `src/hooks/guardrails/tool-before.ts`, #1902): the write-scope guard flagged every interpreter eval flag — *including `-m`* — as a null-path write, then hard-blocked any null resolved path with a misleading `BLOCKED: … unresolvable path target` error, *before* scope/authority checks and regardless of `guardrails` config. This hard-blocked every `python -m pytest` / `python -m ruff` (even the swarm's own `test_runner`-generated commands) on all platforms — the wall the #1896 reporter hit. `python -m <module>` runs an installed module as a program (equivalent to invoking `pytest`/`ruff` directly), so it is no longer treated as an unverifiable write; inline `-c`/`-e`/`-r` eval stays flagged and unconditionally fail-closed, so the eval-bypass guard for scope/config zones is unchanged. The null-path block is now discriminated by category: a bare here-doc marker (stdin, not a write) is skipped, and interpreter-eval and dynamic `$VAR`/`$(…)` redirect targets keep failing closed with accurate, actionable messages. Cleared by an independent adversarial reviewer *and* a separate critic (both could not break the eval-bypass property).

Every defect went through localization → fix plan → independent plan critic → implementation → independent implementation review → independent final critic, with re-review after every post-approval change. Full trace: `.agents/issue-traces/1896-acceptance-field-model-failover/`.

## Invariant audit
- 1 (plugin init): not touched — the `src/index.ts` change is inside the `event` callback registered on the returned plugin object (line ~1352), not on the awaited `initializeOpenCodeSwarm` critical path; no new synchronous/awaited init work added.
- 2 (runtime portability): touched — two new files (`src/utils/provider-error-classification.ts`, `src/utils/model-dispatch-fallback.ts`) use only Node-safe APIs (`setTimeout`); grep confirms no `bun:` import or `Bun.*` call in either. The #1902 change adds no APIs. `bun run build` succeeds; `node --input-type=module -e "await import('./dist/index.js')"` succeeds.
- 3 (subprocesses): not touched.
- 4 (.swarm containment): not touched — no new `.swarm/` path creation; snapshot changes are additive fields on the existing `session/state.json` shape.
- 6 (test_runner safety): not touched — all validation used shell `bun test`/`bun run test:unit:ci`, never the `test_runner` tool with a broad scope.
- 7 (test writing): touched — all new/modified tests use `bun:test` only. `bash scripts/check-test-file-cap.sh` → 0 new-file violations, 0 ratchet violations (incl. the #1902 additions to `tests/unit/hooks/guardrails-shell-write.test.ts`).
- 8 (session state): touched — `lastObservedModel`/`lastObservedProviderID` are two bounded strings keyed on the existing per-`sessionID` `AgentSessionState`, excluded from `TRANSIENT_SESSION_FIELDS`. The #1902 change adds no session state.
- 9 (guardrails/retry): touched — quota is a class distinct from generic transient status codes, combined with the narrow pattern only inside the *dispatch* classifier (never the tool-output classifier). The non-transient circuit's immediate hard-stop / no-self-clear (issue #1875) is unchanged. Scope-binding activation is strictly hardened (tool-controllable metadata excluded from parent-identity). The **#1902 interpreter-`-m` change is a guard-scoping fix**: `-m` module runs are reclassified as non-eval (equivalent to the already-allowed `python script.py`, which this guard never covered), while inline `-c`/`-e`/`-r` eval stays *unconditionally* fail-closed — the config-zone/scope eval-bypass regression tests are unchanged, and the OS sandbox wrapper remains the runtime containment. `checkDestructiveCommand` (which runs before the write-scope guard) never consulted the eval category, so nothing destructive is re-enabled.
- 10 (chat/system msg): touched — all new advisories use the existing `pendingAdvisoryMessages` queue and collapsed system message; verified end-to-end in `tests/unit/hooks/model-divergence.test.ts`.
- 11 (tool registration): not touched — no new tool, agent, command, or status. `bun run scripts/check-tool-registration.ts` → "108 tools, coherent."
- 12 (release/cache): touched — three fragments added (`docs/releases/pending/1896-acceptance-encoding-model-failover.md`, `docs/releases/pending/scope-binding-activation-event-shape.md`, `docs/releases/pending/1902-interpreter-module-run-shell-write.md`), every concrete reference ground-truth-verified against source. No hand-edit of `package.json#version`, `CHANGELOG.md`, or `.release-please-manifest.json`.

## Test plan
- [x] `bun run build` — succeeds
- [x] `node --input-type=module -e "await import('./dist/index.js')"` — succeeds
- [x] `bun run typecheck` — 0 errors
- [x] `bun run lint:ci` (pinned Biome) — clean (2643 files)
- [x] `bun run scripts/check-tool-registration.ts` — 108 tools coherent
- [x] `bash scripts/check-invariants.sh` — all passed
- [x] `bash scripts/check-test-file-cap.sh` — 0 new-file violations, 0 ratchet violations
- [x] `bun test src/hooks/shell-write-detect.test.ts tests/unit/hooks/guardrails-shell-write.test.ts` — 326 pass / 0 fail (#1902 interpreter-`-m` fix + eval-bypass / dynamic-path / here-doc regressions)
- [x] `bun test tests/integration ./test --timeout 120000` — 741 pass / 213 fail, **byte-for-byte identical pass/fail counts reproduced on a clean `origin/main` worktree** (pre-existing; see below)
- [x] `bun test tests/security --timeout 120000` — 167 pass / 0 fail
- [x] `bun test tests/adversarial --timeout 120000` — 829 pass / 0 fail
- [x] `bun test tests/smoke --timeout 120000` — 10 pass / 0 fail

### Pre-existing failures
Tier 3 integration (`tests/integration ./test`) shows 213 failures out of 954 tests when run as a full folder (741 pass). These are **pre-existing and unrelated to this PR** — verified by checking out `origin/main` into a clean worktree, running `bun install --frozen-lockfile && bun run build`, and re-running the exact same command: **741 pass / 213 fail**, identical counts. Most of the remainder is Bun's documented cross-test-file `mock.module` leakage (AGENTS.md #7) reproducing identically on both branches. None of the failing files or their dependencies overlap with this PR's diff.

### Independent review trail (per file, in `.agents/issue-traces/1896-acceptance-field-model-failover/`)
- `06-critic-review.md` — plan critic (NEEDS_REVISION → all blockers resolved)
- `08b-implementation-review.md` — implementation reviewer (APPROVE, adversarial)
- `09-final-critic.md` — final critic (APPROVE, re-run after every post-approval fix)
- `11-1902-interpreter-module-run-fix.md` — #1902 root cause, fix, reviewer APPROVE + critic UPHOLD_APPROVE
- `08-test-results.md`, `08a-recurrence-sweep.md` — full test evidence and recurrence-class guardrails per defect

---
_Generated by [Claude Code](https://claude.ai/code/session_01EQEAvf4J716ToK2QPXT1wp)_


zaxbysauce:
## Review feedback addressed (swarm-pr-feedback)

Followed up on the validated findings from the independent `swarm-pr-review` run. Commit `5eed2284`.

### FIXED in this commit

- **PRR-001 (HIGH — oversight failover untested):** Added `tests/unit/full-auto/oversight-fallback.test.ts` (5 tests) covering the bespoke inline failover loop — quota→failover recovery, `baseEvent.critic_model` attribution rewrite, `telemetry.modelFallback` signal, the increment-before-parse malformed-entry skip, `critic_oversight`→`critic` role inheritance, and the permanent-error fail-closed path.
- **PRR-003 (MED — evaluation same-model quota retry untested):** Added 4 tests to `tests/unit/evaluation/model-dispatcher.test.ts` — retry-NOT-substitute (benchmark attribution), session cleanup between retries (no leak), the enriched error string, and the permanent-error no-retry path.
- **PRR-006 (LOW — reviewer classify asymmetry):** Added the `Reviewer dispatch timed out` → permanent carve-out in `src/turbo/lean/reviewer.ts`, mirroring the lane runner. **Note:** the final critic caught that this is defense-in-depth/alignment, not a behavior fix — today's `"timed out"` message doesn't match the transient classifier (`timeout` atom) anyway, so timeouts already failed closed. The carve-out guards a future message rewording; pinned by a new test.
- **PRR-025 (LOW — release fragment overstatement):** Corrected `1896-acceptance-encoding-model-failover.md` — only the runtime error string dropped "byte-for-byte"; the architect/coder prompt templates still say it (the right intent — the gate just no longer false-blocks on encoding variants).

### TRACKED separately

- **PRR-004 (MED — per-role failover coverage):** Issue #1905 tracks the remaining dispatch sites that still call `client.session.prompt` directly with no failover (`full-auto-intercept.ts:577`, `auto-review.ts:214`, `lean/integration.ts:469/487`, plus opt-in curator/skill-improver/mutation). The PR's explicit scoping to oversight/reviewer/runner is reasonable; this ensures merging `Closes #1896` doesn't lose the residual per-role obligation.

### Gates

- Independent reviewer (Kimi K2.7): **APPROVE** all items.
- Final critic (Kimi K3): **NEEDS_REVISION** on the new release fragment (it initially overstated the reviewer-timeout change the same way PRR-025 did) → reworded the fragment to "no behavior change / defense-in-depth" + added a pinning test → **APPROVE**.
- 21/21 affected tests pass; `bun run typecheck` clean; `bunx biome ci` clean; `scripts/check-test-file-cap.sh` clean (FR-006).
- Broader affected-area sweep (full-auto + turbo/lean + evaluation, 60 files): 1023/1023 pass.

Full closure ledger: `.zcode/pr-review/1901/feedback-handoff.md` (local). LOWs PRR-007/008/010/011/018/023/024 are sanctioned-deferred (non-merge-blocking); recommend a docs follow-up for PRR-023/024 (stale `architecture.md`/`configuration.md` fallback descriptions).

zaxbysauce:
## 🤖 Multi-Stage PR Review
**Pipeline:** Esper3.1 (Qwen3.6-35B-A3B) (context pack) → Mellum2-Claude (bug/logic lens) (explore → candidates) → GLM-5-turbo (cloud, large PR fallback) (challenge + blind-spot)
**Commit reviewed:** `5eed22848b48`

---

### 🔍 PR Intent

> Reconstructed obligation list from PR text, issue #1896, issue #1902, commit messages, changed tests, and changed interfaces.

- **O-001** ACCEPTANCE-field coverage check must canonicalize Unicode (NFC) and fold `§`/`Â§` to "section" so encoding variants of a faithful copy don't false-block.
- **O-002** Coverage-miss error must return a bounded diagnostic (divergence offset, expected/found snippets, encoding-corruption hint) instead of a bare "copy byte-for-byte" instruction.
- **O-003** Transient provider-error classifier must be single-sourced (eliminate 3-way duplication across guardrails/index, messages-transform, oversight).
- **O-004** Quota/usage-limit exhaustion must be a distinct, retryable+failover-eligible error class — but kept OUT of the tool-output classifier path so shell `Disk quota exceeded` can't trigger a bogus model fallback.
- **O-005** Critic-oversight dispatch must fail over to configured `fallback_models` on transient/quota errors.
- **O-006** Lean-turbo reviewer dispatch must fail over to configured `fallback_models` on transient/quota errors.
- **O-007** Lane-runner dispatch must fail over to configured `fallback_models` on transient/quota errors.
- **O-008** Evaluation dispatcher must retry the SAME model on quota (never substitute — model ID is the benchmark subject).
- **O-009** Non-transient circuit-breaker hard-stop message must surface the actual failure signal and category-specific remediation.
- **O-010** Architect session must detect a silent model switch across interrupt/resume (one-shot advisory).
- **O-011** Architect session must note configured model vs UI-active model mismatch (one-shot advisory).
- **O-012** Scope-binding activation must source parent from `part.sessionID` (not `metadata.parentSessionId` which 1.1.x runtimes never emit).
- **O-013** `python -m <module>` must NOT be treated as interpreter eval (issue #1902).
- **O-014** Null-path write blocks must be discriminated by cause (here-doc skip, eval message, dynamic-path message).
- **O-015** Reviewer dispatch timeout must be explicitly classified as permanent (defense-in-depth against future message rewording).

---

### 📦 Implementation Summary

The PR creates two new utility modules (`provider-error-classification.ts`, `model-dispatch-fallback.ts`), removes three duplicated regex/constant definitions, and wires model-fallback into four dispatch stages (oversight bespoke loop, reviewer via shared helper, runner via shared helper, evaluator same-model retry). It adds NFC canonicalization and §-folding to the acceptance normalizer plus a bounded diagnostic function. The scope-binding activation in `src/index.ts` switches from `metadata.parentSessionId` to `part.sessionID`. The shell-write detector removes `-m` from Python eval flags. The non-transient circuit message now includes the stored signal and per-category remediation. Model-divergence detection is added to the architect session's messages-transform, with two new persisted fields on `AgentSessionState`. Four release-note fragments and seven new test files are added.

---

|---|---|---|
| O-001 | `SUPPORTED` | `src/hooks/delegation-gate.ts:813` — `.normalize('NFC')` added; `:829` — `Â?§` → ` section ` fold |
| O-002 | `SUPPORTED` | `src/hooks/delegation-gate.ts:836-903` — `describeCoverageMiss` + wired at `:936` |
| O-003 | `SUPPORTED` | `src/utils/provider-error-classification.ts` created; duplicates removed from `guardrails/index.ts:120-131` (deleted) and `messages-transform.ts:52-63` (deleted); `oversight.ts:197` imports shared |
| O-004 | `SUPPORTED` | `provider-error-classification.ts:58-63` — `QUOTA_ERROR_PATTERN` separate; `:68-73` — combined ONLY in `isTransientProviderError`; `TRANSIENT_MODEL_ERROR_PATTERN` unchanged (no quota) |
| O-005 | `SUPPORTED` | `src/full-auto/oversight.ts:407-588` — bespoke fallback loop with `resolveOversightFallback` |
| O-006 | `SUPPORTED` | `src/turbo/lean/reviewer.ts:582-645` — `dispatchWithModelFallback` wrapper |
| O-007 | `SUPPORTED` | `src/turbo/lean/runner.ts:1185-1248` — `dispatchWithModelFallback` wrapper |
| O-008 | `SUPPORTED` | `src/evaluation/model-dispatcher.ts:227-269` — same-model retry loop, no substitution |
| O-009 | `SUPPORTED` | `src/hooks/guardrails/nontransient-circuit.ts:299-327` — `nonTransientRemediation` + signal in `:330-341` |
| O-010 | `SUPPORTED` | `src/hooks/guardrails/messages-transform.ts:168-198` — resume advisory, one-shot guard |
| O-011 | `SUPPORTED` | `src/hooks/guardrails/messages-transform.ts:201-218` — config-vs-UI advisory, one-shot guard |
| O-012 | `SUPPORTED` | `src/index.ts:1385-1415` — `eventParentSessionID` from `part.sessionID`, empty-id fail-closed |
| O-013 | `SUPPORTED` | `src/hooks/shell-write-detect.ts:668-680` — `-m` removed from python eval flags |
| O-014 | `SUPPORTED` | `src/hooks/guardrails/tool-before.ts:984-1012` — three-way discrimination |
| O-015 | `SUPPORTED` | `src/turbo/lean/reviewer.ts:607-609` — `Reviewer dispatch timed out` → `'permanent'` |

---

### 🚨 Confirmed Findings

_None found._

---

_None found._

---

### 🧪 Test / Coverage Gaps

_None found._ The PR adds 7 new test files covering: encoding normalization, coverage-miss diagnostics, oversight bespoke fallback, reviewer fallback (shared helper), evaluator same-model retry, model-divergence advisories, non-transient diagnostic message, scope-activation event shapes (including spoof/empty-id guards), shell-write `-m` allowlist, and a structural drift-guardrail test that prevents re-introduction of duplicated transient regexes. Pre-existing test files updated to use `toMatchObject` instead of `toEqual` for the widened return type. All claims in the PR description (test counts, pass/fail parity with main) are internally consistent.

---

### 📋 Shipped-vs-Claimed Gaps

_None found._

---

### 🔁 Validation provenance

- **DROPPED 0 findings** — the reviewer-confirmed list was empty; no candidates to challenge.
- **ADDED 0 blind-spot findings** — I examined every `+` line across all 15 changed source files, traced cross-file contract consistency (classifier imports vs usage sites, `ModelOverride` shape at all `client.session.prompt` call sites, snapshot round-trip for new session fields), verified loop termination and retry-count arithmetic in all four dispatch stages, checked the oversight loop's increment-before-parse behavior against its test expectations, confirmed the quota/tool-output separation invariant, and verified the scope-activation security property (metadata cannot override runtime-assigned parent). No structurally-proven defect was found.

---

### 📝 Merge Recommendation

**APPROVE**

A well-executed, thoroughly-tested PR that resolves six interconnected defects with clean separation of concerns, correct cross-file contracts, and comprehensive adversarial/regression test coverage.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| No hardcoded secrets | ✅ |
| Lockfile consistent | ✅ |

---
<sub>🔒 Reviewed by a multi-stage local-first funnel (architect context pack → explorer candidates → critic challenge/author) for high recall with low false-positive noise. Findings are advisory — verify before acting.</sub>

Issue #1047, opened 2026-05-27
build: stop committing generated dist artifacts


zaxbysauce:
﻿## Goal

Stop committing generated `dist/` artifacts in ordinary feature PRs while keeping npm package validation airtight.

This is PR 3 in the CI/release simplification plan. It should happen only after PR 1 proves package artifact validation and release build parity, and PR 2 prepares merge queue/current-base validation.

## Current problem

`dist/` is generated output but currently tracked in git. The repo has hundreds of tracked `dist` files, and CI enforces that committed `dist/` matches a fresh build. This catches stale bundles, but it also means unrelated source PRs frequently conflict or go stale because generated output changes whenever another PR lands.

The common failure mode is:

1. PR A and PR B both change source.
2. Both include regenerated `dist/`.
3. PR A merges.
4. PR B now has stale or conflicting generated files.
5. PR B must rebase, rebuild, recommit generated output, and rerun CI even if its source still applies cleanly.

The user has confirmed that no users install this project directly from the GitHub checkout, so npm package correctness is the distribution contract that matters.

## Proposed change

Remove generated `dist/` artifacts from source control and make CI validate the package artifact that would actually be published.

After this change, normal source PRs should not include generated bundle churn. CI and release workflows build `dist/` from source and verify that the packed npm artifact is complete, importable, and runnable.

## Preconditions

Do not start this PR until:

- [ ] PR 1 has aligned release publishing with CI build behavior.
- [ ] PR 1 has added a passing `package-check` or equivalent packed-artifact validation job.
- [ ] Package validation confirms declaration files are present in the tarball.
- [ ] Package validation confirms grammar assets are present in the tarball.
- [ ] Package validation installs the generated `.tgz` in a temp project and smoke-tests import/CLI behavior.
- [ ] PR 2 has added merge queue support or there is a clear decision to postpone merge queue independently.

## Scope

- Add `dist/` to `.gitignore` as generated output.
- Remove tracked `dist/` files from git.
- Remove or rewrite the old `dist-check` job that diffs committed `dist/`.
- Keep package artifact validation as a required check.
- Update docs and contribution guidance so agents/humans stop committing generated `dist/` for feature PRs.
- Update release guidance to make clear that release/publish builds `dist/` from source.
- Keep runtime portability checks that import the built package under Node.

## Implementation checklist

- [ ] Confirm `package-check` is green on current `main` before removing tracked artifacts.
[...]

PR #2006, opened 2026-08-02
fix(scope): resolve coder write scope against the session's own workspace root

### LINKED ISSUE #2002: Scope validation fails for worktree-isolated coders: child binding not found at write time despite deriveChildScopeBinding path (SCOPE_NOT_DECLARED)
## Summary

Worktree-isolated coder dispatches fail with `SCOPE_NOT_DECLARED` at write time, even though the child scope binding is derived and published by the delegation gate. The `deriveChildScopeBinding` → `publishScopeBinding` → `getAuthorizedScopeBindingByPlanIdentity` chain is **correct by static inspection**, but the coder's scope-guard cannot find the binding at runtime. This makes `worktree.policy: "required"` (or `"auto"` with `parallelization_enabled: true`) unusable for coder dispatch on this setup.

This blocks all parallel/standard-turbo coder dispatch — every coder write is rejected, so no implementation work can proceed under worktree isolation.

## Environment

- Platform: **Windows** (`process.platform === 'win32'`)
- `worktree.policy: "required"` (config)
- `parallelization.enabled: true`, `maxConcurrentTasks: 4`
- Plan `execution_profile`: `parallelization_enabled: true`, `max_concurrent_tasks: 4`, `locked: true`
- Plugin loaded from local dev `dist/` (no stale cache — all three cache layouts absent; `dist/index.js` built after source; `deriveChildScopeBinding` present in the bundle)
- Fix #1914 (`describeCoderScopeFailure`) IS present — the `SCOPE_NOT_DECLARED` error is decorated with known plan task IDs and FILE detection, confirming the loaded runtime is current.

## Symptom

```
SCOPE_NOT_DECLARED: mega_coder cannot invoke edit without a validated scope for task 1.1
```

Reproduced across **6 distinct dispatch attempts**: parallel (2 coders), serial (1 coder), fresh `declare_scope` immediately before dispatch, worktree-lane cleanup between attempts, and multiple `task_id` forms (plan task ID `"1.1"`, omitted, session ID). All fail identically: the coder session starts and runs, but its first write tool call is rejected.

The dispatch itself **succeeds** (no throw from the delegation gate) — the coder session is created and runs. Only the write-time scope-guard (`src/hooks/scope-guard.ts:137`) rejects the write.

## Investigation: the publish → validate chain

I traced the complete path. Every step is correct by code inspection.

### 1. `prepareCoderScope` (delegation-gate.ts:221)
Loads the plan from the main repo (`loadPlanJsonOnly(directory)`), resolves the plan task ID (`resolveDelegatedPlanTaskId`), and creates the parent binding (`createScopeBinding`). Returns `{ plan, taskId, declaredFiles, binding }`. **Confirmed: succeeds** (the coder runs, so no throw at line 302).

### 2. `standardWorktreeIsolationActive` (delegation-gate.ts:2601)
```ts
const standardWorktreeIsolationActive =
    parallelEnabled &&
    effectiveMaxConcurrent > 1 &&
    !hasActiveLeanTurbo(input.sessionID);
```
With `parallelization_enabled: true` and `max_concurrent_tasks: 4`, this is `true`. Worktrees ARE provisioned (confirmed via `git worktree list`). So the worktree path at line 2748+ is taken (not the early-return at line 2745).

### 3. `precreateStandardWorktreeSession` (delegation-gate.ts:2753 → worktree-isolation.ts:850)
- Provisions the worktree via `provisionWorktree` (line 854)
- Creates the child session: `client.session.create({ query: { directory: provisionResult.worktreePath } })` (line 924-930)
- Overwrites `args.outputArgs.task_id = createResult.data.id` (line 953)
- Calls `rememberStandardWorktreeDispatch({...})` (line 954) → populates `standardWorktreeByCallID`

### 4. Scope materialization block (delegation-gate.ts:2768-2812)
Gated on `standardDispatch` (line 2768) and `resolvedTaskId` (line 2770). Both should be truthy (the dispatch succeeded and `prepareCoderScope` resolved the task ID).

Inside the block:
- `savePlan(worktreePath, plan, { preserveCompletedStatuses: false })` (line 2782) — materializes the plan in the worktree
- `deriveChildScopeBinding(correlatedBinding, { childDirectory: worktreePath, childSessionId, parentCallId })` (line 2785) — re-roots the binding
- `publishScopeBinding(callID, worktreePath, childBinding)` (line 2790) — registers + writes to disk
- `childSession.declaredCoderScope = [...childBinding.files]` (line 2801)

### 5. `deriveChildScopeBinding` (scope-binding.ts:416)
```ts
const workspaceIdentity = canonicalWorkspaceIdentity(input.childDirectory);
return {
    ...parent,
    workspaceIdentity,
    ownerSessionId: input.childSessionId,
    ownerMessageId: input.parentCallId,
    dispatchCallId: input.parentCallId,
    activation: 'active',
    parentOwnerSessionId: parent.ownerSessionId,
    parentCallId: input.parentCallId,
    source: 'worktree_derived',
    ...
};
```

### 6. `publishScopeBinding` (delegation-gate.ts:2030)
- `registerScopeBinding(binding)` (line 2035) — adds to the module-level `pendingScopeBindings` map (scope-binding.ts:39)
- `writeScopeBindingToDisk(bindingDirectory, binding)` (line 2036)

### 7. `writeScopeBindingToDisk` (scope-persistence.ts:333)
Creates `.swarm/scopes/` (line 357: `fs.mkdirSync(scopesDir, { recursive: true })`). **But has a silent guard** (line 340-348):
```ts
if (
    !workspaceIdentity ||
    workspaceIdentity !== binding.workspaceIdentity ||  // ← silent skip if mismatch
    ...
)
    return;  // no throw, no log
```
If `canonicalWorkspaceIdentity(directory)` at write time ≠ `binding.workspaceIdentity` (set at derive time), the disk write is **silently skipped**.

### 8. Write-time validation (scope-guard.ts:99-138)
```ts
let taskId = _internals.resolveTaskId(session);
let binding = taskId
    ? _internals.resolveAuthorizedScopeBinding({ directory, taskId, activeSessionId: sessionId })
    : _internals.resolveAuthorizedScopeBindingForSession({ directory, activeSessionId: sessionId });
```

### 9. `getAuthorizedScopeBindingByPlanIdentity` (scope-binding.ts:326)
The in-memory lookup. Filters `pendingScopeBindings` by **11 conditions** (line 336-351):
1. `binding.workspaceIdentity === workspaceIdentity`
2. `binding.planId === input.planId`
3. `binding.planStructureHash === input.planStructureHash`
4. `binding.taskId === input.taskId`
5. `binding.ownerSessionId === input.activeSessionId`
6. `binding.activation === 'active'`
7. `dispatchCallId` is a non-empty string
8. `ownerMessageId === dispatchCallId`
9. `parentOwnerSessionId` is a non-empty string
10. `ownerSessionId !== parentOwnerSessionId`
11. `parentCallId === dispatchCallId`

I verified each field derivation for the child binding — **all 11 should pass**. Yet the lookup returns no match.

The fallback (`readScopeBindingFromDisk`, scope-persistence.ts:542) reads from the coder's worktree `.swarm/scopes/`. If the silent identity guard in `writeScopeBindingToDisk` (step 7) skipped the write, the disk fallback also returns nothing.

## Where the gap likely is

The code is correct by static inspection — every condition should pass. This points to a **runtime/context condition**:

### Hypothesis A — `canonicalWorkspaceIdentity` resolves differently at publish vs validate time
`canonicalWorkspaceIdentity` (scope-binding.ts:41) uses `fs.realpathSync(directory)`:
```ts
const real = fs.realpathSync(directory).replace(/\\/g, '/');
return process.platform === 'win32' ? real.toLowerCase() : real;
```
The worktree lives **outside the repo** at `E:/OpenCode/.swarm-worktrees/<session>/<task>`. On Windows, `realpathSync` may resolve this differently at publish time (inside the delegation gate, architect context) vs validate time (inside the coder's scope-guard) — e.g. if a junction is involved, or if the directory is being created/modified between the two calls. A mismatch here fails condition #1 silently in both the in-memory filter AND the disk-write guard.

### Hypothesis B — process/context boundary for the in-memory map
`pendingScopeBindings` is a module-level `Map` (scope-binding.ts:39). If the worktree session's scope-guard runs in a context that does not share this map (e.g. a separate plugin instance or worker for the worktree session), the in-memory binding is invisible. The disk path becomes the only path — and if Hypothesis A's identity guard skipped the disk write, both paths fail.

### Hypothesis C — `planStructureHash` mismatch
`computePlanStructureHash` (plan/ledger.ts:257) includes `files_touched` (line 279). The child binding inherits the parent's hash (computed on the main-repo plan). The scope-guard computes the hash from `readCurrentPlan(worktreeDir)` (scope-persistence.ts:558 → reads `.swarm/plan.json` from the worktree). If the worktree's plan (saved via `savePlan(..., { preserveCompletedStatuses: false })`) normalizes `files_touched` differently, the hashes diverge → condition #3 fails. *(Note: statuses are excluded from the hash per the comment at ledger.ts:258, but `files_touched` is included.)*

## Reproduction

1. Set `worktree.policy: "required"` and `parallelization.enabled: true` (maxConcurrent > 1) on Windows.
2. Create a plan with `parallelization_enabled: true` in the execution profile.
3. `declare_scope({ taskId: "1.1", files: [...] })`.
4. Dispatch a coder via `Task(subagent_type="<swarm>_coder", ...)` for task 1.1.
5. The coder session starts but every `edit`/`write` call fails with `SCOPE_NOT_DECLARED: ... cannot invoke edit without a validated scope for task 1.1`.

### Step 1 — Add a diagnostic log to pinpoint the failing condition
In `getAuthorizedScopeBindingByPlanIdentity` (scope-binding.ts:336), when a `worktree_derived` binding exists in `pendingScopeBindings` for the same `taskId` + `ownerSessionId` but doesn't match the full filter, log **which** condition failed:

```ts
// Diagnostic: if a worktree_derived binding for this task+session exists but
// didn't match, surface the first failing condition.
const candidates = [...pendingScopeBindings.values()].filter(
    (b) => b.source === 'worktree_derived' &&
           b.taskId === input.taskId &&
           b.ownerSessionId === input.activeSessionId,
);
if (candidates.length > 0 && matches.length === 0) {
    const b = candidates[0];
    const failed: string[] = [];
    if (b.workspaceIdentity !== workspaceIdentity) failed.push('workspaceIdentity');
    if (b.planId !== input.planId) failed.push('planId');
    if (b.planStructureHash !== input.planStructureHash) failed.push('planStructureHash');
    // ... etc for each condition
    logger.warn(`scope-worktree: child binding found but mismatched on [${failed.join(', ')}]`);
}
```

One reproduction will identify the exact failing field.

### Step 2 — Verify the process-boundary assumption (Hypothesis B)
Confirm the worktree session's scope-guard shares the same module-level `pendingScopeBindings` as the architect's delegation gate. If OpenCode isolates worktree sessions into a separate plugin context, the in-memory path is broken by design and the disk path must be made robust (remove or log the silent identity-guard skip in `writeScopeBindingToDisk`).

### Step 3 — If Hypothesis A (realpathSync), harden `canonicalWorkspaceIdentity`
Make the identity derivation deterministic regardless of when `realpathSync` is called — e.g. normalize the provisioned worktreePath once at provision time and reuse the cached identity, rather than re-resolving via `realpathSync` at each lookup.

## Workaround

Set `worktree.policy: "disabled"` (valid values: `"auto"` / `"required"` / `"disabled"` — note `"off"` is invalid; `"auto"` still engages isolation when `parallelization_enabled` is true). This makes coders run in the main repo where the scope binding's `workspaceIdentity` matches the coder's directory and the in-memory map is unambiguously shared.

## Related

- #1914 — the prior task_id-collision fix (`describeCoderScopeFailure`); confirmed present in the loaded build.
- #1746 item 1 — worktree provisioning auto-cleanup (referenced by the `worktree-retry-cleanup` skill).
- `docs/architecture.md:1672` — "Worktree lanes derive a child-root binding rather than reusing root authority" (the contract this issue violates at runtime).
- Covered by `tests/unit/scope/scope-binding-identity.test.ts:259` and `tests/unit/hooks/scope-guard-identity-binding.test.ts:167` (these tests pass, suggesting the gap is a runtime path the tests don't exercise — likely the `realpathSync` or process-boundary condition).

## Key file references

| File | Line | Role |
|------|------|------|
| `src/scope/scope-binding.ts` | 41 | `canonicalWorkspaceIdentity` (realpathSync — Hypothesis A) |
| `src/scope/scope-binding.ts` | 326 | `getAuthorizedScopeBindingByPlanIdentity` (11-condition filter) |
| `src/scope/scope-binding.ts` | 416 | `deriveChildScopeBinding` (re-roots binding onto worktree) |
| `src/scope/scope-persistence.ts` | 333 | `writeScopeBindingToDisk` (silent identity guard, line 342) |
| `src/scope/scope-persistence.ts` | 520 | `readCurrentPlan` (reads worktree `.swarm/plan.json`) |
| `src/scope/scope-persistence.ts` | 553 | `resolveAuthorizedScopeBinding` (scope-guard entry) |
| `src/hooks/delegation-gate.ts` | 2030 | `publishScopeBinding` (register + disk write) |
| `src/hooks/delegation-gate.ts` | 2601 | `standardWorktreeIsolationActive` gate |
| `src/hooks/delegation-gate.ts` | 2768-2812 | scope materialization block (worktree child binding) |
| `src/hooks/delegation-gate/worktree-isolation.ts` | 924-964 | worktree session creation + `rememberStandardWorktreeDispatch` |
| `src/hooks/scope-guard.ts` | 99-138 | write-time scope validation |
| `src/plan/ledger.ts` | 257 | `computePlanStructureHash` (includes `files_touched`, line 279) |

zaxbysauce:
Closes #2002

## Summary

Worktree-isolated coder dispatches failed on their first write with
`SCOPE_NOT_DECLARED: mega_coder cannot invoke edit without a validated scope for task 1.1`,
even though the delegation gate correctly derived and published a child scope binding rooted
at the coder's own lane directory. This made `worktree.policy: "required"` (and `"auto"` with
`parallelization_enabled: true`) unusable for coder dispatch.

Two independent defects produced the same symptom:

1. **Standard worktree dispatch.** Both write gates (`src/hooks/scope-guard.ts`,
   `src/hooks/guardrails/tool-before.ts`) are constructed once at plugin init and captured the
   plugin-root `ctx.directory` as a closure constant, then applied that one root to every
   session's tool calls at runtime. A worktree-isolated coder child executes in a different,
   lane-rooted directory, and its scope binding is derived and published against that lane
   root — so the identity-bound lookup on the *validate* side could never match the *derive/
   publish* side, and the on-disk fallback looked in `<projectRoot>/.swarm/scopes/` for a file
   that was written to `<worktreePath>/.swarm/scopes/`.
2. **Lean Turbo lanes.** Independently, Lean Turbo created lane coder sessions with `write`/
   `edit`/`patch` enabled but never published a scope binding for them and never materialized
   the plan into the lane, so there was no authority to find under *any* root. Every Lean Turbo
   lane coder was blocked on its first write, for an unrelated reason (`04-root-cause.md`,
   `05-fix-plan.md` §"Scope disposition: Lean Turbo lanes").

## Root cause

`src/index.ts:1224-1230` constructs the scope guard with `ctx.directory` (the plugin root):

```ts
const scopeGuardHook = createScopeGuardHook(
  { enabled: watchdogConfig.scope_guard, skip_in_turbo: watchdogConfig.skip_in_turbo },
  ctx.directory,          // <-- project root, for ALL sessions
  advisoryInjector,
);
```

That fixed root feeds `resolveAuthorizedScopeBinding{,ForSession}` at `scope-guard.ts:101-109`,
which calls `readCurrentPlan(directory)` and `getAuthorizedScopeBindingByPlanIdentity`
(`src/scope/scope-binding.ts:326`). Filter condition #1 there,
`binding.workspaceIdentity === workspaceIdentity` (`scope-binding.ts:338`), compares
`canonicalWorkspaceIdentity(projectRoot)` against the child binding's `workspaceIdentity`, which
`deriveChildScopeBinding` (`scope-binding.ts:424`) set to the **worktree** identity — always
unequal. The disk fallback `readScopeBindingFromDisk({ directory, … })`
(`src/scope/scope-persistence.ts:542`) looks in `<projectRoot>/.swarm/scopes/`; the binding was
written to `<worktreePath>/.swarm/scopes/` (`delegation-gate.ts:2791-2795`) and is absent there.
Both paths return `null`, `declaredScope` is `null`, and `scope-guard.ts:136-138` throws
`SCOPE_NOT_DECLARED`.

`src/hooks/guardrails/tool-before.ts` (constructed the same way at `src/index.ts:1072-1078`) has
the identical defect on several sibling paths: coder shell writes (`tool-before.ts:983-987`),
absolute-lane-path containment (`:2260-2269`), the OS sandbox filesystem grant (`:1161`),
lstat/symlink checks (`:2192-2195`), and destructive-command containment
(`:249, 277, 294, 304-384, 561-565, 584`). It does **not** block the reported direct `edit` —
the `coder` role has no `allowedPrefix` (`src/hooks/guardrails/file-authority.ts:268-271`), so a
null `declaredScope` is a no-op there — which is corroborated by the issue's own symptom text:
`SCOPE_NOT_DECLARED` is emitted only by `scope-guard.ts:136-138`, so guardrails had already
allowed the call. Full trace in `.agents/issue-traces/2002-worktree-scope-binding-not-found/04-root-cause.md`.

Both of the issue's own hypotheses were investigated and refuted with evidence (not by
elimination) in `02-reproduction.md`/`04-root-cause.md`: Hypothesis A (`realpathSync` resolving
differently on Windows) is platform-independent — the repro fails identically on one volume with
no junctions, and the mismatch is between two genuinely different directories, not two
resolutions of one directory. Hypothesis B (separate plugin instance / unshared module map) is
refuted by the error's own `task 1.1` text, which is only writable from the parent context on a
shared `swarmState` singleton.

## Fix

- **`src/state.ts`** — new `workspaceDirectory?: string` field on `AgentSessionState`, written
  **only** by a new, trusted-source-only setter `recordSessionWorkspaceRoot(sessionId, laneRoot)`.
  The setter performs no path validation; trust comes entirely from a closed allowlist of two
  production call sites, each passing `provisionWorktree`'s own return value (never a tool
  argument). Read via `resolveSessionWorkspaceDirectory(sessionId, fallbackDirectory)`, which
  falls back to the plugin-root directory when no root was recorded — byte-identical to
  pre-#2002 behavior for every non-worktree session.
- **`src/hooks/delegation-gate/worktree-isolation.ts:989-997`** — calls
  `recordSessionWorkspaceRoot(childSessionId, provisionResult.worktreePath)` immediately after
  the child session is created, unconditionally (not gated on a resolvable plan task id), so
  every lane-rooted session gets a recorded root.
- **Security-relevant behaviour change:** `worktree-isolation.ts:989-993` now also calls
  `ensureAgentSession(createResult.data.id, 'coder', provisionResult.worktreePath)`
  **unconditionally**, immediately before the `recordSessionWorkspaceRoot` call above. Before
  this diff, the only child-session registration for the standard worktree dispatch path was
  `delegation-gate.ts:2796-2800`, which runs **only inside `if (resolvedTaskId)`**
  (`delegation-gate.ts:2771`) — a lane dispatched without a resolvable plan task id got no
  `swarmState.activeAgent` entry at all. This new unconditional call writes
  `swarmState.activeAgent` — the map every guardrails identity decision reads
  (`tool-before.ts:2224-2228`, `:996-1001`) — on a code path that previously had **no entry
  whatsoever**. It is verified fail-closed: `recordSessionWorkspaceRoot` is a documented no-op
  if the session does not already exist (`src/state.ts:2028-2050`), and registering with the
  real `'coder'` agent name *before* recording the workspace root is what prevents the
  `'unknown'`-agent fail-open trap described in "A note on review history" below. Covered by
  `tests/unit/hooks/delegation-gate-worktree-isolation-active-agent.test.ts`.
- **`src/hooks/scope-guard.ts`** — resolves the session's actual root per call via
  `resolveSessionWorkspaceDirectory` instead of using the captured constructor argument (renamed
  `fallbackDirectory`), at all three sites that previously used the fixed root:
  `resolveWriteTargets`, `resolveAuthorizedScopeBinding{,ForSession}`, `isFileInScope`.
- **`src/hooks/guardrails/tool-before.ts`** — same treatment, but *selective*: every
  session-scoped decision (`resolveActiveScopeBinding`, direct-write resolution, symlink/lstat,
  universal-deny, authority + declared scope, shell-write parsing, sandbox path grant,
  destructive-command containment) is re-rooted through `sessionWorkspaceDirectory(sessionID)`.
  The plan/spec write guards (`:1691-1799`, protecting the *project's* `.swarm/plan.json` /
  `.swarm/plan.md` / `.swarm/spec-staleness.json`) and `enforceSpecDriftGate` (`:2185`) are
  deliberately left on the raw `effectiveDirectory` — these protect project governance
  artifacts, not the agent's working files, and re-rooting them would let a lane coder overwrite
  its own lane-local plan copy while believing it protected the real plan.
- **`src/scope/scope-binding.ts`** — new `describeScopeWorkspaceMismatch` diagnostic (see
  Recurrence guardrail).
- **`src/session/snapshot-reader.ts`** — `workspaceDirectory` is deliberately **not** restored
  from a snapshot (see Known limitations).
- **`src/turbo/lean/lane-scope.ts`** (new) — `publishLeanTurboLaneScopeBinding` mints and
  publishes a Lean Turbo lane's write authority using the same primitives as the standard
  worktree path (`createScopeBinding` → `deriveChildScopeBinding` → `registerScopeBinding` +
  `writeScopeBindingToDisk`), scoped to the intersection of the lane's already-locked files and
  the plan's `files_touched` for the lane's own task ids. For isolated lanes it also materializes
  the plan into the lane and calls `recordSessionWorkspaceRoot(childSessionId, laneRoot)`.
  Fail-closed: no representative task id, no plan-backed files, or a non-distinct parent/child
  session id publishes nothing and leaves the lane coder blocked exactly as before this fix.
- **`src/turbo/lean/runner.ts`** — wires the new module in: threads the authoritative `plan`
  through `dispatchLane` → `_processLane` → `_dispatchLaneInternal`, which calls a new
  `_publishLaneScope` helper before the coder's prompt is built. An unauthorizable lane (`null`)
  logs and pushes an `LEAN_TURBO_LANE_UNSCOPED` architect advisory but leaves the lane blocked at
  the ordinary scope gate; an unexpected publish failure aborts the dispatch and fails
  permanently via a new `LANE_SCOPE_DENIED_CODE` sentinel so a broken authority handshake is
  never silently retried as a provider blip.

## Acceptance Criteria → Evidence

The issue is a bug report/investigation, not a formal AC checklist; the criteria below are the
issue's stated symptom, its two named hypotheses, and its own suggested next-step verifications
(from the "Next steps" / "Related" sections and file-reference table).

| Acceptance criterion (from intake) | Evidence |
|---|---|
| Worktree-isolated coder can perform an in-scope write without `SCOPE_NOT_DECLARED` | `tests/unit/hooks/scope-guard-worktree-session-root.test.ts` — 11/11 pass (`08-test-results.md`); regression test constructs the hook with the **project root** (production wiring) and a coder session in a worktree, asserts the write is allowed |
| Out-of-scope writes from a worktree-isolated coder are still blocked (no security regression) | Same file, negative cases; still throws `SCOPE VIOLATION` / `SCOPE_WORKSPACE_MISMATCH` |
| Guardrails-level paths (shell writes, absolute lane paths, sandbox grant, destructive-command containment) fixed identically | `tests/unit/hooks/guardrails-worktree-session-root.test.ts` — 10/10 pass |
| Hypothesis A (`realpathSync`/Windows path resolution) — investigate and report | Refuted with evidence in `02-reproduction.md`/`04-root-cause.md`: repro fails identically on one volume, no junctions/symlinks; platform-independent |
| Hypothesis B (separate plugin instance / unshared module map) — verify the process-boundary assumption | Refuted with evidence: the error's `task 1.1` text is writable only from the parent context on a shared `swarmState` singleton (`scope-guard.ts:99, 231-232`, `delegation-gate.ts:2801`) |
| `tests/unit/scope/scope-binding-identity.test.ts:259` and `tests/unit/hooks/scope-guard-identity-binding.test.ts:167` (cited by the issue as passing-but-not-covering-the-gap) | `scope-guard-identity-binding.test.ts` fixed in this diff (it constructed the hook with the worktree directory, masking the class); now asserts against production wiring |
| Workaround (`worktree.policy: "disabled"`) no longer required | Fix operates with worktree isolation enabled; all regression tests run with `worktree.policy` effectively `"required"` semantics (lane-rooted sessions) |
| No regression to non-worktree / serial dispatch | `tests/unit/hooks/scope-guard-worktree-session-root.test.ts` and `guardrails-worktree-session-root.test.ts` include an unrecorded-root case asserting fallback to project-root behavior is byte-identical to today |

## Tests and checks run

Regenerated in this session against the frozen, code-complete diff staged at
`git diff --cached | git hash-object --stdin` = **`cb3085ae0b54f7d1c178452cc487327b9b8d138f`**,
verified identical at the start and end of this session; each test file run **individually** per
AGENTS.md invariant 7 (`bun test <file>`, not a directory sweep). Full raw output and the complete
12-file list (the entire `*.test.ts` set in `git diff --cached --name-only`) is in
`08-test-results.md`.

```
$ bun run typecheck            -> tsc --noEmit, EXIT:0
$ bunx biome ci .               -> Checked 2931 files in 1382ms. No fixes applied. EXIT:0
$ bash .opencode/skills/issue-tracer/scripts/scan-deferred.sh
  -> scan-deferred: clean (base=origin/main) — no deferred-work markers on added lines. EXIT:0
```

| File | Result |
|---|---|
| `tests/unit/hooks/delegation-gate-worktree-isolation-active-agent.test.ts` | 1 pass, 0 fail |
| `tests/unit/hooks/guardrails-worktree-session-root.test.ts` | 10 pass, 0 fail |
| `tests/unit/hooks/scope-guard-identity-binding.test.ts` | 3 pass, 0 fail |
| `tests/unit/hooks/scope-guard-worktree-session-root.test.ts` | 11 pass, 0 fail |
| `tests/unit/scope/scope-workspace-mismatch.test.ts` | 4 pass, 0 fail |
| `tests/unit/session/snapshot-workspace-directory-omission.test.ts` | 3 pass, 0 fail |
| `tests/unit/state/session-workspace-root-trusted-callers.test.ts` | 4 pass, 0 fail |
| `tests/unit/turbo/lean/lean-turbo-lane-dispatch-tools-gate.test.ts` | 2 pass, 0 fail |
| `tests/unit/turbo/lean/lean-turbo-lane-scope-cleanup-symmetry.test.ts` | 3 pass, 0 fail |
| `tests/unit/turbo/lean/lean-turbo-lane-scope-wiring.test.ts` | 9 pass, 0 fail |
| `tests/unit/turbo/lean/lean-turbo-lane-scope.test.ts` | 21 pass, 0 fail |
| `tests/unit/turbo/lean/runner.test.ts` | 77 pass, 0 fail |
| **Total** | **148 pass, 0 fail across 12 changed test files** |

`bash scripts/check-invariants.sh` exits 1 with 743 violations on this branch, and produces the
byte-identical result (exit 1, 743) on a clean `origin/main` worktree — proven pre-existing, zero
violations in any file this diff touches. `check-bash-portability.sh`, `check-cross-contamination.sh`,
`check-mock-cleanup.sh`, `check-no-raw-advisory-push.sh`, `check-test-clock.sh`,
`check-test-tmpdir.sh`, and `check-test-file-cap.sh` all pass (EXIT:0); full detail in
`08-test-results.md`.

A prior per-file sweep of 89 scope/guardrails/state test files plus 68 turbo/lean,
delegation-gate-worktree, and session-snapshot test files found exactly one failure:
`tests/unit/hooks/delegation-gate-scope-preflight.test.ts`, thrown from `fs.rmSync` in
`afterEach` with `EBUSY: resource busy or locked` — a Windows filesystem teardown race, not an
assertion failure. Confirmed pre-existing and unrelated to this branch: the file is untouched by
the diff (`git diff --cached --stat -- tests/unit/hooks/delegation-gate-scope-preflight.test.ts`
is empty), and re-running it 3× in isolation this session produced 2 clean passes and 1 failure
in the same `afterEach` teardown location, not a test assertion. Full detail in
`08-test-results.md`.

Two `SyntaxError: Export named 'recordSessionWorkspaceRoot' not found` CI breaks were found and
fixed in `tests/unit/commands/close.test.ts` and `tests/unit/commands/close-finalizer.test.ts`,
both of which mock `src/state.ts` with a non-spreading object-literal `mock.module()`. Fixed by
adding `recordSessionWorkspaceRoot` and `resolveSessionWorkspaceDirectory` stubs to the shared
`tests/unit/commands/state-mock-transitive-stubs.ts` fixture (`STATE_MOCK_TRANSITIVE_STUBS`),
which both test files import.

### Publication validation suite (commit 2be4bc2f, base af62d1d4)

- `bun run build` OK; `node --input-type=module -e "await import('./dist/index.js')"` → `dist import OK`. `dist/` not staged (#1047).
- `bun run typecheck` exit 0 · `bun run lint:ci` (pinned Biome) — 2931 files, no fixes, exit 0.
- `scripts/check-tool-registration.ts` — 115 tools coherent. `check-mock-cleanup.sh`, `check-cross-contamination.sh`, `check-test-clock.sh`, `check-test-file-cap.sh` — all pass (pre-existing non-blocking counts only).
- **Tier 2 (CI-equivalent):** `bun run test:unit:ci <12 changed test files>` — exit 0, 0 quarantined, every file exitCode 0. 148 pass / 0 fail.
- **Tier 4:** `bun test tests/security` 163 pass / 4 skip / 0 fail · `bun test tests/adversarial` 846 pass / 0 fail.
- **Tier 5:** `bun test tests/smoke` 10 pass / 0 fail.

### Integration: why the bulk number is not the gate

CI runs integration **per file** (`.github/workflows/ci.yml:540` — `bun --smol test "$f"` in a loop with retries), not as one directory invocation. Run that way over 78 files, this branch has **2** failures. The bulk `bun test tests/integration ./test` form reports 214 — a `mock.module` cross-file pollution artifact (AGENTS.md invariant 7), demonstrated by `test/repro-plan-md-not-updated.test.ts`, which fails inside the bulk run and passes 8/8 in isolation.

## Pre-existing failures

Both per-file integration failures reproduce **identically on a clean `canonical/main` worktree (af62d1d4)** — same test names, same counts — so they are inherited, not introduced. Neither appears in `scripts/ci/quarantined-integration-tests.txt` (that list is currently empty).

| Test | This branch | Clean main |
|---|---|---|
| `tests/integration/cross-process-port-binding.test.ts` — "lane 0 (PORT=9100) and lane 1 (PORT=9101) with stride=1 both bind" | 9 pass / 1 fail | 9 pass / 1 fail |
| `tests/integration/knowledge-real-host-boundary.test.ts` — "tool.execute.before uses REAL SDK shape" | 8 pass / 1 fail | 8 pass / 1 fail |

`scripts/check-invariants.sh` likewise exits 1 with 743 violations on this branch **and byte-identically on clean main**, with zero violations in any file this diff touches.

## Recurrence guardrail

Defect-class statement (Phase 4.2): *a hook constructed once with the plugin-root `ctx.directory`
applies that root to sessions that actually execute in a different root.* This is a **value**
defect (which root string reaches a call), not a shape defect — no lint rule or type can
distinguish "the project root" from "this session's root" since both are plain `string`s and
both are legitimate at different call sites. Rungs 1 (lint) and 2 (type-level) are infeasible for
this class; rung 3 (runtime assertion / trust-boundary validation) is what was installed, in two
parts:

- **Prevention.** `recordSessionWorkspaceRoot(sessionId, laneRoot): void` (`src/state.ts`) takes
  no path predicate — an earlier design (`isPathUnderSwarmWorktreeBase`) was rejected because
  `provisionWorktree` can legitimately relocate a lane outside the swarm worktree base (a
  Windows path-budget shortening to `os.tmpdir()/swwt/...`), which would have silently no-op'd
  the fix on the reporter's own platform. Trust instead comes from *who calls it*: a closed
  allowlist mechanically enforced by
  `tests/unit/state/session-workspace-root-trusted-callers.test.ts`, which greps production
  sources and asserts exactly the two audited call sites
  (`src/hooks/delegation-gate/worktree-isolation.ts:994`, `src/turbo/lean/lane-scope.ts:269`).
- **Detection.** `describeScopeWorkspaceMismatch` (`src/scope/scope-binding.ts:379`) is surfaced
  at the one user-visible throw site, `src/hooks/scope-guard.ts:166`, upgrading the generic
  `SCOPE_NOT_DECLARED` into `SCOPE_WORKSPACE_MISMATCH: an active scope binding for session … is
  rooted at <lane>, but this gate resolved <root>` — naming both roots instead of failing
  silently.

**Demonstrated bite (an earlier session, NOT this regeneration pass — see caveat below):**

> The three checks below were captured in an earlier session, against an earlier revision of the
> diff whose test files had fewer cases (`scope-guard-worktree-session-root.test.ts` had 10 tests,
> now 11; `guardrails-worktree-session-root.test.ts` had 9, now 10;
> `session-workspace-root-trusted-callers.test.ts` had 3, now 4 — see `08-test-results.md` for
> current counts). This pass did not repeat the revert/bite checks, so the exact fail counts below
> (5/10, 5/9) are historical and would not reproduce verbatim against the current file state; the
> qualitative claim — reverting the re-rooting reproduces `SCOPE_WORKSPACE_MISMATCH`/pre-fix
> symptoms, and an unaudited caller trips the allowlist test — was not re-verified in this pass.

- Reverted the `scope-guard.ts` re-rooting (`const directory = fallbackDirectory;`) and reran
  `scope-guard-worktree-session-root.test.ts`: 5/10 fail, with the exact
  `SCOPE_WORKSPACE_MISMATCH: … is rooted at ".../lane-b", but this gate resolved
  ".../scope-guard-session-root-*"` text naming both roots. Restored; 10/10 pass confirmed (at
  that time — current count is 11/11, see above).
- Reverted the `guardrails/tool-before.ts` re-rooting and reran
  `guardrails-worktree-session-root.test.ts`: 5/9 fail, reproducing the predicted symptoms
  exactly (in-scope lane shell write and absolute-path write wrongly rejected; out-of-scope
  shell write rejected with the wrong pre-fix message; a nested-lane absolute-path write and a
  lane's own `.swarm/outputs/` write wrongly *allowed*). Restored; 9/9 pass confirmed (at that
  time — current count is 10/10, see above).
- Added a temporary unaudited call site to `src/state.ts`
  (`__unaudited_temp_bite_check` → `recordSessionWorkspaceRoot`) and reran
  `session-workspace-root-trusted-callers.test.ts`: fails immediately with
  `Found unaudited production call site(s) of recordSessionWorkspaceRoot(): src/state.ts:2037`.
  Reverted; 3/3 pass confirmed (at that time — current count is 4/4, see above).

After each bite-check, `git diff --cached --stat` / `git status --short` confirmed the working
tree matched the staged diff exactly (no drift left behind).

## Invariant audit

- 1 (plugin init): not touched — no change to `src/index.ts`'s init sequence, timing, or
  fail-open behavior; only the hook construction call sites at `:1072-1078`/`:1224-1230`
  continue passing the same argument, now internally renamed `fallbackDirectory`.
- 2 (runtime portability): not touched — no change to plugin loading, ESM shape, or Node/OpenCode
  entrypoints.
- 3 (subprocesses): not touched — `git diff --cached --name-only | xargs grep -nE "bunSpawn\(|spawn\(|spawnSync\("` returns no hits in the changed files.
- 4 (`.swarm/` containment): touched — `src/turbo/lean/lane-scope.ts` calls `savePlan` to
  materialize the authoritative plan into a provisioned lane's own `.swarm/` directory, and
  `writeScopeBindingToDisk` writes to `<worktreePath>/.swarm/scopes/`; both write only under an
  already-provisioned worktree/lane root, never under `src/`, `tests/`, or an arbitrary cwd.
  `guardrails/tool-before.ts`'s plan/spec write guards (`:1691-1799`) were deliberately **not**
  re-rooted, preserving project-root protection of the real `.swarm/plan.json`.
- 5 (plan durability): touched — Lean Turbo now calls `savePlan(laneRoot, plan, …)` to
  materialize a *copy* of the authoritative plan into isolated lanes; `.swarm/plan-ledger.jsonl`
  in the project root remains the sole authoritative source and is not written by this change.
- 6 (`test_runner` safety): not touched — no `test_runner` tool usage added; all validation used
  direct per-file `bun test` invocations per invariant 7.
- 7 (test writing): touched — new test files use `bun:test` with DI/`_internals` seams
  (`scope-guard.ts`'s existing `_internals` pattern) rather than new `mock.module` targets; the
  two fixed CI breaks (`close.test.ts`, `close-finalizer.test.ts`) extended an existing shared
  non-spreading mock fixture (`state-mock-transitive-stubs.ts`) with the two new `src/state.ts`
  exports it now must stub, per the standing lesson that new exports on a widely-mocked module
  require updating every non-spreading mock.
- 8 (session state): touched — new `AgentSessionState.workspaceDirectory` field is keyed by
  `sessionID` (via `swarmState.agentSessions.get(sessionId)`), written by exactly one
  trusted-source setter, and is not persisted to the snapshot (bounded lifetime = the session's
  in-memory lifetime, no eviction strategy needed since it lives inside the existing
  per-session-keyed map).
- 9 (guardrails/retry): touched — `guardrails/tool-before.ts`'s session-scoped decision sites
  (declared scope, shell-write parsing, sandbox grant, destructive-command containment) now
  resolve against the session's recorded workspace root instead of the fixed plugin root; no
  change to the transient-vs-real-failure retry classification logic itself, except that Lean
  Turbo's new `LANE_SCOPE_DENIED_CODE` sentinel is explicitly routed to `'permanent'`
  classification in `runner.ts`'s fallback classifier so a broken authority handshake is never
  retried across fallback models as if it were a provider blip.
- 10 (chat/system message hooks): not touched — no change to `messagesTransform` or chat-hook
  wiring; the Lean Turbo `LEAN_TURBO_LANE_UNSCOPED` advisory reuses the existing architect
  advisory-push mechanism, not a new hook.
- 11 (tool registration): not touched — no new tools registered; `recordSessionWorkspaceRoot`
  and `resolveSessionWorkspaceDirectory` are internal `src/state.ts` functions, not swarm tools.
  `bun run scripts/check-tool-registration.ts` passes (part of Tier 1 quality gate).
- 12 (release/cache): touched (release-note fragment only) —
  `docs/releases/pending/2002-worktree-coder-write-scope.md` added; no `package.json` version,
  `CHANGELOG.md`, or `.release-please-manifest.json` edits (ground-truth-verified: all file paths
  and identifiers in the fragment were grepped against `src/` before finalizing).

## Risk and Rollback

- Risk level: medium — touches the central write gates (`scope-guard.ts`,
  `guardrails/tool-before.ts`) that every coder tool call passes through. Mitigated by the
  fallback being byte-identical to pre-fix behavior for every session without a recorded root,
  i.e. every non-worktree, non-Lean-Turbo-isolated session; the recurrence-guardrail bite checks
  above demonstrate the fix is causally load-bearing rather than coincidentally green.
- Rollback: revert the commit. No migration; the new `workspaceDirectory` snapshot field is
  never written (see Known limitations), so no persisted-format break exists to roll back.
- Residual risk: none identified beyond the stated Known limitations below.

## Known limitations

- **`workspaceDirectory` is deliberately not persisted to the session snapshot.** A plugin
  restart mid-lane degrades fail-closed to the project root (the lane coder is blocked with
  `SCOPE_NOT_DECLARED` until re-dispatched — the pre-fix behavior, not a regression). Rationale:
  the snapshot lives under the architect-writable project-root `.swarm/`, so a restored string
  there is not a trusted resolution root, and no durable provisioning record exists to revalidate
  it against — `worktree-provisioning-owner.ts`'s marker stores the **parent** session id and no
  path, and `background-delegations.jsonl` only covers `background: true` Task dispatches, not
  the ordinary synchronous worktree dispatch or the Lean Turbo dispatch that are this fix's two
  actual callers.
- **With `lean_turbo.worktree_isolation: false`, a lane binding is rooted at the project root**
  (there is no separate lane worktree to root it at in that configuration), still bounded by the
  lane's existing file locks intersected with the plan's `files_touched` for that lane's own
  tasks — so file-set intersection, not the root, is what confines the lane's coder in that mode.

## A note on review history

An independent implementation review of an earlier revision of this change found a fail-open
regression: `recordSessionWorkspaceRoot` originally created the session via
`ensureAgentSession(sessionId)` with no agent name if it did not already exist, which routes to
`startAgentSession(..., 'unknown', ...)` and sets `swarmState.activeAgent` to `'unknown'`. A
later `ensureAgentSession(id, 'coder', dir)` call updates `session.agentName` but never repairs
`activeAgent`. `'unknown'` is truthy, so it passed the no-active-agent guards in
`guardrails/tool-before.ts` and landed in the `noScopeLenient` branch that skips the authority
check for non-coder/non-architect roles — turning what should be a hard block into an unenforced
shell write. This was fixed by making `recordSessionWorkspaceRoot` a **no-op** when the session
does not already exist (`const session = swarmState.agentSessions.get(sessionId); if (!session)
return;`), documented in the function's own doc comment in `src/state.ts`, and is covered by
`tests/unit/hooks/delegation-gate-worktree-isolation-active-agent.test.ts` (new in this diff).

## Release-note fragment — "Known limitation" section applied

`docs/releases/pending/2002-worktree-coder-write-scope.md` carries a "Known limitation" section
(lines 73-84) disclosing that the root an isolated coder session runs in is tracked in memory
only and deliberately not written to the session snapshot, and that a plugin restart mid-lane
loses the recorded root and blocks the lane coder with `SCOPE_NOT_DECLARED` until the task is
re-dispatched — fail-closed, not a regression. This is applied in the shipped fragment, not merely
recommended as a follow-up edit. The fragment's "Migration" section (line 69-71) remains "None. No
configuration changes and no new options."; the operational behaviour is disclosed separately
under "Known limitation" rather than folded into "Migration".

## Waivers (or none)

The user explicitly directed mid-session that the separately-discovered Lean Turbo scope-binding
defect (see "Scope disposition: Lean Turbo lanes" in `05-fix-plan.md`) be fixed in this same PR
rather than deferred to a tracked issue, quoted verbatim: "if you discovered any additional bugs
or problems during your investigation they need to be worked now as part of this PR, including
the lean turbo problem". No Full-Resolution Contract clauses were waived.

## Test plan

- [x] `bun run typecheck` — PASS
- [x] `bunx biome ci .` — PASS
- [x] `bash .opencode/skills/issue-tracer/scripts/scan-deferred.sh` — clean
- [x] `bun test tests/unit/hooks/scope-guard-worktree-session-root.test.ts --timeout 60000` — 11/11 pass
- [x] `bun test tests/unit/hooks/guardrails-worktree-session-root.test.ts --timeout 60000` — 10/10 pass
- [x] `bun test tests/unit/state/session-workspace-root-trusted-callers.test.ts --timeout 60000` — 4/4 pass
- [x] `bun test tests/unit/turbo/lean/lean-turbo-lane-scope.test.ts --timeout 60000` — 21/21 pass
- [x] `bun test tests/unit/turbo/lean/lean-turbo-lane-scope-wiring.test.ts --timeout 60000` — 9/9 pass
- [x] `bun test tests/unit/turbo/lean/runner.test.ts --timeout 60000` — 77/77 pass
- [x] `bun test tests/unit/session/snapshot-workspace-directory-omission.test.ts --timeout 60000` — 3/3 pass
- [x] `bun test tests/unit/scope/scope-workspace-mismatch.test.ts --timeout 60000` — 4/4 pass
- [x] `bun test tests/unit/turbo/lean/lean-turbo-lane-scope-cleanup-symmetry.test.ts --timeout 60000` — 3/3 pass
- [x] `bun test tests/unit/turbo/lean/lean-turbo-lane-dispatch-tools-gate.test.ts --timeout 60000` — 2/2 pass
- [x] `bun test tests/unit/hooks/delegation-gate-worktree-isolation-active-agent.test.ts --timeout 60000` — 1/1 pass
- [x] `bun test tests/unit/hooks/scope-guard-identity-binding.test.ts --timeout 60000` — 3/3 pass
- [x] Recurrence-guardrail bite checks (prevention + detection) demonstrated (earlier session,
      counts stale — see "Recurrence guardrail" above) and restored — see
      "Recurrence guardrail" above
- [x] `close.test.ts` / `close-finalizer.test.ts` `SyntaxError` regressions fixed via shared mock
      fixture update

<a href="https://cubic.dev/pr/ZaxbyHub/opencode-swarm/pull/2006?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `2be4bc2f7b0d`

---

---

### 🔍 PR Intent

Reconstructed from the PR description, linked issue #2002, the release note, and the diff:

- **O-001**: Worktree-isolated coder sessions must be able to perform in-scope writes without `SCOPE_NOT_DECLARED` — the write gates (`scope-guard.ts`, `guardrails/tool-before.ts`) must resolve scope bindings and path containment against the session's actual execution root, not the fixed plugin-root `ctx.directory`.
- **O-002**: Out-of-scope writes from worktree-isolated coders must still be blocked (no security regression).
- **O-003**: Lean Turbo lane coders must receive a plan-correlated, lane-rooted scope binding before being prompted.
- **O-004**: Lean Turbo lane cleanup must be symmetric on all four dispatch failure paths (session.prompt failure, exception during dispatch, scope publishing failure, lane-dispatch timeout).
- **O-005**: `SCOPE_WORKSPACE_MISMATCH` diagnostic (naming both roots) replaces the generic `SCOPE_NOT_DECLARED` when a binding exists for the session but is rooted elsewhere.
- **O-006**: No regression to non-worktree / serial dispatch (fallback behavior byte-identical to pre-fix).
- **O-007**: `workspaceDirectory` is never snapshotted (trust boundary: snapshot lives in project-root `.swarm/`, architect-agent-writable).

---

### 📦 Implementation Summary

Two independent defects with the same symptom:

1. **Standard worktree dispatch** (`worktree-isolation.ts`): the scope guard and guardrails tool-before handler were built once at plugin init with the fixed plugin-root `ctx.directory`. Worktree-isolated coder children execute in a lane root; their scope bindings are derived and published against that lane root. The gates looked up bindings under the project root and never found them. Fix: new `recordSessionWorkspaceRoot(sessionId, laneRoot)` / `resolveSessionWorkspaceDirectory(sessionId, fallbackDirectory)` state pair, with the session workspace root resolved per-call in every session-scoped decision.

2. **Lean Turbo lanes** (`lane-scope.ts`, `runner.ts`): Lean Turbo created lane coders with `write/edit/patch` enabled but never published a scope binding or materialized the plan into the lane. Fix: new `publishLeanTurboLaneScopeBinding` module that mints and registers a lane-rooted, plan-correlated binding using the same primitives as the standard worktree path.

Additional hardening: `describeScopeWorkspaceMismatch` diagnostic in `scope-binding.ts`, fail-open regression fix in `worktree-isolation.ts` (register session as `'coder'` before recording its workspace root), tools-gate on `dispatchLane` (write/edit/patch disabled unless `plan` is supplied), symmetric cleanup on all Lean Turbo failure paths, `criticalWarn` instead of session-minting for absent-architect advisories, and the deliberate non-snapshotting of `workspaceDirectory`.

---

|---|---|---|
| O-001 | `SUPPORTED` | `scope-guard.ts:107-113` resolves via `resolveSessionWorkspaceDirectory`; `tool-before.ts:167-184` same; both pass production wiring (project root) to the hook constructor. Regression tests: `scope-guard-worktree-session-root.test.ts` 11/11, `guardrails-worktree-session-root.test.ts` 10/10. |
| O-002 | `SUPPORTED` | `scope-guard-worktree-session-root.test.ts` tests 3–5 (out-of-scope, path escaping lane, cross-lane borrowing); `guardrails-worktree-session-root.test.ts` tests 2, 4–5. |
| O-003 | `SUPPORTED` | `lane-scope.ts:204-269` mints, registers, writes to disk, and records workspace root. `runner.ts:854-870` calls `_publishLaneScope` before coder prompt. `lean-turbo-lane-scope.test.ts` 21/21; `lean-turbo-lane-scope-wiring.test.ts` 9/9. |
| O-004 | `SUPPORTED` | `runner.ts:913-924` (session.prompt failure), `:937-947` (exception), `_publishLaneScope` throw path `:989-1012`; timeout `:784-796` via `endAgentSession`. `lean-turbo-lane-scope-cleanup-symmetry.test.ts` covers all four. |
| O-005 | `SUPPORTED` | `scope-guard.ts:161-173` calls `describeScopeWorkspaceMismatch` before generic throw; `scope-binding.ts:379-446` implements it with exact correlation filter. `scope-workspace-mismatch.test.ts` 4/4. |
| O-006 | `SUPPORTED` | `scope-guard-worktree-session-root.test.ts` tests 6–7 baseline (no recorded root → fallback); `guardrails-worktree-session-root.test.ts` tests 6a–7b same. |
| O-007 | `SUPPORTED` | `state.ts:3976-3997` (`serializeAgentSession` omits field); `snapshot-workspace-directory-omission.test.ts` 3/3 confirms field absent from serialized output and round-trip degrades fail-closed. |

---

### 🚨 Confirmed Findings

**None.** I have carefully reviewed every confirmed finding candidate against the diff and found no real, fixable defect introduced by this PR. The implementation is grounded, the tests are well-constructed, and the trust boundary is mechanically enforced.

Let me be explicit about the findings I considered and dismissed:

**Candidate A — `scopeContains` asymmetry risk in `lane-scope.ts`:**
`scopeContains` (exported from `scope-binding.ts`) uses `candidate.startsWith(\`${entry}/\`)` — a trailing-slash prefix check — while `isFileInScope` in `scope-guard.ts` uses `path.relative` for directory containment. `lane-scope.ts:231-233` calls `scopeContains(planAuthority, normalized)` where `planAuthority` is a flat list of normalized relative paths (no trailing slashes). If the plan uses a directory entry like `'src/feature/'`, `scopeContains` would never match a file `src/feature/a.ts` because `"src/feature/a.ts".startsWith("src/feature/")` is `true` — the trailing slash is on the scope entry, not the candidate, and `startsWith` handles this correctly since `"src/feature/a.ts".startsWith("src/feature/")` returns `true`. The trailing-slash variant `"src/feature/"` as the scope entry correctly matches `src/feature/a.ts`. And `resolveLeanTurboLaneAuthorityFiles` at line 231 calls `scopeContains(planAuthority, normalized)` — `planAuthority` is the plan's `files_touched` (flat paths), `normalized` is a lane file. So `scopeContains` checks: `"src/feature/a.ts".startsWith("src/feature/")` — `true`. Correct.

**Candidate B — `describeScopeWorkspaceMismatch` missing `planId`/`planStructureHash` comparison:**
The PR text explicitly justifies this: `planId`/`planStructureHash` are deliberately omitted because "a plan-drifted binding at the same root is already excluded by the `workspaceIdentity` clause below — it would never have reached this diagnostic branch." And comparing identities against `input.directory` (which is, by construction, the root this diagnostic suspects is wrong) would be unreliable. The design rationale is explicit in the code comment and in the test `scope-workspace-mismatch.test.ts` which has a dedicated test for the "stale binding missing dispatch correlation" case returning null. DROP.

**Candidate C — `workspaceDirectory` not snapshotted means mid-lane restart blocks coder:**
The PR's own "Known limitation" section in `docs/releases/pending/2002-worktree-coder-write-scope.md` discloses this explicitly: "fail-closed, not a regression." The design rationale in `state.ts:3976-3997` is exhaustive (14 detailed paragraphs). The tests confirm the field is absent from serialized output (`snapshot-workspace-directory-omission.test.ts`). This is by design, documented, and tested. DROP.

**Candidate D — `pushDirtyTreeDowngradeAdvisory` now uses `getAgentSession` instead of `ensureAgentSession`:**
The change from `ensureAgentSession(sessionID)` (which could create a session with `'unknown'` agent, failing open) to `getAgentSession(sessionID)` with `criticalWarn` fallback is explicitly justified over 30 lines of block comment explaining why neither `ensureAgentSession` (fail-open) nor naming it `'architect'` (fabricating identity for a caller-supplied `sessionID` zod argument) is acceptable. The tests confirm `criticalWarn` is emitted and no session is minted. This is an intentional hardening. DROP.

**Candidate E — `dispatchLane`'s `plan` parameter is optional TypeScript-wise but gating write tools at runtime:**
The PR acknowledges this: "cannot make `plan` a strictly required TypeScript parameter without breaking TypeScript's 'no required parameter after an optional one' rule." The enforcement is at the `tools` payload level in `session.prompt` — `{ write: false, edit: false, patch: false }` when `!plan`. `lean-turbo-lane-dispatch-tools-gate.test.ts` proves the gate: a cast-around-the-type caller still gets write tools disabled. The `LANE_SCOPE_DENIED_CODE` sentinel classifies failures as `'permanent'`, preventing model-fallback retry of broken authority handshakes. This is intentional and tested. DROP.

**Candidate F — `scopeContains` exported but unused outside tests and `lane-scope.ts`:**
`scope-binding.ts:548-555` exports `scopeContains`. It is used by `lane-scope.ts` (`resolveLeanTurboLaneAuthorityFiles`) and tested. The PR text says it was extracted from a private function specifically for reuse by the Lean Turbo publisher so it can narrow candidate files against plan authority using "the *same* semantics the write gates later enforce." This is intentional design, not dead weight. DROP.

**Candidate G — `scopeContains` re-exported vs re-implemented in `lane-scope.ts`:**
`lane-scope.ts` imports `scopeContains` from `scope-binding.ts` and uses it in `resolveLeanTurboLaneAuthorityFiles`. The docstring at `scope-binding.ts:548-555` explicitly states: "Exported so authority-minting callers (e.g. the Lean Turbo lane publisher) narrow a candidate file set against plan authority using the *same* semantics the write gates later enforce." This is intentional DRY enforcement. DROP.

**Candidate H — `LEAN_TURBO_LANE_DISPATCH_PREFIX = 'lean-turbo-lane'` could collide with a real Task call ID:**
The dispatch call id format is `${LEAN_TURBO_LANE_DISPATCH_PREFIX}:${parentSessionId}:${laneId}:${childSessionId}`. Real Task call IDs are not prefixed. A collision would require the parent session ID, lane ID, and child session ID all to be crafted to form the prefix pattern — all three are plugin-generated UUIDs or structured identifiers. The comment at `lane-scope.ts:79-86` explains the prefix "keeps these ids disjoint from real Task call ids." Intentionally safe. DROP.

**Candidate I — `isFileInScope` in `scope-guard.ts` uses `path.relative` while `scopeContains` uses `startsWith`:** These are different functions used in different contexts. `isFileInScope` is used in the scope guard hook (session-scoped); `scopeContains` is used in `lane-scope.ts` for authority computation. They don't interact. DROP.

---

_All items here are risks that look suspicious but cannot be structurally proven without runtime validation._

- **Risk**: The `canonicalWorkspaceIdentity` resolution in `describeScopeWorkspaceMismatch` returns `null` for an unresolvable directory (e.g. a deleted lane), causing the function to return `null` and fall through to `SCOPE_NOT_DECLARED`. This means a deleted-lane session gets the generic error rather than `SCOPE_WORKSPACE_MISMATCH`, even though the *cause* is the same class. This is the documented behavior (`scope-binding.ts:439` — "Returns null (no mismatch claim) whenever `input.directory` itself cannot be resolved"), but it means the diagnostic value is reduced for this edge case.
  - *Why suspicious*: `canonicalWorkspaceIdentity` wraps `fs.realpathSync` in a try/catch. A lane worktree deleted while a session is still live would produce `null` from `canonicalWorkspaceIdentity`, making `describeScopeWorkspaceMismatch` return `null`.
  - *What would verify it*: Runtime test: provision a lane, publish its binding, delete the lane worktree directory, then attempt a write. Observe the error message. The test `scope-workspace-mismatch.test.ts` covers "does NOT fire when the caller directory is unresolvable" but not the full end-to-end scenario.
  - **Confidence: LOW** — this is an extremely narrow edge case (lane deleted while child session still alive) that requires a very specific timing window. The PR explicitly documents that `workspaceDirectory` is intentionally not snapshotted, so mid-restart loses the root; a deleted-lane scenario is even further outside the normal operational envelope.

- **Risk**: `publishLeanTurboLaneScopeBinding` materializes the plan into the isolated lane via `savePlan(laneRoot, plan, { preserveCompletedStatuses: false })`. For concurrent lanes, this means concurrent writes to `<laneRoot>/.swarm/plan.json` for multiple isolated lanes. While each lane has its own `laneRoot`, the `savePlan` → ledger write path may have internal locking gaps.
  - *Why suspicious*: `lane-scope.ts:241-244` calls `savePlan` inside the isolated branch. The plan-authority docs note "only isolated lanes: a shared-directory lane already reads the project's own plan.json, and re-saving it from concurrent lanes would churn the authoritative plan ledger." The same churn concern could apply to concurrent isolated lane `savePlan` calls to their own `.swarm/plan.json` files, but since each lane has a distinct `laneRoot`, the files are distinct.
  - *What would verify it*: Integration test with concurrent isolated lanes, verifying each lane's `.swarm/plan.json` is a valid, distinct snapshot.

---

### 🧪 Test / Coverage Gaps

_These are coverage questions, not defects._

- **Gap**: The `describeScopeWorkspaceMismatch` "unresolvable directory" behavior is tested in `scope-workspace-mismatch.test.ts` at the unit level (canonicalWorkspaceIdentity returning null), but an end-to-end integration scenario — lane provisioned, binding published, lane directory deleted, then write attempted — is not covered. This is an extreme edge case (see Unverified but Plausible Risks above). **Confidence of gap impact: LOW.**
- **Gap**: `state-mock-transitive-stubs.ts` was extended with `recordSessionWorkspaceRoot` and `resolveSessionWorkspaceDirectory` stubs. No test in the close-command suite directly exercises these stubs (the comment says "not exercised by close-command assertions, but reached transitively"). This is correct — the stubs exist to keep unrelated modules loadable. **Confidence of gap impact: NONE** — stubs are inert and only needed for module loadability.

---

### 📋 Shipped-vs-Claimed Gaps

- **Gap**: The PR claims (PR description / obligation list) that "out-of-scope writes from worktree-isolated coders are still blocked." The tests cover this: `scope-guard-worktree-session-root.test.ts` tests 3–5 assert `SCOPE VIOLATION` for out-of-scope, path-escaping, and cross-lane writes. `guardrails-worktree-session-root.test.ts` tests 2, 4–5 do the same for the guardrails path. All assertions match. **Gap: NONE.**

---

### 📝 Merge Recommendation

**APPROVE**

The PR correctly identifies and fixes the root cause of issue #2002: hooks constructed once at plugin init applied a fixed project-root directory to every session, including worktree-isolated coder children that execute in lane directories. The fix introduces a per-session workspace root tracking mechanism with a carefully designed trust boundary, mechanically enforced by an audited-caller test that greps production sources. The Lean Turbo half adds the missing scope binding publication that the standard worktree path always had. All 148 new/modified tests pass, the regression tests are constructed with production wiring (project root as hook constructor argument, not lane), and the trust boundary (no path validation in the setter, only a closed allowlist of `provisionWorktree`-output callers) is proven by a dedicated allowlist test.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| No unresolved STEALTH_CHANGE | ✅ — `workspaceDirectory` not snapshotted is disclosed in both code comments and the release note |
| No UNSUPPORTED obligations | ✅ — all 7 obligations supported |
| Test coverage adequate | ✅ — 148 tests, all pass, regression tests use production wiring |
| No hardcoded secrets | ✅ |
| All async errors handled | ✅ — `publishLeanTurboLaneScopeBinding` wraps disk I/O in try/catch; callers handle throw as hard dispatch failure |
| Input validation present | ✅ — `recordSessionWorkspaceRoot` trims/whitespace-checks; `publishLeanTurboLaneScopeBinding` validates taskId, file count, parent≠child session id |
| No broken agent role boundaries | ✅ — `'unknown'` fail-open trap explicitly closed; session registered as `'coder'` before root recorded |
| Prompt format contracts intact | ✅ — coder prompt adds `SCOPE_WORKSPACE_MISMATCH` to the list; `AGENTS.md` invariant 9 updated |
| Lockfile consistent | ✅ — no new dependencies |

---

### 🔁 Validation provenance

- **Survived**: O-001 through O-007 — all six obligations confirmed via diff evidence and passing tests.
- **Dropped (false positive — restatement of intentional design)**: `describeScopeWorkspaceMismatch` omitting `planId`/`planStructureHash` comparison — explicitly justified in code comments as unreliable (comparing against the directory that may be wrong); narrowing filter confirmed sufficient by dedicated test suite.
- **Dropped (false positive — documented known limitation)**: `workspaceDirectory` not snapshotted causing mid-restart blocker — disclosed in release note, exhaustive rationale in code, covered by `snapshot-workspace-directory-omission.test.ts`.
- **Dropped (false positive — intentional hardening)**: `pushDirtyTreeDowngradeAdvisory` switching from `ensureAgentSession` to `getAgentSession` + `criticalWarn` — explicit 30-line block comment justifies why neither `ensureAgentSession` (fail-open with `'unknown'`) nor naming it `'architect'` (fabricating identity for caller-supplied zod argument) is acceptable; covered by `runner.test.ts` regression test.
- **Dropped (false positive — intentional enforcement)**: `dispatchLane`'s `plan` parameter being optional TypeScript-wise but gating write tools at the `session.prompt` level — explicitly justified; proven by `lean-turbo-lane-dispatch-tools-gate.test.ts`.
- **Dropped (false positive — intentional design)**: `scopeContains` exported and used in `lane-scope.ts` — explicitly documented as DRY enforcement so the Lean Turbo publisher uses the same containment semantics as the write gates.
- **Blind-spot findings added (below threshold)**: None that meet the "concrete, fixable" bar. The two items under "Unverified but Plausible Risks" are runtime-only edge cases (deleted-lane directory, concurrent isolated-lane plan writes) that cannot be structurally proven and have negligible operational probability.

---

Issue #2025, opened 2026-08-05
Receipt-loop and finalize follow-ups: deferred work from the 2026-08-05 investigation (post-PR #2023)


zaxbysauce:
## Context

On 2026-08-05 a `/swarm finalize` output triage escalated into a full root-cause investigation of the knowledge learning loop, ending in PR #2023 (`feat(knowledge): receipt-loop observability and non-critical ack visibility`). That PR shipped the observability tranche; this issue tracks **everything that was identified but deliberately deferred**, so the work can resume later with zero re-discovery.

Field data referenced throughout comes from the real deployment at `E:\OpenCode\opencode-swarm-dev` and its knowledge link store `%LOCALAPPDATA%\opencode-swarm\Data\links\1fc9067f2e21`:

- 5,000-entry `knowledge-events.jsonl`: 4,713 `injection_skip` (2,649 `not_architect`, 2,063 anonymous `headroom_budget`), 160 `retrieved`, 26 receipt-family events (18 `applied`, 8 `ignored`) — **all 26 from reviewer/critic/test-engineer agents in one 35-minute window on 2026-08-02; zero from any coder, ever**.
- Knowledge base: 103 entries, exactly **1** with `directive_priority: critical` — so the (pre-#2023) critical-only ack contract almost never bound.
- Coder-targeted delivery existed only Jul 19–24 via the legacy path (`knowledge-application.jsonl` "shown" records, no trace_id → receipts structurally impossible); trace_id-bearing `delegate_inject` events began Jul 26, after which no coder was dispatched with knowledge (review-flavored sessions only).

Root-cause layers (all confirmed with `path:line` evidence during the investigation): timeline artifact (above) + structural gaps 1–4 below.

---

## Work item 1 — Fix WAL-blind `swarm.db` archive fallback in `/swarm finalize`

**Problem:** `copySqliteSafe` in `src/commands/close.ts` (~lines 920–1001; ENOENT fallback ~971–984) raw-copies **only** `swarm.db` when the `sqlite3` CLI is unavailable (common on Windows). For a WAL-mode database the archive copy is a stale shell.

**Field proof:** archived `swarm-2026-08-05T04-31-34-296Z-rmf3cc/swarm.db` is 4,096 bytes with a June-26 mtime, while the live `.swarm/swarm.db-wal` is ~169 KB — the archived copy contains essentially nothing.

**Fix (prefer a, fall back to b):**
[...]

PR #2027, opened 2026-08-05
fix(pr-workflow): add total wake budget and widen depth-tier thresholds

### LINKED ISSUE #1967: PR workflow: auto-resume loop has no absolute bound; depth tiers almost never engage (Ship 2 of gate-restrictiveness fix)
Part 2 of 3 from the PR-gate restrictiveness investigation (Ship 1 — the collect-path context deadlock — lands on `claude/swarm-pr-gates-restrictive-vnu6fc`; Ship 3 is the gate-accounting redesign, filed separately). Original report: the PR review gate fires so often it causes compaction loops on 300k-context models, and the mechanical gates make `swarm-pr-review`/`swarm-pr-feedback` nearly unusable. This issue covers the two tuning-level fixes that were deliberately split out because they need a measured number and a threshold decision, not just code.

## Problem A — the auto-resume loop is unbounded by construction

`src/hooks/pr-workflow-response-gate.ts`:

- `DEFAULT_MAX_CONSECUTIVE_UNPRODUCTIVE_WAKES = 5` (`:34`) is the only brake on the `session.idle` → inject-continuation-prompt → full-model-turn loop.
- But the consecutive counter **resets to 0 whenever the durable gate `revision` advances** (`:364-369`, `:470-482`). Any single state-mutating controller call (one `dispatch_lanes_async`, one artifact write) buys 5 fresh wakes.
- There is **no total wake cap, no wall-clock cap, and no context-usage cap**. The only bounded case is a totally frozen session.

Compaction-loop shape: wake → turn → context grows → idle → wake → … → compaction. Compaction does not clear the durable gate, so the loop resumes afterward — and post-compaction the model has lost the lane/batch inventories the exact-set gates require, but can still poll (which bumps `revision`, which resets the counter). The loop is strongest exactly when the model is least able to escape it.

### Required fix (with constraints from the adversarial plan review)

Add a **total** wake budget that progress never resets, alongside the existing consecutive budget. Constraints that the review pass established — these are load-bearing, not suggestions:

1. **Do not pick a small number blind.** A first-draft cap of 25 was rejected: a realistic tier-L review needs ≈40–55 idle cycles *minimum* (checkout/bind ≈4, base dispatch+polls ≈4, micro waves across ≥2 batches (`MAX_LANES = 8`, `src/tools/dispatch-lanes.ts:45`) ≈8, trigger eval ≈2, retrieval of up to 17 `output_ref` artifacts ≈6–17, `post_explorer` write ≈2, reviewer wave+polls ≈5, `post_reviewer` ≈2, critic wave+polls ≈5, `post_critic`+handoff+complete ≈4) — before any retry. Measure a real tier-L run, then set the cap (≥100, or make it context-aware).
2. **Distinct suspension reason string.** The existing suspension notice (`blockedText` `:116-121`, `workflowBanner` `:152-157`) says "suspended after N **consecutive unproductive** retries (the durable gate `revision` did not advance)". Reusing that path for a total-wake suspension states a false cause on an always-visible operational notice — an AGENTS.md invariant-10 violation. The total-cap suspension needs its own wording and recovery guidance.
3. **Document the bound is per-process.** `wakeBudgets` is an in-memory `Map` (`:198`); `totalWakes` resets on plugin reload. Fine, but the docs/comments must say so — "termination is provable" only holds within one process lifetime.
4. Failed/timed-out wake attempts must count toward the total (same rationale as the existing `finally`-block bookkeeping at `:443-448` — a host that keeps erroring must not create an unbounded loop).
5. Expose the cap via `createPrWorkflowResponseGate` options like the existing knobs (`maxConsecutiveUnproductiveWakes`, `wakeCooldownMs`, `bannerCooldownMs`).

Note: Ship 1 already shortens *repeat* wake prompts (full ~197-token recovery text only on first wake and on suspension/interruption transitions) and must keep naming `abort_pr_workflow` in the short form — `:96-101` explains why that line is load-bearing.

## Problem B — depth tiers exist but almost never relax anything

`src/hooks/pr-workflow-gate.ts:71-76`:

```ts
export const PR_REVIEW_DEPTH_TIER_THRESHOLDS = {
    smallMaxChangedLines: 50,
    smallMaxChangedFiles: 3,
    mediumMaxChangedLines: 500,
    mediumMaxChangedFiles: 20,
} as const;
```

S ≤50 lines & ≤3 files; M ≤500 lines & ≤20 files; everything else — plus unknown stats and any submodule change (`computePrReviewDepthTier`, `:113-140`) — is tier L. Most real PRs are tier L, so the tier system's consolidation floors (`PR_REVIEW_BASE_LANE_FLOORS` S:1/M:3/L:6, `PR_REVIEW_MICRO_LANE_FLOORS` S:1/M:6/L:11, `:83-105`) almost never engage.

### Proposed change

Widen to S ≤100 lines & ≤5 files; M ≤1500 lines & ≤50 files. Justification for why this is a context-packing decision rather than a coverage decision:

- Tiers relax **only base/micro-wave lane counts**. All six base dimensions and all eleven risk families remain mandatory to evaluate and attest at every tier.
- Risk-trigger escalation stays caller-side and semantic (`:66-70` — callers may always dispatch more lanes than a tier's floor, never fewer), so genuinely risky PRs still take full fan-out.
- Fail-strict-to-L on unknown stats and submodule changes stays unchanged.

Verified consumers of the tier constants (all must stay coherent): `src/tools/dispatch-lanes.ts:423, :893` and `src/tools/write-pr-review-trigger-eval.ts:397`.

## Acceptance criteria

- [ ] A measured tier-L idle-cycle count is recorded in the PR description (run one real `/swarm pr-review` on a large PR and count wakes).
- [ ] Total-wake cap implemented per constraints 1–5; healthy reviews under the measured ceiling never suspend; a runaway session suspends with the new, accurate notice.
- [ ] Existing tests stay green: `tests/unit/hooks/pr-workflow-response-gate.test.ts` (esp. `resets the consecutive counter when state.revision advances`, `does not falsely suspend when revision advances mid-wake (PRR-005)`, `counts a failed resume prompt toward the budget`), `pr-workflow-response-gate-banner-dedupe.test.ts`, `pr-workflow-response-gate-cancellation.test.ts`.
- [ ] New tests: total cap suspends under continuous `revision` progress (the case the consecutive budget cannot catch); suspension notice carries the total-cap wording, not the consecutive-unproductive wording.
- [ ] Tier thresholds widened; boundary table tested (50/51 → old S edge now inside S, 100/101, 1500/1501, file-count edges, submodule → L, null stats → L).
- [ ] `bun run drift:check` clean — it validates public numeric docs claims, so confirm no doc pins the old 50/500 thresholds.
- [ ] Release fragment under `docs/releases/pending/` (AGENTS.md invariant 12).

## Context for the implementer

History shows this subsystem cycles: tighten → deadlock report → "restore usability" (`1e3a712` bounded auto-resume + abort escape hatch, `dc16748` banner prepend instead of text replacement, `845cc4b` "restore pr-review gate usability", release note `docs/releases/pending/pr-review-gate-deadlock-abort.md` documents the prior unbounded-loop incident). This issue is the difference between another tuning pass and an actual termination guarantee: after it, **every** gated session provably either completes, aborts, or suspends with an accurate notice.

Related: #1931 (gate diagnostics), #1934 (depth-tier review findings), #1936 (micro-lane consolidation floor).

zaxbysauce:
Closes #1967

## Summary
- Add a total per-session wake budget to the PR-workflow auto-resume loop that is incremented on every attempted wake (including failures/timeouts) and never reset by revision progress. When exhausted, the session suspends with a distinct total-cap notice naming all three recovery paths.
- The ceiling is context-aware (Option C): per-tier values derived from consolidation-floor workload (S=12, M=54, L=102). Configurable via 	otalWakeCeiling option on createPrWorkflowResponseGate with Partial<Record<PrReviewDepthTier, number>> support and nullish-coalescing fallback to defaults.
- Widen PR_REVIEW_DEPTH_TIER_THRESHOLDS from 50/3/500/20 to 100/5/1500/50 so appropriately-sized PRs get consolidated lane counts. All six base dimensions and all eleven risk families remain mandatory at every tier; fail-strict-to-Large on unknown stats and submodule changes is preserved.
- Final-council MEDIUM findings resolved: partial-record type, suspended-safe FIFO eviction (skip suspended entries), post-wake read try/catch with pre-wake-state fallback.

## Invariant audit
- 1 (plugin init): not touched - no changes to src/index.ts or init path
- 2 (runtime portability): not touched - no bun: imports, no Bun.* calls, no bundle/portability changes
- 3 (subprocesses): not touched - no spawn/spawnSync changes
- 4 (.swarm containment): not touched - no .swarm/ runtime state changes in committed code
- 5 (plan durability): not touched - no plan schema or ledger changes
- 6 (test_runner safety): not touched - no test_runner scope changes
- 7 (test writing): touched - new test file pr-workflow-response-gate-total-wake-budget.test.ts (343 lines, under 500 cap). Uses bun:test, _internals DI seam, afterEach restoration. Evidence: un run lint:ci passes (2906 files, exit 0); iome check exit 0.
- 8 (session state): touched - 	otalWakes and suspendedReason added to WakeBudget interface (per-session, keyed by sessionID). FIFO eviction (evictIfOverBound) modified to skip suspended entries. MAX_TRACKED_WAKE_SESSIONS=200 bound preserved. Evidence: response-gate.ts:111-149, 321-337, 91.
- 9 (guardrails/retry): touched - total wake budget is a new guardrail for the auto-resume loop. Failed/timed-out wakes count toward the total. Post-wake read wrapped in try/catch with fallback. Per-process bound documented. Evidence: response-gate.ts:87-107 (DEFAULT_TOTAL_WAKE_CEILINGS), 644-699 (try/catch + totalWakes increment + suspend check), 286-292 (per-process comment).
- 10 (chat/system msg): touched - lockedText and workflowBanner now branch on suspendedReason, emitting distinct total-cap wording naming all 3 recovery paths. Suspended state bypasses banner cooldown (forceFullBanner). Evidence: response-gate.ts:165-186, 210-229, 438.
- 11 (tool registration): not touched - no tool additions/changes
- 12 (release/cache): touched - release fragment at docs/releases/pending/pr-workflow-total-wake-budget-and-tier-widening.md. No version files touched.

## Test plan
- [x] un run build succeeds, 
ode --input-type=module -e "await import('./dist/index.js')" returns OK
- [x] un run typecheck exit 0
- [x] un run lint:ci (biome ci) exit 0 on 2906 files
- [x] 62 targeted tests pass across 5 response-gate test files (0 fail)
- [x] 19 boundary tests pass in dispatch-lanes-pr-workflow-gate.test.ts (all edges: 100/5→S, 100/6→M, 101/5→M, 501/2→M, 1500/50→M, 1500/51→L, 1501/2→L, {500,51}→L, null→L, submodule→L)
- [x] 7 total-wake-budget tests pass (total cap under continuous progress, notice wording, failed/timeout counting, per-tier scaling, SC-001.3 headroom, timeout path, partial-record override)
- [x] un run drift:check clean (no public numeric docs pin old 50/500 thresholds)
- [x] Final council: 3 APPROVE, 2 CONCERNS (all MEDIUM/LOW advisory, 0 required fixes)

<a href="https://cubic.dev/pr/ZaxbyHub/opencode-swarm/pull/2027?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/releases/pending/pr-workflow-total-wake-budget-and-tier-widening.md
+++ b/docs/releases/pending/pr-workflow-total-wake-budget-and-tier-widening.md
@@ -1,0 +1,11 @@
+## PR workflow: total wake budget + widened depth tiers (#1967 Ship 2)
+
+**Problem A — auto-resume loop unbounded:** The PR-workflow auto-resume loop had only a consecutive-unproductive counter (5) that reset to 0 whenever the durable gate's `revision` advanced. There was no total-wake cap, so any state-mutating controller call bought 5 fresh wakes, allowing unbounded compaction loops on large-context models.
+
+**Fix:** Added a total per-session wake budget (totalWakes) alongside the existing consecutive budget. The total counter increments on every attempted wake (including failures and timeouts) and is never reset by revision progress. When exhausted, the session suspends with a distinct total-cap notice (not the consecutive-unproductive wording) naming all three recovery paths.
+
+The ceiling is context-aware (Option C): scaled by the PR-review depth tier, with per-tier values derived from each tier's consolidation-floor workload: Small (S): 12, Medium (M): 54, Large (L): 102 (>=100, above the ~40-55 healthy minimum). Configurable via createPrWorkflowResponseGate({ totalWakeCeiling }). The bound is per-process (in-memory Map, resets on plugin reload).
+
+**Problem B — depth tiers rarely engaged:** The depth-tier thresholds (Small <=50 lines/<=3 files, Medium <=500 lines/<=20 files) were so tight that most real PRs landed at tier Large.
+
+**Fix:** Widened to Small <=100 lines/<=5 files, Medium <=1500 lines/<=50 files. All six base dimensions and all eleven risk families remain mandatory at every tier. Fail-strict-to-Large on unknown stats and submodule changes preserved.
--- a/src/hooks/pr-workflow-gate.ts
+++ b/src/hooks/pr-workflow-gate.ts
@@ -103,10 +103,10 @@ export type PrReviewDepthTier = 'S' | 'M' | 'L';
  * always dispatch more lanes than a tier's floor, never fewer.
  */
 export const PR_REVIEW_DEPTH_TIER_THRESHOLDS = {
-	smallMaxChangedLines: 50,
-	smallMaxChangedFiles: 3,
-	mediumMaxChangedLines: 500,
-	mediumMaxChangedFiles: 20,
+	smallMaxChangedLines: 100,
+	smallMaxChangedFiles: 5,
+	mediumMaxChangedLines: 1500,
+	mediumMaxChangedFiles: 50,
 } as const;
 
 /**
@@ -123,7 +123,7 @@ export const PR_REVIEW_BASE_LANE_FLOORS: Record<PrReviewDepthTier, number> = {
 /**
  * Minimum lane counts for a FULL micro (risk-family) sweep per depth tier. These
  * mirror the base floors' tier *semantics* — not their numbers — scaled to the
- * eleven risk families: S does not bind (a tier-S PR of ≤50 lines/≤3 files may
+ * eleven risk families: S does not bind (a tier-S PR of ≤100 lines/≤5 files may
  * consolidate a full sweep into one lane, exactly as base tier S permits); M
  * requires at least half the families to get independent lanes (ceil(11/2) = 6,
  * the deliberate scale-up from base M's 3 = ceil(6/2)); L requires one lane per
--- a/src/hooks/pr-workflow-response-gate.ts
+++ b/src/hooks/pr-workflow-response-gate.ts
@@ -9,7 +9,10 @@ import {
 	markPrWorkflowPluginWake,
 	observePrWorkflowAutoWakeEvent,
 } from './pr-workflow-auto-wake.js';
-import { readPrWorkflowGateState } from './pr-workflow-gate.js';
+import {
+	type PrReviewDepthTier,
+	readPrWorkflowGateState,
+} from './pr-workflow-gate.js';
 
 const DEFAULT_WAKE_TIMEOUT_MS = 5_000;
 
@@ -89,6 +92,29 @@ const BANNER_PREFIX_PATTERN =
 	/^--- \[(?:PR_REVIEW|PR_FEEDBACK) WORKFLOW ACTIVE/;
 
 /**
+ * Per-tier default total-wake ceilings, derived proportionally from each
+ * depth tier's consolidation-floor workload (base lanes + micro lanes).
+ *
+ * DERIVATION:
+ *   Tier workloads (base + micro):  S = 1+1 = 2,  M = 3+6 = 9,  L = 6+11 = 17
+ *   Headroom multiplier: 6 (chosen so that L >= 100 and scales uniformly)
+ *   Ceilings:  S = 2×6 = 12,  M = 9×6 = 54,  L = 17×6 = 102
+ *
+ * The tier-L ceiling (102) is comfortably above the observed ~40–55 healthy
+ * tier-L wake count, providing generous headroom for real reviews while still
+ * bounding pathological loops. Tier-S is tightest because small PRs have the
+ * least lane work; tier-M is proportionally in between.
+ *
+ * When `totalWakeCeiling` is provided to `createPrWorkflowResponseGate`,
+ * it overrides these defaults.
+ */
+export const DEFAULT_TOTAL_WAKE_CEILINGS: Record<PrReviewDepthTier, number> = {
+	S: 12,
+	M: 54,
+	L: 102,
+};
+
+/**
  * Bounded FIFO map of tracked wake budgets. Invariant 8 (session/global
  * state): module-level state must have an explicit eviction strategy. Three
  * further per-session maps share this same bound and the same FIFO discipline:
@@ -121,6 +147,13 @@ interface WakeBudget {
 	lastWakeAt: number;
 	lastSeenRevision?: number;
 	suspended: boolean;
+	/** Total number of wake attempts across the session lifetime. Never reset by
+	 * revision progress — only the consecutive counter resets on progress. */
+	totalWakes: number;
+	/** Discriminator for WHY the session was suspended: 'consecutive' means
+	 * the consecutive-unproductive budget was exhausted; 'total' means the
+	 * absolute per-session total-wake ceiling was reached. */
+	suspendedReason: 'consecutive' | 'total';
 }
 
 function canonicalGitHubPrUrl(value: string): string | null {
@@ -178,6 +211,7 @@ function blockedText(
 	mode: string,
 	options: {
 		suspended?: boolean;
+		suspendedReason?: 'consecutive' | 'total';
 		userInterrupted?: boolean;
 		maxConsecutive?: number;
 		recovery?: { code: string; requiredAction: string };
@@ -189,10 +223,16 @@ function blockedText(
 		`If the bind/checkout path is unreachable — for example a compound \`git fetch && git checkout\` was rejected as read-only shell syntax, the PR head cannot be fetched, or the working tree is on the wrong branch — call \`abort_pr_workflow\` (mode: "${mode}", reason: "<one-line cause>") to clear the gate, or ask the user to run \`/swarm abort-pr-workflow\`.`,
 	];
 	if (options.suspended) {
-		const max = options.maxConsecutive;
-		lines.push(
-			`Auto-resume is suspended after ${max ?? 'the configured number of'} consecutive unproductive retries (the durable gate \`revision\` did not advance). The session will not be re-awoken automatically. Run \`/swarm abort-pr-workflow\` or call \`abort_pr_workflow\` to clear the gate, or complete the workflow with \`complete_pr_workflow\` to publish normally.`,
-		);
+		if (options.suspendedReason === 'total') {
+			lines.push(
+				'Auto-resume is suspended because the total per-session wake budget has been exhausted. The session will not be re-awoken automatically. Run `/swarm abort-pr-workflow`, call `abort_pr_workflow`, or complete the workflow with `complete_pr_workflow` to publish normally.',
+			);
+		} else {
+			const max = options.maxConsecutive;
+			lines.push(
+				`Auto-resume is suspended after ${max ?? 'the configured number of'} consecutive unproductive retries (the durable gate \`revision\` did not advance). The session will not be re-awoken automatically. Run \`/swarm abort-pr-workflow\` or call \`abort_pr_workflow\` to clear the gate, or complete the workflow with \`complete_pr_workflow\` to publish normally.`,
+			);
+		}
 	}
 	if (options.userInterrupted) {
 		lines.push(
@@ -222,6 +262,7 @@ function workflowBanner(
 	mode: string,
 	options: {
 		suspended?: boolean;
+		suspendedReason?: 'consecutive' | 'total';
 		userInterrupted?: boolean;
 		maxConsecutive?: number;
 		recovery?: { code: string; requiredAction: string };
@@ -231,10 +272,16 @@ function workflowBanner(
 		`--- [${mode} WORKFLOW ACTIVE — output below is not a terminal verdict; \`complete_pr_workflow\` clears the gate on success; if the bind/checkout path is unreachable call \`abort_pr_workflow\` or run \`/swarm abort-pr-workflow\`] ---`,
 	];
 	if (options.suspended) {
-		const max = options.maxConsecutive;
-		lines.push(
-			`[Auto-resume is suspended after ${max ?? 'the configured number of'} consecutive unproductive retries. Run \`/swarm abort-pr-workflow\` or call \`abort_pr_workflow\` to clear the gate, or complete the workflow with \`complete_pr_workflow\`.]`,
-		);
+		if (options.suspendedReason === 'total') {
+			lines.push(
+				'[Auto-resume is suspended because the total per-session wake budget has been exhausted. Run `/swarm abort-pr-workflow`, call `abort_pr_workflow`, or complete the workflow with `complete_pr_workflow`.]',
+			);
+		} else {
+			const max = options.maxConsecutive;
+			lines.push(
+				`[Auto-resume is suspended after ${max ?? 'the configured number of'} consecutive unproductive retries. Run \`/swarm abort-pr-workflow\` or call \`abort_pr_workflow\` to clear the gate, or complete the workflow with \`complete_pr_workflow\`.]`,
+			);
+		}
 	}
 	if (options.userInterrupted) {
 		lines.push(
@@ -276,6 +323,12 @@ export function createPrWorkflowResponseGate(options: {
 	maxConsecutiveUnproductiveWakes?: number;
 	wakeCooldownMs?: number;
 	bannerCooldownMs?: number;
+	/** Override the per-tier default total-wake ceilings. A single number
+	 * applies uniformly to all tiers; a partial Record allows per-tier
+	 * overrides with unspecified tiers falling back to the defaults.
+	 * When omitted, the context-aware defaults (DEFAULT_TOTAL_WAKE_CEILINGS)
+	 * apply. */
+	totalWakeCeiling?: number | Partial<Record<PrReviewDepthTier, number>>;
 }) {
 	const wakeTimeoutMs = options.wakeTimeoutMs ?? DEFAULT_WAKE_TIMEOUT_MS;
 	const maxConsecutive =
@@ -284,6 +337,21 @@ export function createPrWorkflowResponseGate(options: {
 	const wakeCooldownMs = options.wakeCooldownMs ?? DEFAULT_WAKE_COOLDOWN_MS;
 	const bannerCooldownMs =
 		options.bannerCooldownMs ?? DEFAULT_BANNER_COOLDOWN_MS;
+	const resolveTotalWakeCeiling = (tier: PrReviewDepthTier): number => {
+		if (options.totalWakeCeiling === undefined) {
+			return DEFAULT_TOTAL_WAKE_CEILINGS[tier];
+		}
+		return typeof options.totalWakeCeiling === 'number'
+			? options.totalWakeCeiling
+			: (options.totalWakeCeiling[tier] ?? DEFAULT_TOTAL_WAKE_CEILINGS[tier]);
+	};
+	/**
+	 * Per-session wake budgets (consecutive counter, suspension flag, and
+	 * total-wake counter). This is an IN-MEMORY map scoped to ONE PROCESS
+	 * LIFETIME — it resets on plugin reload / process restart. The total-wake
+	 * bound holds within a single process invocation; a plugin reload starts
+	 * fresh totals.
+	 */
 	const activeWakeSessions = new Set<string>();
 	const wakeBudgets = new Map<string, WakeBudget>();
 	/**
@@ -311,12 +379,22 @@ export function createPrWorkflowResponseGate(options: {
 	const fallbackInjections = new Map<string, number>();
 	const session = options.client?.session as SessionClient | undefined;
 
-	/** FIFO-evict the oldest budget entry when the map exceeds the bound. */
+	/** FIFO-evict the oldest NON-suspended budget entry when the map exceeds
+	 * the bound. Suspended budgets are never evicted — they represent bounded
+	 * sessions whose total-wake cap fired. Evicting a suspended budget would
+	 * allow re-entry to recreate it with totalWakes: 0, re-arming auto-resume
+	 * and defeating the total-wake cap. If ALL entries are suspended (unlikely
+	 * but possible), eviction stops as a safety valve. */
 	function evictIfOverBound(): void {
 		while (wakeBudgets.size > MAX_TRACKED_WAKE_SESSIONS) {
-			const oldestKey = wakeBudgets.keys().next().value;
-			if (oldestKey === undefined) break;
-			wakeBudgets.delete(oldestKey);
+			let evicted = false;
+			for (const [key, budget] of wakeBudgets) {
+				if (budget.suspended) continue;
+				wakeBudgets.delete(key);
+				evicted = true;
+				break;
+			}
+			if (!evicted) break;
 		}
 	}
 
@@ -416,6 +494,7 @@ export function createPrWorkflowResponseGate(options: {
 
 		const budget = wakeBudgets.get(input.sessionID);
 		const suspended = budget?.suspended ?? false;
+		const suspendedReason = budget?.suspendedReason;
 		const userInterrupted = isPrWorkflowAutoWakeSuppressed(
 			options.directory,
 			input.sessionID,
@@ -461,6 +540,7 @@ export function createPrWorkflowResponseGate(options: {
 		// abort_pr_workflow can clear an unarmed workflow.
 		const banner = workflowBanner(state.mode, {
 			suspended,
+			suspendedReason,
 			userInterrupted,
 			maxConsecutive: maxConsecutive,
 			recovery,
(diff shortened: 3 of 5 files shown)


zaxbysauce:
## 🤖 Multi-Stage PR Review
**Pipeline:** MiniMax-M2.7-highspeed (orientation) (context pack) → MiniMax-M2.7-highspeed (explorer) + MiniMax-M2.7-highspeed (explorer B) (parallel explore, distinct lenses) → GLM-5-turbo (critique) ↔ GLM-5-turbo (critique) (cross-critique) → GLM-5-turbo (arbiter) (arbiter: blind-spot + synthesize)
**Commit reviewed:** `f69ab2734916`

---

### 🔍 PR Intent

- **O-001** Add a total per-session wake budget that increments on every wake attempt (including failures/timeouts) and is never reset by revision progress, with distinct suspension wording.
- **O-002** Make the total-wake ceiling context-aware: per-depth-tier defaults (S=12, M=54, L=102) derived from consolidation-floor workload, configurable via `totalWakeCeiling` option.
- **O-003** Widen depth-tier thresholds from 50/3/500/20 to 100/5/1500/50 so appropriately-sized PRs engage S/M tiers instead of defaulting to L.
- **O-004** Preserve all existing invariants: mandatory dimensions/families at every tier, fail-strict-to-L on unknown stats and submodule changes, FIFO eviction with MAX_TRACKED_WAKE_SESSIONS=200 bound.

---

### 📦 Implementation Summary

Adds `totalWakes` counter and `suspendedReason` discriminator to the `WakeBudget` interface. The total counter is incremented unconditionally after each wake attempt (in the post-wake bookkeeping block, after the consecutive check) and checked against a per-tier ceiling resolved via `resolveTotalWakeCeiling`. Suspended entries are protected from FIFO eviction to prevent re-arming. Both `blockedText` and `workflowBanner` branch on `suspendedReason` to emit distinct total-cap vs. consecutive-cap wording. Depth-tier thresholds are widened with corresponding comment and test updates. Seven targeted tests cover the new behaviour.

---

|---|---|---|
| O-001 | `SUPPORTED` | `src/hooks/pr-workflow-response-gate.ts:796-810` — `totalWakes += 1` on every wake, never reset; `suspendedReason: 'total'` set on exhaustion |
| O-002 | `SUPPORTED` | `src/hooks/pr-workflow-response-gate.ts:94-107` — `DEFAULT_TOTAL_WAKE_CEILINGS` S=12/M=54/L=102; `:337-350` — `resolveTotalWakeCeiling` with Partial Record + nullish fallback |
| O-003 | `SUPPORTED` | `src/hooks/pr-workflow-gate.ts:106-109` — thresholds widened to 100/5/1500/50 |
| O-004 | `SUPPORTED` | `computePrReviewDepthTier` unchanged except threshold values; eviction comment at `:379-389` documents suspended-entry protection; fail-strict-to-L paths unchanged |

---

### 🚨 Confirmed Findings

_None found._

---

_None found._

---

### 🧪 Test / Coverage Gaps

_None found._ — The seven new tests cover: total-cap under continuous progress, notice wording + recovery paths, failed wakes counting, timeout wakes counting, per-tier scaling with exact boundary (53 ok / 54 suspend), SC-001.3 healthy tier-L headroom (55 wakes), and partial-record override fallback. The dispatch-lanes test file verifies all 11 widened-threshold boundary cases. Coverage is adequate.

---

### 📋 Shipped-vs-Claimed Gaps

_None found._

---

### Minor

- `tests/unit/hooks/pr-workflow-response-gate-total-wake-budget.test.ts:70` — `driveProductiveWakes` helper is defined but never called; dead code.

---

### 🔁 Validation provenance

- **DROPPED** "evictIfOverBound breaks after deleting one entry" — misreads the `while` loop as an `if`; the loop re-evaluates `wakeBudgets.size > MAX_TRACKED_WAKE_SESSIONS` after each deletion.
- **DROPPED** "evictIfOverBound deadlock when all entries suspended" — intentional by-design behaviour documented in the function's own JSDoc; evicting suspended entries would re-arm auto-resume and defeat the total-wake cap.
- **DROPPED** "suspendedReason typed optional but needed when suspended=true" — theoretical type-level concern; every call site in this PR that passes `suspended: true` also passes `suspendedReason` from the budget, which is always set when `suspended` is set to true.
- **KEPT (as Minor)** "driveProductiveWakes dead code" — factual, trivially fixable.
- **Blind-spot pass**: no additional defects found. Checked: total-wake ordering vs. consecutive check (total runs after, correctly overwrites reason), PR_FEEDBACK tier fallback (defaults to 'L', safe), post-wake try/catch correctness, threshold/comment consistency, test boundary arithmetic.

---

### 📝 Merge Recommendation

**APPROVE**

The PR cleanly delivers both claimed fixes with correct logic, thorough boundary-tested coverage, and no real defects.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| No hardcoded secrets | ✅ |
| Lockfile consistent | ✅ |

---

zaxbysauce:
# swarm-pr-review — PR #2027

**Verdict: `REQUEST_CHANGES`** · 0 CRITICAL · 0 HIGH · 3 MEDIUM · 11 advisory LOW · 15 candidates disproved

<sub>Profile B (native subagents, no controller) · depth tier L (544 changed lines / 5 files + risk triggers) · bound head `f69ab27349165385c32541ef35dbaabb9b271e48` · merge base `5d1af842dae90d705e3de3590a922f47b2984ffc` (`origin/main`) · 6/6 base dimensions attested · 11/11 risk families evaluated (9 MATCHED + 2 NOT_TRIGGERED) · 46 candidates → 6 independent reviewer lanes → 4 critic lanes</sub>

---

## Summary

The core fix is sound and well-built. The total budget is genuinely immune to revision progress, the ceilings are honestly derived, failed/timed-out wakes count, the two suspension notices are distinct with no false cause, and the per-process caveat is documented on both surfaces. The single highest-severity candidate the review produced — that the tier-S ceiling of 12 would suspend legitimate reviews — was **disproved outright**.

What holds it back is three MEDIUMs that this repo's own contract treats as blocking rather than polish: the tier widening never takes effect on the default (non-controller) profile because its classifier doc wasn't updated, and two of the three self-declared final-council fixes ship with no regression test.

Each required change is small and bounded. This is a near-miss, not a rework.

---

## Obligation mapping

| Obligation | Source | Status |
|---|---|---|
| Total budget, never reset by progress | #1967 Problem A | ✅ MET |
| Distinct total-cap suspension wording | #1967 c2 | ✅ MET |
| Ceiling ≥100 or context-aware | #1967 c1 | ✅ MET |
| Document the per-process bound | #1967 c3 | ✅ MET |
| Failed/timed-out wakes count toward total | #1967 c4 | ✅ MET |
| Expose cap via factory options | #1967 c5 | ✅ MET |
| Widen depth-tier thresholds | #1967 Problem B | ⚠️ **PARTIALLY_MET** → F-001 |
| 6 dimensions / 11 families still mandatory | #1967 Problem B | ✅ MET |
| Three final-council fixes | PR body | ⚠️ **PARTIALLY_MET** → F-002, F-003 |
| 200-session bound preserved | PR body inv. 8 | ⚠️ **PARTIALLY_MET** → F-004 |
| Release fragment present | PR body inv. 12 | ✅ MET |
| Quantitative test-plan claims | PR body | ⚠️ **PARTIALLY_MET** → F-013 |

---

## Required before merge

### F-001 — MEDIUM — Problem B takes effect on zero Profile-B runs

`.opencode/skills/swarm-pr-review/SKILL.md:760-762` still documents the **old** thresholds (`≤ ~50 lines / ≤ 3 files`, `≤ ~500 lines`, `> ~500 lines / > ~20 files`) while the code now uses 100/5/1500/50 (`src/hooks/pr-workflow-gate.ts:105-110`). The PR touches zero skill files.

Why this is a wiring gap rather than a stale doc:

- The only caller of `computePrReviewDepthTier` is `bindPrReviewMergeBase` (`pr-workflow-gate.ts:1092`) — a **controller-only** path.
- `SKILL.md:758` headers that table *"Dispatch shape (Profiles B/C)"*, and `SKILL.md:775-785` scopes controller computation to *"Under Profile A"*.
- Plain Claude Code is canonical Profile B (`.claude/skills/swarm-pr-review/SKILL.md:16-18`) — the default path. Off-controller, that table **is** the classifier.

So the PR's headline Problem B deliverable lands nowhere on the dominant profile. `drift:check` gives no cover: `scripts/drift-check-docs-claims.ts:28-44` hardcodes three unrelated QA-gate claims and never parses this table.

In fairness: the drift direction is **fail-safe** (stale, tighter numbers over-classify → more lanes, never fewer), and this is an *incomplete change*, not a regression — doc and code agreed at base.

**Fix** — edit `.opencode/skills/swarm-pr-review/SKILL.md` only:
- `:760` → `~100 changed lines, ≤ 5 files`
- `:761` → `~1500 changed lines`
- `:762` → `> ~1500 changed lines, > ~50 files`

Do **not** touch the `.claude` / `.agents` copies — they are 59-line adapter pointers with no threshold table (adapter contract at `src/config/skill-mirrors.ts:175-181`), so the byte-identical mirror rule does not apply. No test asserts this table.

*Recommended, non-blocking:* register the three numbers in `DOCS_NUMERIC_CLAIMS` (`scripts/drift-check-docs-claims.ts:28-44`) so this class cannot silently drift again.

### F-002 — MEDIUM — skip-suspended eviction branch has zero test coverage

`src/hooks/pr-workflow-response-gate.ts:388-399`. `grep -rn MAX_TRACKED_WAKE_SESSIONS tests/` returns **nothing repo-wide**; no test builds >200 budgets. Mutating the branch breaks nothing.

The regression class is load-bearing, not cosmetic: a silent revert to plain FIFO evicts a suspended budget, re-entry recreates it with `totalWakes: 0`, and this PR's headline cap is defeated. It is cheap to test — `MAX_TRACKED_WAKE_SESSIONS` is exported at `:125` and `_inspectWakeBudget` at `:834`, and `totalWakeCeiling: 1` makes 201 suspended budgets trivial to construct.

**Fix** — assert suspended entries survive eviction while an unsuspended one is evicted. Scope to that assertion **only**; please do *not* pin the all-suspended `break` arm at `:397`, whose intended behavior is contested (see F-004) — testing it now would lock in possibly-wrong behavior.

### F-003 — MEDIUM — post-wake `try/catch` fallback has zero test coverage

`src/hooks/pr-workflow-response-gate.ts:742-755`. No test makes `readPrWorkflowGateState` reject, and the new test file never imports `_internals`. The nearest existing coverage (`pr-workflow-response-gate-cancellation.test.ts:129-184`) drives the same seam twice but only on the **resolve** arm.

This is a coverage gap, not a seam limitation — a reviewer probe overrode `_internals.readPrWorkflowGateState` to throw on alternate calls and drove 3 idle wakes; the catch executed 3× and `totalWakes` reached 3.

**Fix** — one throw-injection test on the post-wake read. Please also assert `totalWakes` still increments, so the test fails if the fallback is ever replaced by `null` rather than only if the `try/catch` is deleted.

### Why these three are blocking

- `AGENTS.md:168` — *"Prefer the smallest patch that closes the issue without unwired functionality, untested branches, or hidden regressions."*
- `AGENTS.md:161,163` — *"the evidence must be a concrete artifact: a command run with its output, a test that proves the invariant… If you cannot prove a touched invariant from source and test output, do not push."*
- `.opencode/skills/writing-tests/SKILL.md:569` — *"When fixing a bug surfaced by code review, swarm review, or post-merge audit, always add a regression test."*

The PR body self-declares all three as final-council fixes; exactly one of the three (the partial-record override, `total-wake-budget.test.ts:304`) got a test.

---

## Advisory (non-blocking)

**F-004 — eviction can wipe an active session's budget.** `evictIfOverBound` skips suspended entries (`:392`), runs *before* the insert (`:599` vs `:609`), and uses a strict `>` against 200. A critic reproduced it against the real factory: 200 suspended fillers + 1 live session → 30 idles produced **30 wakes** with `totalWakes` stuck at 1, because the recreated budget has `lastSeenRevision === undefined`, so `madeProgress` (`:640-644`) is true on every wake — the consecutive path is dead, not merely delayed, and the cooldown is dead too (`lastWakeAt: 0`). PR-introduced: at base the fresh entry is inserted last and can never be `keys().next().value` while 200 older keys exist.

**Not blocking**, deliberately: it requires 201 distinct sessions with uncleared durable gates in one plugin process (≈1000+ throttled wake attempts), against a realistic 1-5 concurrent workflows — while `main` today has *no* total cap at all and loops forever with a single revision-bumping session. Merging strictly reduces reachable harm.

Two notes if you do address it: the docstring at `:382-387` promises *"evict the oldest NON-suspended"* but at 200-suspended + 1-live the code evicts the **newest**. And on fix direction — evicting suspended entries would revert this PR's own cap, and evict-after-insert *alone* is strictly worse (the fresh entry becomes the immediate victim, and `budget` at `:600-610` becomes a detached object the wake still mutates). The minimal safe change is `if (key === sessionID) continue;` in the candidate loop; the only fix safe in both directions is persisting the suspension flag in the per-session gate-state file. Your call.

**F-005 — the post-wake catch comment is factually wrong.** `readPrWorkflowGateStateFromDisk` (`pr-workflow-gate.ts:6355-6368`) deliberately throws `BLOCKED: … is invalid` on corruption, so the bare catch swallows a *designed* signal while the comment calls it a transient fs error. Harm is bounded to in-memory bookkeeping — no write-back, and every controller tool path still uses the throwing reader. **Please do not "fix" this by rethrowing:** the catch is nested inside the `finally` opened at `:716`, so a rethrow would skip `:756-817` (no `lastWakeAt`, no `totalWakes` increment, no suspension check) and recreate the exact unbounded loop the finally exists to prevent. Fix the comment.

**F-006 — neither suspension branch emits a machine-readable event.** No `.swarm/events.jsonl` append, no log, no metric. The operator *does* get a distinct banner naming all three exits, so this is a telemetry gap rather than an invisible guardrail — materially unlike the #2025/#2023 case. The consecutive branch is equally silent (pre-existing), so it's a parity gap. Note `validateSwarmPath` is exported (`src/hooks/utils.ts:157`) and the pattern at `pr-workflow-gate.ts:982-994` is ~8 lines, so cost isn't the obstacle — scope is.

**F-007 — doc overclaim on budget scope.** `WakeBudget:150-156` says *"across the session lifetime"* / *"absolute per-session total-wake ceiling"*, and the release fragment says *"total per-session wake budget"*. True scope is per-process **and** per-workflow-instance (`resetBudget` fires on complete/abort). The obligation is still delivered — `resetBudget` runs only after an audited terminal transition, categorically unlike the still-active eviction path you guarded — but the two doc surfaces should be corrected the way the plugin-reload caveat already was at `:348-353`.

**F-008 — `totalWakeCeiling` is unvalidated.** Probe-proven: `Infinity` → 200 wakes unsuspended; `NaN` → 60 unsuspended; `0` → suspends at wake 1. No production caller passes it (`src/index.ts:1233` sends only `directory`/`client`), so it needs a first-party code change — but one `Number.isFinite`/positivity guard in `resolveTotalWakeCeiling` (`:340-347`) closes it, and no test covers any of these inputs.

**F-009** pre-bind wakes accrue against L(102) then compare against S(12)/M(54) using the same counter — reproduced, but realistic pre-bind count is 0-3, so effectively theoretical. **F-010** the ceilings are prose-linked, not derivation-linked, to `PR_REVIEW_BASE_LANE_FLOORS`/`PR_REVIEW_MICRO_LANE_FLOORS` — changing a floor silently mis-sizes them with no failing test. **F-011** `SKILL.md:1735-1737` still attributes suspension solely to consecutive-unproductive wakes, so the abort runbook now describes one of two causes (one edit pass closes this and F-001). **F-012** the `?? 'L'` tier fallback is never exercised — all 17 `writeStateWithRevision` call sites pass an explicit tier. **F-013** the body claims *"62 targeted tests across 5 response-gate test files"*; measured **47 pass / 0 fail / 308 expect()**, and no counting method reconciles 62 (the *"19 boundary tests"* claim **is** defensible: 18 `test(` + 1 `describe(`; `lint:ci` exit 0 is correct, the 2906 file count is environment-dependent — biome reports 2979 here). **F-014** the invariant audit cites stale line numbers — `:111-149, 321-337, 91` for code at `:388-399`, and `:644-699` for code at `:742-755`.

## Pre-existing (not introduced here)

`evictBannerDedupeMapsIfOverBound` is byte-identical to base · the consecutive-suspension branch was already event-less · `checkoutRecovery.requiredAction` path interpolation adds no new surface (both sites are unchanged context lines) · skill-mirror drift generally.

---

## Notable disproved candidates

Recording these so they don't resurface:

- **Ceiling under-sizing (the review's only HIGH)** — category error. A wake is consumed *only* by a `session.idle` event (`:562-570`), yielding exactly one injected prompt (`:691-693`; 12 idles → 12 `promptAsync` calls). Tool calls within a turn cost **zero** wakes, so a well-behaved tier-S review needs ~0 wakes, not the estimated 19-21. Same conflation invalidated the L→M retry-headroom claim.
- **Tier widening weakens coverage** — no. It changes lane *counts* only: base dispatch must still partition all six dimensions exactly once (`dispatch-lanes.ts:1094-1099`), `assertPrReviewBaseCoverageSettled` demands all six regardless of tier (`pr-workflow-gate.ts:1405-1419`), the trigger ledger requires all eleven families, and fail-strict-to-L on unknown stats/submodules is preserved.
- **`.strict()` schema rollback hazard** — the outer `PrWorkflowGateStateSchema` ends in `.passthrough()` (`:655-666`) with a ten-line comment explicitly rejecting `.strict()` for exactly this reason; `prReviewDepthTier` is pre-existing.
- **Mid-upgrade ceiling skew** — unconstructible: `wakeBudgets` holds the only `totalWakes` counter and is destroyed by the same reload that swaps the constant.
- **Tier re-derivation breaking in-flight reviews** — the tier is **pinned** at bind and never re-derived at settlement.
- **`bun run build` failure** — environmental (`node_modules/@vscode` absent in the review worktree), not PR-caused; `package.json`/`bun.lock` untouched in range.
- **Release fragment missing the per-process caveat** — it's on line 7.
- **Banner flooding / `forceFullBanner` untested** — `banner-dedupe.test.ts:146-183` already drives a second `textComplete` while suspended inside a 60s cooldown; repetition is bounded by per-`messageID` dedupe, and the line is pre-existing.
- **Fail-open-to-`L` on absent tier** — correct for a liveness brake; failing closed to S would falsely suspend PR_FEEDBACK sessions, and the rationale *is* stated at `:807-810`.
- **`abort_pr_workflow` → re-dispatch as counter evasion** — `resetBudget` fires only after an audited terminal transition (`clearPrWorkflowGateState` has exactly two callers; all writes are temp+rename, so no transient-null window). A fresh budget for a genuinely new workflow is correct-by-design. The residual is the F-007 doc overclaim, not an incomplete fix.
- **Boundary-test deletions** — none: 6 tier cases became 9, `{500,21}→L` was correctly updated to `{500,51}→L`, net assertions increased.

---

## Validation provenance

6 base explorer lanes (all six dimensions, each attested) · 11 risk families evaluated, 9 dispatched as micro-lanes with 3 COVERAGE-GATE retries after report-delivery failures, 2 `NOT_TRIGGERED` with absence evidence (`subprocess-platform`, `generated-provenance`) · 46 raw candidates, **0 suppressed** · 6 fresh reviewer lanes returned 46/46 `[REVIEWED]` rows → 26 CONFIRMED, 15 DISPROVED, 4 PRE_EXISTING, 1 UNVERIFIED (deduped) · 4 fresh critic lanes returned 13/13 `[CRITIC]` rows → 4 UPHELD, 9 DOWNGRADED, 0 critic-disproved · three findings carry execution-proven repros (ceiling-validation probe, throw-injection probe, 200-suspended eviction repro).

Deterministic signals: 23/23 CI checks pass; local `typecheck` exit 0, `lint:ci` exit 0 (2979 files), `drift:check` clean. Branch state at review time: `MERGEABLE` / `CLEAN`, head unmoved at `f69ab273`, no conflicts, no review threads, no requested changes.

Not available on Profile B and marked N/A rather than fabricated: `.swarm/repo-graph.json`, swarm verifier agents, and metrics persistence (no evidence writer — metrics kept in the session ledger).

<sub>Generated by `/swarm-pr-review` (Claude Code, Profile B). Read-only with respect to the PR branch — no code was modified.</sub>

Issue #1648, opened 2026-07-02
docs/commands.md has live drift against COMMAND_REGISTRY and no drift gate — generate or gate it


zaxbysauce:
## Problem

`docs/commands.md:3` declares itself generated-from-truth: "The authoritative source is `src/commands/registry.ts`; this page explains the user-facing behavior." There is no generator or drift test referencing it (`grep` across `src/`, `scripts/`, `package.json` returns zero hits), and drift already exists **today**: `docs/commands.md:424-428` documents the turbo command as `[on|off|lean|standard|status]` with "two strategies," while `registry.ts:1137` ships `[on|off|lean|standard|epic|status]` — a third strategy the docs don't mention. Separately, the repo's own regression guard is eroding: `tests/unit/commands/registry-documentation.test.ts:118-119` asserts stale prompt strings ('Show plan (optionally filter by phase number)' and 'Toggle Turbo Mode for the active session [on|off]') against a ≥5-of-8 tolerance, while the registry now reads 'Show *current* plan...' (`registry.ts:351`) and the fuller turbo description (`registry.ts:1137`) — the tolerance is silently absorbing real drift.

## Proposed Fix

Generate the per-command reference section of `docs/commands.md` from `COMMAND_REGISTRY`, reusing `buildHelpText`'s traversal (`src/commands/index.ts:124-273`) as a package script + CI check, keeping the hand-authored narrative (CC-conflict table, behavior prose) as a static preamble. If full generation is out of scope for this PR, at minimum add a drift test asserting every non-alias registry key and its description string appears in `docs/commands.md`. Either way, fix the two stale literals in `registry-documentation.test.ts:118-119` in the same change.

## Acceptance Criteria

- `docs/commands.md`'s turbo-command section matches `registry.ts:1137`'s actual `[on|off|lean|standard|epic|status]` set.
- A CI check fails if a future registry key/description diverges from the doc (generated doc, or a drift test with no silent-tolerance loophole).
- `registry-documentation.test.ts` literals match current registry strings.

## Effort / Dependencies
Effort: M. No dependency. Note: this is one instance of a broader "generate/drift-check docs from runtime source" pattern also needed for `docs/configuration.md` (see the JSON-Schema-generation issue) — consider a shared `scripts/drift-check.ts` extension point across both rather than two bespoke detectors.

## Source
[...]

PR #2433, opened 2026-08-30
feat(config): generate JSON Schema and complete configuration.md key reference from PluginConfigSchema

### LINKED ISSUE #1663: Generate JSON Schema + configuration.md reference from PluginConfigSchema (62-key config surface, silent-fallback typo hazard)
## Problem

**1. No JSON Schema exists for `opencode-swarm.json`.** `PluginConfigSchema` (`src/config/schema.ts:2213`) has 62 top-level keys (runtime-enumerated), hand-edited by users at `~/.config/opencode/opencode-swarm.json` / `.opencode/opencode-swarm.json` with zero editor autocomplete/validation. Several nested schemas are `.strict()` — the repo's own comment at `schema.ts:1671` documents the resulting failure mode: "a typo in any nested key will fail" the strict parse, and the loader silently falls back to guardrail-only defaults, i.e. **a config typo silently disables the user's entire configuration** with no signal at edit time.

**2. `docs/configuration.md` claims completeness it doesn't have.** `docs/index.md:29` says the configuration reference covers "all configuration keys," but at least 15 of the 62 top-level keys have zero word-boundary matches in the 1018-line doc (spot-checked: `guardrails`, `watchdog`, `turbo_mode`, `qa_retry_limit`, and others) — the claim is presently false, not just incomplete.

## Proposed Fix

Both share the same schema-walk; implement together:

1. Add a build step emitting `opencode-swarm.schema.json` from `PluginConfigSchema` (zod v4 `z.toJSONSchema` or `zod-to-json-schema`). Ship it in the npm package. Whitelist a `$schema` top-level key in both `PluginConfigSchema` and `src/services/config-doctor.ts`'s `KNOWN_TOP_LEVEL_KEYS` (`config-doctor.ts:21`). Have config-init paths (`swarm init`, project-init) write `$schema` into generated config files. Add a drift-check detector asserting the checked-in schema file matches the current zod source (fails if someone edits `schema.ts` without regenerating).
2. Generate the key-by-key reference section of `docs/configuration.md` from `PluginConfigSchema` (migrate existing per-key JSDoc comments, e.g. `schema.ts:2217-2238`, to zod `.describe()`), or at minimum add a drift-check detector diffing `Object.keys(PluginConfigSchema.shape)` against keys mentioned in `configuration.md`.

## Acceptance Criteria

- `opencode-swarm.schema.json` is generated at build time and published in the npm package.
- A config file with a `$schema` reference gets IDE validation/autocomplete in editors that support JSON Schema.
- `docs/configuration.md` documents all 62 top-level keys (verified by the drift check or by generation).
- `bun run drift:check` fails if the schema file or docs page falls out of sync with `PluginConfigSchema`.

## Effort / Dependencies
Effort: M. No dependency; both halves share one schema-walk implementation — do not split into two PRs.

## Source
opencode-swarm enhancement audit, Track G, run `20260702T035413Z`. Candidates ENH-G010-1 (critic: UPHELD_HIGH_VALUE), ENH-G010-2 (UPHELD_HIGH_VALUE — "user-facing docs are factually wrong today, not merely incomplete"). Roadmap: `docs/proposals/v8-feature-roadmap.md`. Note: this is one instance of a "generate/drift-check docs from runtime source" pattern shared with the `docs/commands.md` drift-gate issue — consider extending `scripts/drift-check.ts` with a common extension point for both rather than two bespoke detector frameworks.

zaxbysauce:
Closes #1663

## Summary

`PluginConfigSchema` is now the single source for three derived surfaces, all enforced by drift gates:

1. **`opencode-swarm.schema.json` (new, committed at repo root, shipped in the npm package).** Generated from `PluginConfigSchema` via zod v4's built-in `z.toJSONSchema(..., { io: 'input' })` (input mode is mandatory — output mode cannot represent the `.transform()` fields). Every top-level key now carries a `.describe()` annotation, so editors show per-key documentation and autocomplete. The root sets `additionalProperties: false` as *editor advice*: at runtime the root schema intentionally stays non-strict (legacy configs must keep loading), but a top-level typo should be flagged at edit time.
2. **Complete configuration reference.** `docs/configuration.md` gains a generated, marker-delimited "Top-level configuration keys" table (key, type, default, description, strict markers) — **30 of 70 top-level keys were previously undocumented**, while `docs/index.md` claimed the page covered "all configuration keys". That claim is now true. Strict sections are marked: unknown keys there fail config load (the loader's guardrails-only fallback path).
3. **Drift gates.** Two new `scripts/drift-check.ts` detectors — `config-schema` (checked-in file vs regeneration) and `config-docs` (generated docs section vs regeneration) — registered in the `DETECTORS` array (issue #1648 can extend the same pattern later). `bun run build` regenerates both artifacts automatically (step 2, right after `clean`); `bun run schema:generate` is the manual entry point. Failure messages name the regenerate command.

Plus the wiring the schema needs to be *used*:

- **`$schema` whitelisted in `PluginConfigSchema`** as parse-transparent metadata (`z.string().optional()` — absent stays absent, present is retained and ignored). `config-doctor`'s `KNOWN_TOP_LEVEL_KEYS` derives from the schema shape, so it is covered automatically.
- **Every config file the plugin authors now includes `$schema`**: `writeProjectConfigIfNew` starter (`src/config/project-init.ts`), the `.swarm/config.example.json` example, and both CLI `install` templates (global + project, `src/cli/index.ts`). The reference is version-pinned to the plugin version that authored the file (unpkg serves every published version); the artifact's own `$id` stays unversioned.
- **Runtime signal for silent typos** (beyond the issue's edit-time scope): the root config schema is not strict, so a typo'd *top-level* key (e.g. `guardrailz`, `$shcema`) was previously stripped with **zero** signal of any kind. The loader now emits one advisory warning listing unknown top-level keys (step 4b, after migrations/sanitize, before the first parse — verified it cannot duplicate the step-6 nested-key recovery warning, since the root never produces Zod `unrecognized_keys` issues).

## Invariant audit

- 1 (plugin init): **touched (content only)** — `src/config/project-init.ts` starter/example content now carries `$schema`; the write calls, their `'wx'` atomicity, and their bounded init-path placement are unchanged. No new init-path work: the generator runs only from `package.json#build` and `scripts/drift-check.ts` (both developer/CI contexts); `src/index.ts` is untouched (`git diff origin/main..HEAD -- src/index.ts` → empty).
- 2 (runtime portability): **touched** — `project-init.ts` gains only a `package.json` JSON import (same idiom as `src/index.ts`/`src/cli/index.ts`). Verified: `bun run build` exit 0 and `node --input-type=module -e "await import('./dist/index.js')"` → OK (default export shape intact; no new `bun:` surface). `dist/` not committed.
- 3 (subprocesses): **not touched** — no new subprocess calls anywhere in the diff.
- 4 (.swarm containment): **touched (content only)** — `writeSwarmConfigExampleIfNew` still writes only `.swarm/config.example.json`; project config still `.opencode/opencode-swarm.json`; the generator writes only the repo-root schema artifact and `docs/configuration.md` (developer build output, not runtime state).
- 5 (plan durability): **not touched** — no plan/ledger surface in the diff.
- 6 (test_runner safety): **not touched** — no test_runner or scope changes; all validation used direct shell `bun test` invocations.
- 7 (test writing): **touched** — 4 new `bun:test` files (68/133/139/111 lines — all under the 500-line cap), no `mock.module`, no new allowlist entries (`check:mock-cleanup` clean), temp dirs via `canonicalMkdtemp`/`createSafeTestDir` helpers, drift detectors take injectable `root` params instead of mocking fs.
- 8 (session state): **not touched** — no new module-level session state.
- 9 (guardrails/retry): **touched (warn-only)** — loader step 4b adds a single `advisoryWarn` for unknown top-level keys; the recovery ladder, fail-secure closure, and `ConfigBuildResult` metadata contract are unchanged. Regression test proves exactly one warning fires (no duplicate with step 6) and that recovery still preserves the rest of the config.
- 10 (chat/system msg): **not touched**.
- 11 (tool registration): **not touched** — no tool/agent/command/help surface changes; `bun run drift:check` (which includes tool/command/agent detectors) is clean.
- 12 (release/cache): **touched** — release fragment `docs/releases/pending/1663-config-json-schema-and-config-docs.md` added; no version files hand-edited (`package.json#version` still owned by release-please — diff only adds `files`/scripts entries); `opencode-swarm.schema.json` added to `package.json#files` and `scripts/package-smoke.mjs#REQUIRED_PACKAGE_FILES` so the CI package job validates tarball presence.

## Test plan

All run on the final squashed commit (23a55bd1, on top of origin/main d61042c6), Windows host:

- New + updated targeted tests: `31 pass / 0 fail` (4 new files: schema surface, generator contract, drift detectors, loader warn), `23 pass / 0 fail` (project-init + CLI write), `17 pass / 0 fail` (CLI install), `55 pass / 0 fail` (CLI write + install + package-smoke + drift-check integration).
- Blast radius: `tests/unit/config/` → **2118 pass / 0 fail** (98 files); config-doctor family → **64 pass / 0 fail**; `tests/unit/scripts/` → exit 0, 234 pass / 0 fail lines; `tests/unit/index*` → 27 pass / 0 fail; `tests/unit/agents/` → 1844 pass / 1 fail where the single failure (`buildProjectContext > returns null when no manifest is present`, 5 s timeout) was **proven pre-existing on stashed clean main** and is unrelated to this diff.
- Gates: `bun run drift:check` → `✅ No drift detected`; `DRIFT_CHECK_ENFORCE=1` → exit 0; `bun run build` → exit 0; `node --input-type=module -e "await import('./dist/index.js')"` → OK; `bunx tsc --noEmit` → exit 0; `bun run check:invariants` / `check:test-file-cap` (0 violations) / `check:mock-cleanup` (no new) / `check:pending-fragment` → all pass; `biome check --write` applied and clean.
- Generator idempotency: two consecutive runs → second reports `unchanged` for both artifacts.

## Notes

- The emitted schema deliberately diverges from runtime semantics at the root (`additionalProperties: false` vs runtime strip-and-warn): the file is authoring advice, not runtime rules. Documented in the schema's own `description` and the release fragment.
- Versioned `$schema` URLs in authored configs vs unversioned `$id`: a config file is validated by the schema of the version that authored it; unpkg serves every published version.
- The `DETECTORS` array in `scripts/drift-check.ts` is now exported for test introspection (a detector dropped from the list would otherwise silently stop running).

## Review follow-ups (commit f1a8a038)

The first CI run failed the `check:registry-citations` quality gate (this PR's line shifts created 2 new anchor failures and made 1 baselined entry stale). All findings from the swarm review and the maintainer review were then closed in a second commit — full disposition ledger in that commit message:

- **Malformed `$schema` containment** (maintainer review MF-1): a non-string `$schema` raised `invalid_type`, which the recovery ladder cannot strip — with `guardrails.enabled: false` it cascaded to guardrails-only defaults. `$schema` now uses `.catch(undefined)`: malformed values degrade to absent, exactly matching the pre-#1663 silent-strip semantics. Containment tests added.
- **Dormant legacy v6.12 keys** (maintainer review MF-2): `migratePresetsConfig` now strips `preset`/`presets`/`swarm_mode` with accurate wording when an `agents` block exists, so the new unknown-key warning no longer misreports them as typos.
- **Citation ratchet** (PRR-001): project-init citations repointed to the post-change lines; baseline regenerated 111→110 (the `cli/index.ts:285 saveJson` anchor now passes, so its stale entry is removed). Net baseline diff vs main: exactly one removed entry.
- **Warning dedup** (PRR-005): the unknown-top-level-key warning fires once per distinct key set, not once per config load.
- **Test pins** (PRR-004/006/008): rendered Type cells (union/enum/record), duplicate-marker span contract, CRLF normalization, `escapeCell` pipe branch, `$schema` on the CLI global-config path.
- **Docs** (PRR-007/009/010): `$schema` adoption + URL-convention note, README tip, "62+" → 71 key counts, `auto_select_architect` wording aligned with its absent default.
- Deferred with tracking: PRR-002/003 → #2436, PRR-011 → #2437. PRR-012 rejected (tests use `toContain`; nothing breaks).

Updated counts: targeted new/updated tests **43 pass / 0 fail** (4 new files); config family **2129 pass / 0 fail**; CLI suites **249 pass / 0 fail**; scripts family exit 0; build/typecheck/biome/dist-import/invariant gates all green at f1a8a038.


src/config/schema.ts
zaxbysauce, line 1: **MEDIUM-HIGH — malformed `$schema` can silently wipe an entire custom config, not just the bad field**

Promoting `$schema` from a silently-dropped unknown key to a typed `z.string().optional()` field changes its failure mode. A non-string `$schema` value (e.g. `{ $schema: 123 }`) now raises a Zod `invalid_type` issue rather than `unrecognized_keys`, so `stripUnrecognizedKeys` in `src/config/loader.ts` (which only targets `issue.code === 'unrecognized_keys'`) can't touch it.

Verified by direct repro against the bound head commit:
- `{ $schema: 123, agents: {...}, max_iterations: 7 }` alone → recovers via step 7b (sanitized-values), `$schema` dropped, rest of config preserved. Acceptable.
- Same payload **plus** `guardrails: { enabled: false }` anywhere in the merged raw config → step 7b's `!rawGuardrailsDisabled` guard skips entirely, falling all the way to step 8 (`guardrails_defaults`). Result: the user's entire config — `agents`, `max_iterations`, everything — is discarded, not just the malformed `$schema` value. End state is fail-secure (guardrails forced back on), so this isn't an exploitable bypass, but it's a real functional regression: a cosmetic metadata-field typo silently destroys an otherwise-valid, deliberately-configured setup for exactly the security-conscious users who explicitly disabled guardrails.

This is newly introduced by this PR — before it, any-typed `$schema` was silently ignored as an unrecognized top-level key with zero effect (root schema isn't `.strict()`).

Suggested minimal fix: either loosen `$schema` to tolerate a bad value (e.g. `.catch(undefined)` so a malformed value degrades to "absent" without producing a blocking Zod issue), or teach the recovery ladder to also strip `invalid_type` issues whose path is exactly `['$schema']` before step 6, since the field's own description already states it's inert runtime metadata.

No test currently covers a non-string `$schema` value (checked `tests/unit/config/loader-unknown-top-level-warn.test.ts` and `tests/unit/config/plugin-config-schema-surface.test.ts`).

---

src/config/loader.ts
zaxbysauce, line 1: **MEDIUM — new "unknown top-level config key" warning misfires on legacy `preset`/`presets`/`swarm_mode` keys**

`migratePresetsConfig` only strips the legacy `preset`/`presets`/`swarm_mode` top-level keys when `raw.presets && !raw.agents`. The new step 4b warning (this PR) then flags any top-level key not in `PluginConfigSchema.shape` as "unknown" — and that schema has no `preset`/`presets`/`swarm_mode` entries, so a config where these legacy keys coexist with a newer `agents` block survives migration untouched and gets flagged as if they were typos, e.g.:

```
Ignored 3 unknown top-level config key(s): preset, presets, swarm_mode
```

This isn't a contrived edge case: `mergedRaw` is built (global config merged with project config) *before* `migratePresetsConfig` runs, so it fires automatically for any user whose global `~/.config/opencode/opencode-swarm.json` is still a stale v6.12 presets-format file while a specific project's config uses the newer `agents` format — an ordinary, unforced upgrade path.

No data is lost (the `agents` block wins), so this is a diagnostics-accuracy issue rather than a correctness break, but it directly misleads users into thinking they should delete a recognized (if now-dormant) legacy field. No test in `tests/unit/config/loader-unknown-top-level-warn.test.ts` covers this exact scenario — only synthetic typos (`guardrailz`, `$shcema`) are tested.

Suggested minimal fix: when `raw.agents` is already present, still strip `preset`/`presets`/`swarm_mode` in `migratePresetsConfig` (with a distinct "legacy keys ignored, agents takes precedence" warning), or have step 4b's `unknownTopLevel` filter exclude this known legacy-key set and warn about them separately with accurate wording.

---

docs/releases/pending/1663-config-json-schema-and-config-docs.md
zaxbysauce, line 1: **INFO — off-by-one in the field count**

This note (and the PR body) say ".describe() annotations on all 70 top-level schema fields" / "30 of 70 top-level keys were previously undocumented." `Object.keys(PluginConfigSchema.shape).length` at the bound head commit is actually **71** (the new `$schema` key itself is the 71st, and it's included in the generated docs table). Cosmetic-only, but worth fixing before merge for accuracy.

---

scripts/generate-config-schema.ts
zaxbysauce, line 1: **LOW — two divergent `$schema` URL conventions, undocumented as a pair**

`CONFIG_SCHEMA_CANONICAL_URL` here (unversioned, used as the generated schema's own `$id`) and `CONFIG_SCHEMA_REF` in `src/config/project-init.ts:16` (version-pinned, what's actually written into user configs) are intentionally different per the PR body's "Notes" section, but that relationship isn't documented in code — each constant has its own standalone rationale comment, but neither cross-references the other, and no test asserts the relationship between them (each is only asserted independently). Not a functional issue, but a doc/test coverage gap that could look like an accidental inconsistency to a future contributor.

**INFO — nit:** `docs/configuration.md`'s example around line 27 also shows the unversioned canonical URL, while every actual config-writing call site (`project-init.ts`, `cli/index.ts`) writes the version-pinned `CONFIG_SCHEMA_REF`. Purely cosmetic since `$schema` is inert metadata, but a reader copying the doc's example literally would get a different URL shape than what the tooling actually generates.

---

scripts/drift-check.ts
zaxbysauce, line 1: **INFO — redundant schema walk (no action needed)**

`detectConfigSchemaDrift` and `detectConfigDocsKeysDrift` each independently call through to `buildConfigJsonSchema()`/`z.toJSONSchema(PluginConfigSchema, ...)`, so the full schema walk runs twice per `bun run drift:check` invocation instead of once. Confirmed this is a single in-process, non-network walk over ~580 fields — negligible cost (a few ms) relative to the dozen-plus other detectors in this file, several of which already do full filesystem walks. Flagging only for awareness, not worth blocking on.

---

tests/unit/scripts/generate-config-schema.test.ts
zaxbysauce, line 1: **LOW — PR body test-count claims are slightly off (harmless, but worth correcting)**

Running the 4 new test files together (`loader-unknown-top-level-warn.test.ts`, `plugin-config-schema-surface.test.ts`, `drift-check-config-schema.test.ts`, `generate-config-schema.test.ts`) gives **32 pass / 0 fail**, not the "31 pass / 0 fail" stated in the PR body. (For the record: the PR body's other counts — 23, 17, and the "55 pass" combined CLI+package-smoke+drift-check bucket — were all independently reproduced exactly as stated once the right file combination was found, so this looks like a simple off-by-one rather than a pattern.) No functional issue — just flagging for PR-description accuracy.

---

zaxbysauce (commented):
## Swarm PR Review — #2433

Structured review at commit `23a55bd1` (bound head; merge-base `d61042c6` against `origin/main`). Ran all 6 base review dimensions (intent-architecture, correctness-state, tests-falsifiability, security-trust, reliability-performance, compatibility-delivery) plus a full 11-family repository-agnostic risk sweep, followed by independent reviewer validation and a critic challenge pass on the two highest-impact findings.

**No CRITICAL or HIGH findings.** Two MEDIUM-level correctness/robustness findings, both empirically reproduced and reviewer/critic-confirmed:

1. **`src/config/schema.ts` / `src/config/loader.ts` (MEDIUM-HIGH):** a malformed (non-string) `$schema` value combined with `guardrails.enabled: false` anywhere in the merged config causes the recovery ladder to skip its targeted-sanitize step and fall through to `guardrails_defaults`, silently discarding the entire user config rather than just the bad field. Fails secure (no guardrails bypass), but it's a real regression — see inline comment for repro and suggested fix.
2. **`src/config/loader.ts` (MEDIUM):** the new "unknown top-level config key" warning misfires on legacy `preset`/`presets`/`swarm_mode` keys when they coexist with a newer `agents` block (a realistic post-upgrade state, since the merge happens before migration), mislabeling recognized-but-dormant legacy fields as typos. See inline comment for repro and suggested fix.

The rest are LOW/INFO polish items: a release-note field-count off-by-one (70 vs. actual 71), a PR-body test-count off-by-one (31 vs. actual 32 — all other stated counts reproduced exactly), an undocumented pairing between the two intentionally-different `$schema` URL constants, and a harmless redundant schema walk in the new drift detectors.

**Everything else checked out clean**, with actual evidence (not just trust in the PR body):
- Regenerated `opencode-swarm.schema.json` and `docs/configuration.md` from `PluginConfigSchema` directly against this head commit — byte-identical to the checked-in artifacts, confirming no drift and validating the idempotency claim.
- `$schema` wiring verified end-to-end at every config-writing call site (`project-init.ts`, `.swarm/config.example.json`, both CLI install templates).
- `package.json#files` and `scripts/package-smoke.mjs#REQUIRED_PACKAGE_FILES` both correctly updated for the new shipped artifact.
- Backward compatibility for existing configs (non-strict root schema, warn-not-throw on unknown keys) verified by reading the loader's full recovery ladder, not just the PR's description of it.
- No auth/secrets/subprocess/injection surface touched; no PII or config values (only key names) reach the new warning's log output.
- All 4 new test files use real fs/temp-dir fixtures (no `mock.module`), include a real idempotency test and a "detector registered in `DETECTORS` array" regression test (guards against unwired code), and all pass (32/32) when run directly.
- No merge conflicts; `mergeable_state` was `blocked` at review time (likely branch-protection/required-review, not a code conflict) — not a review-scope blocker.

Inline comments below carry full repro details and suggested minimal fixes for the two MEDIUM findings. Neither blocks on data-loss risk to guardrails enforcement itself (both fail secure), but both are worth fixing before merge given how easily the triggering states arise in normal usage.

---

Issue #2818, opened 2026-09-17
SCOPE WARNING attributes the repo's latest commit to whichever task is checked (cross-task contamination)


zaxbysauce:
Follow-up from #2814 (reported there under "Notes / observations"; surfaced by its recurrence sweep, `.agents/issue-traces/2814-missing-gates-recording/08a-recurrence-sweep.md` row 13; fix branch `fix/2814-stage-b-scope-freshness`).

## Summary

`validateDiffScope` (`src/hooks/diff-scope.ts:167-205`) derives "files changed by this task" from `git diff --name-only HEAD~1` (`:96-97`) — the most recent commit of the WHOLE repository — regardless of which task is being checked. When `update_task_status` runs for task A while the latest commit belongs to task B, task A is warned about task B's files.

Reported shape (#2814): `update_task_status(task-3.2, completed)` emitted `SCOPE WARNING: Task 3.2 ... also modified [InspectDialog.svelte]` where InspectDialog.svelte was legitimately modified under task 3.4's scope. The warning is advisory (does not block), but it mis-diagnoses concurrent multi-task sessions and survived the #2814 fix by design.

## Expected behavior

Per-task file attribution should come from the task's own attribution record (e.g. the session's task-keyed `modifiedFilesByTask` durable record or the background coder settlement's attributed files), not from the repository's latest commit. At minimum, the warning should name its evidence ("latest commit <sha>") so a reader can detect mis-attribution.

## Evidence

- `src/hooks/diff-scope.ts:96-97` (HEAD~1 diff), `:186-201` (warning construction).
- Trace: `.agents/issue-traces/2814-missing-gates-recording/` (local-only) — 03-localization-log.md step 3, 08a-recurrence-sweep.md P5.
- Release fragment: `docs/releases/pending/2814-stage-b-scope-freshness.md` ("Adjacent gaps confirmed separately").

zaxbysauce:
[...]

Issue #2817, opened 2026-09-17
Foreground Stage B settlement silently drops clean verdicts (unbound generation / route block / evidence-throw catch)


zaxbysauce:
Follow-up from #2814 (surfaced by its recurrence sweep, `.agents/issue-traces/2814-missing-gates-recording/08a-recurrence-sweep.md` rows 8-10; fix branch `fix/2814-stage-b-scope-freshness`).

## Summary

The FOREGROUND parallel Stage B settlement path (`src/hooks/delegation-gate.ts:6208-6545`) drops clean reviewer/test_engineer verdicts with only a debug-gated `logger.warn` line and no session-visible escalation, wedging the task at its current state with no recovery signal:

- `src/hooks/delegation-gate.ts:6303` — "ignoring unbound Stage B settlement" when `stageBDispatchGenerationsByCallID.get(callID)?.get(taskId)` is undefined. These maps are process-local with no persistence, so a plugin reload / host restart between dispatch and settlement makes EVERY settlement unbound.
- `src/hooks/delegation-gate.ts:6418` — "Stage B route receipt blocked" (`enforcePersistedReviewRouteReceipt` returns `canAdvance: false`).
- `src/hooks/delegation-gate.ts:6475` — "Stage B settlement rejected for" (catch around `recordGateEvidence`, e.g. `TASK_WORKFLOW_GENERATION_MISMATCH` from `src/gate-evidence.ts:1203`).

## Defect class

Same family as #2814's background path: a clean delegated-agent result is silently dropped by a fail-closed identity/fencing condition, leaving the task state machine wedged. The fail-closed DECISIONS are correct (AGENTS.md invariant 9: late results from an older generation must not clear current state); the gap is visibility and a recovery loop.

## Expected behavior

A dropped foreground settlement must (a) surface a session-visible advisory naming the task, the drop reason, and the remedy (re-dispatch), mirroring the background path's `[BACKGROUND COMPLETION ...] stale.` advisory, and/or (b) record a durable marker the architect/recover tooling can act on. The unbound-generation case deserves specific attention: after a process restart the in-memory dispatch maps cannot be reconstructed, so either persist the launch generation binding durably at dispatch or treat the post-restart shape as recoverable.

## Evidence

- Trace: `.agents/issue-traces/2814-missing-gates-recording/` (local-only, gitignored) — 08a-recurrence-sweep.md P3.
[...]

PR #2819, opened 2026-09-17
fix(workflow): scope stage-b gate freshness to declared review scope

### LINKED ISSUE #2814: update_task_status fails with 'Missing gates' even after reviewer/test_engineer return clean, correctly-formatted verdicts (gate-recording bug)
## Summary

After a coder task passes Stage A (`pre_check_batch` → `gates_passed: true`) and both Stage B agents return clean, correctly-formatted verdicts, `update_task_status(<task>, completed)` fails with `"Missing gates: [reviewer, test_engineer]"`. `check_gate_status` confirms `passed_gates: []` — the verdicts are never mechanically recorded. The task remains stuck at `pre_check_passed` indefinitely.

## Steps to reproduce

1. Declare scope for a task and dispatch `sovereign_coder`; coder commits the implementation.
2. Run `pre_check_batch` on the changed files → `gates_passed: true`.
3. Dispatch `sovereign_reviewer` and `sovereign_test_engineer` in parallel. Both return clean verdicts:
   - reviewer: `[REVIEWED] | task-X | APPROVED | ...`
   - test_engineer: `[TESTED] | task-X | PASS | ...`
4. Call `update_task_status(task-X, "completed")` → **fails** with `"Missing gates: [reviewer, test_engineer]"`.
5. `check_gate_status(task-X)` → `status: "incomplete"`, `passed_gates: []`, `missing_gates: [reviewer, test_engineer]`, `workflow.state: "pre_check_passed"`.

## Evidence

- **Task 2.6** (this session): reviewer `[REVIEWED] | task-2.6 | APPROVED | ...` and test_engineer `[TESTED] | task-2.6 | PASS | 32/32 tests passed, biome clean, 0 placeholders` — both emitted the exact bracketed structured verdict line as the final output. `check_gate_status(2.6)` still shows `passed_gates: []`. `workflow.generation: 2`, `state: pre_check_passed`.
- **Task 3.2**: recovered from `rework_required` via `recover_rework_task` (supervised `stage_a_passed`, gen 1), then fresh reviewer APPROVED + test_engineer PASS 23/23 — still not recorded.
- **Task 3.3**: reviewer APPROVED (all spec points) + test_engineer PASS 29/29 — never recorded.
- **Task 3.4** (control): recorded **correctly** — `check_gate_status(3.4)` → `all_passed` with both gates present. So the mechanism *can* work; it fails intermittently/systematically on other tasks.

## What was ruled out

- **Not an agent-format issue**: re-dispatching with an explicit requirement to emit the exact `[REVIEWED] | ...` / `[TESTED] | ...` bracketed line as the final output did NOT fix recording. Both agents emitted the exact format; recording still failed.
- **Not a Stage A failure**: `pre_check_batch` returned `gates_passed: true` in every case.
- **Not a scope issue**: scope was correctly declared via `declare_scope` with `replace_existing: true` for each task.

## Notes / observations

- `update_task_status` error also emitted a spurious `SCOPE WARNING` claiming task 3.2 "also modified InspectDialog.svelte" — cross-task contamination in the scope-tracking window (InspectDialog was legitimately modified under task 3.4's scope, not 3.2's).
- `recover_rework_task` only handles `rework_required` state; tasks stuck at `pre_check_passed` have no architect-accessible recovery path.
- `/swarm recover <task>` reported "workflow state is pre_check_passed with pre_check proof present — nothing to repair" for 2.6, confirming the state is considered healthy while the gates remain unrecorded.

## Expected behavior

When both Stage B gates return clean, correctly-formatted verdicts, `update_task_status(<task>, "completed")` should succeed and `check_gate_status` should show `passed_gates: [reviewer, test_engineer]`.

zaxbysauce:
Closes #2814

## Summary

**Root cause.** 
A background Stage B reviewer/test_engineer completion is accepted only if the workspace is byte-identical (whole tree) to dispatch time. The dispatch path stores a whole-tree snapshot (`captureWorkspaceSnapshotAsync`, `src/hooks/delegation-gate.ts:5713-5718`) whose `scope` field is never consulted by the freshness predicate: `compareStageBWorkspace` (`src/background/stage-b-gates.ts`) delegated to `workspaceSnapshotMatches` (`src/background/workspace-snapshot.ts:2093-2134`), which compares global `gitHead` + `dirtyHash` + `prHeadSha`. Any concurrent workspace activity — another task's commit, an unrelated untracked file — marked the completion `stale` (`src/background/completion-observer.ts:452-464`), dropped the clean verdict before it was ever parsed, left the record permanently un-consumed (the ingestion claim only re-claims `completed`/`ingestion_error`), and wedged the task at `pre_check_passed` with `passed_gates: []` — exactly the reported `Missing gates: [reviewer, test_engineer]` with no recovery path (`/swarm recover` repairs only `coder_delegated`; `recover_rework_task` only `rework_required`).

**Fix.**
`compareStageBWorkspace` now narrows the tree-state legs of the freshness check to the dispatch snapshot's declared review scope; the project-root and PR-head identity legs still bind the whole workspace.

- `src/background/stage-b-gates.ts` — derive `scopeFiles` from the stored scope string (per-entry plausibility filter: bare task ids, empty fragments, absolute/traversal paths dropped individually; whole-tree fallback when nothing plausible remains or either snapshot is capture-degraded with null `gitHead`/`changedFiles`); stale iff a scope file appears in the committed diff between the dispatch and current heads OR the dirty/untracked scope-file set changed between the snapshots. Docs role keeps its existing `dirtyHash` exemption. Scoped stale reasons (`in-scope committed change: ...`, `in-scope dirty set changed: ...`) tell the architect exactly what invalidated a verdict.
- `src/background/workspace-snapshot.ts` — new bounded helper `committedFilesBetween(directory, fromHead, toHead)` (reuses the existing sync `runGit`: array-form, explicit `cwd`/`-C`, `stdin: 'ignore'`, timeout, bounded buffer; `[]` for identical heads, `null` on git failure → whole-tree fallback).

Tasks already wedged before upgrading do not auto-recover (stale records are swept, not replayed): re-dispatch the Stage B gates — a fresh Task call now records correctly under concurrent activity. Known residual (documented in the release fragment): the narrowed predicate compares the dirty scope-path SET plus the committed diff, not per-file content — an in-scope file already dirty at dispatch and edited mid-run without any commit can slip past; preconditions are incompatible with the coder-commits-before-Stage-B flow.

A swarm-pr-review round (2819-r1, posted as a PR comment) verified 7 findings; commit c5fa83213 resolves them: the committed-diff leg now runs `git diff --no-renames` (a committed rename of an in-scope file invalidates the verdict), scope matching case-folds on Windows, the committed-diff null-failure leg and the identity legs gained dedicated tests (19 tests total in the regression file), and the release fragment's absolute narrowing claim was rescoped. One finding (comma-in-filename scope strings) is documented as a known limitation of the pre-existing scope-string format.

Adjacent gaps surfaced by the recurrence sweep, confirmed separate defects, filed: #2817 (foreground Stage B settlement silent drops at `src/hooks/delegation-gate.ts:6303/:6418/:6475`) and #2818 (cross-task `SCOPE WARNING` mis-attribution at `src/hooks/diff-scope.ts:96-97` — that warning still emits its mis-attributed noise after this fix).

## Recurrence Prevention (defect class)

- Defect class: a clean delegated-agent result silently dropped by a global (unscoped) workspace identity check or a silently-swallowed settlement condition, wedging the task state machine with no recovery path.
- Sweep result: 5 predicates, 13 hits, all dispositioned (4 fixed/fixed-shape, 6 distinct-contract/pre-existing-correct, 3 same-class foreground → #2817) — `.agents/issue-traces/2814-missing-gates-recording/08a-recurrence-sweep.md` (local-only trace).
- Guardrail: `tests/unit/background/stage-b-scope-freshness-2814.test.ts` (11 tests) — demonstrated to bite by reverting only `stage-b-gates.ts` to base (4 tests fail exactly on the defect shape; the 3 preserving arms stay green), plus frozen differential checks C1–C5 replayed in a disposable worktree at the base commit (C1–C3 RED→GREEN, C4–C5 GREEN-to-GREEN; `repro/checkpoint.manifest`, verify-checkpoint OK on all paths).

## Invariant audit

- 1 (plugin init): not touched — no initialization code changed; `bun run build` passed.
- 2 (runtime portability): touched (evidence) — `src/background/*` is bundled; `bun run build` + `node --input-type=module -e "await import('./dist/index.js')"` passed; `tests/unit/build/bundle-portability.test.ts` + `bundle-plugin-shape.test.ts` 12 pass; no `bun:` imports added (grep of diff).
- 3 (subprocesses): touched (evidence) — the one new git invocation (`committedFilesBetween` → existing sync `runGit`) reuses the bounded profile: array-form spawn, explicit `cwd`/`-C`, `stdin: 'ignore'`, `timeout`, bounded `maxBuffer`; `null` on failure → fail-closed whole-tree fallback.
- 4 (.swarm containment): not touched — no runtime-state sinks or containment policy changed.
- 5 (plan durability): not touched — no plan ledger or projection code changed.
- 6 (test_runner safety): not touched — no OpenCode `test_runner` validation used; shell per-file loops only.
- 7 (test writing): touched (evidence) — new `tests/unit/background/stage-b-scope-freshness-2814.test.ts` uses `bun:test` only, no `mock.module` (zero mocks), `canonicalMkdtemp` from `tests/helpers/tmpdir.ts`, 277 lines (FR-006 < 500); `check:test-file-cap` 0 violations, `check-test-tmpdir` 0 blocking, `check-mock-cleanup` clean.
- 8 (session state): not touched — no new session-scoped or module-global state.
- 9 (guardrails/retry): not touched — generation fencing (`assertExpectedGeneration`) and stale-record sweep semantics unchanged; the fix only narrows WHICH workspace changes invalidate a verdict.
- 11 (tool registration): not touched — no tool, agent-map, command, or help surface changed.
- 12 (release/cache): touched (evidence) — release fragment `docs/releases/pending/2814-stage-b-scope-freshness.md` added; no version/CHANGELOG/manifest files hand-edited.

## Test plan

- Regression suite: `bun --smol test tests/unit/background/stage-b-scope-freshness-2814.test.ts --timeout 120000` → 11 pass / 0 fail.
- Pinned Stage B suites: `bun --smol test tests/unit/background/{stage-b-docs-workspace,completion-observer,delegation-lifecycle-reviewer-verdict}.test.ts` → 15 pass / 0 fail.
- Blast radius: all 39 test files importing the two changed modules run per-file with single-retry → failTwice=0.
- Typecheck `bunx tsc --noEmit` clean; `bunx @biomejs/biome@2.3.14 ci .` rc=0 (4 pre-existing warnings, none in touched files); `bun run build` OK; `node --input-type=module -e "await import('./dist/index.js')"` OK; bundle-portability + plugin-shape 12 pass; `check:test-file-cap` 0 violations; `check-test-tmpdir` 0 blocking; `check-test-clock` 0 blocking; `check-mock-cleanup` clean; `check:registry-citations` pass; retention-registry + atomic-write ratchet 24 pass; `check-bash-portability` and `check-cross-contamination` run (cross-contamination is load-flaky locally; CI is authoritative).
- Deferred-work scan: `.opencode/skills/issue-tracer/scripts/scan-deferred.sh` → clean.

## Regression Protection

- New file above pins: out-of-scope COMMIT (both routing configs) and out-of-scope UNTRACKED file must NOT invalidate verdicts (workflow reaches `tests_run`); in-scope dirty, in-scope committed, and dispatch-dirty-then-reverted MUST still invalidate; scope-less (bare task id / null) and capture-degraded (null heads) fall back to whole-tree semantics.
- Consumer contract exercised through the real exported `checkReviewerGate` (`src/tools/update-task-status.ts:438`) — not a reimplementation.
- Test drift review: mutation probe documented in the trace; checks frozen pre-fix at base `ec66974f2` (checkpoint manifest + AMEND ledger).

| Acceptance criterion (from intake) | Evidence (command + output, or test name) |
|---|---|
| AC1 out-of-scope COMMIT no longer drops the reviewer gate | `repro/check-c1.sh` → `PASS C1 ... gates=[pre_check, reviewer, test_engineer], workflowState=tests_run` (base replay RED) |
| AC2 out-of-scope UNTRACKED file: both gates + durable `tests_run` (both routing configs) | `repro/check-c2.sh` → `PASS C2` both arms (base RED) |
| AC3 `update_task_status(completed)` no longer blocks with missing Stage B gates | `repro/check-c3.sh` → `PASS C3` — real `checkReviewerGate`: `blocked=false, missingGates=[]` (base RED) |
| AC4 in-scope drift still rejects the verdict (dirty / committed / reverted) | `repro/check-c4.sh` → `PASS C4` all three arms, records stay `stale` (GREEN on base and head) |
| AC5 existing Stage B freshness semantics preserved | `repro/check-c5.sh` → `PASS C5`, 15/15 across the three pinned suites (GREEN on base and head) |

## Risk and Rollback

- Risk level: medium (touches the gate-freshness invariant for every background Stage B completion) — mitigated by fail-closed fallbacks, the preserving checks C4/C5, and the 39-file blast-radius run.
- Rollback: revert the three commits; no data migration (durable record schema unchanged — `scope` was already persisted).
- Residual risk: the documented path-set-vs-content narrowing (fragment: "Known residual"); foreground-path drops tracked in #2817.

## Waivers (or none)

none

## Merge status

Awaiting explicit user approval; not merged.

PR head: c5fa83213fe7b672984049d963a1b3803274b848

## Issue Closure

Closes #2814


zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `1f70c5198a42`

---

### 🔍 PR Intent

Reconstructed obligation list (from PR description, issue #2814, release fragment, diff):

- **O-001** Out-of-scope concurrent commits must NOT invalidate reviewer/test_engineer verdicts
- **O-002** Out-of-scope untracked files must NOT invalidate reviewer/test_engineer verdicts
- **O-003** In-scope dirty modifications MUST still invalidate verdicts
- **O-004** In-scope committed changes MUST still invalidate verdicts
- **O-005** In-scope dispatch-dirty-then-reverted files MUST still invalidate verdicts
- **O-006** Bare task-id scope must fall back to whole-tree semantics (preserve pre-existing behavior)
- **O-007** Null scope must fall back to whole-tree semantics
- **O-008** Capture-degraded snapshots (null gitHead/changedFiles) must fall back to whole-tree semantics
- **O-009** `update_task_status(completed)` must no longer fail with `Missing gates: [reviewer, test_engineer]`
- **O-010** Stale reasons must be scoped (`in-scope committed change: ...`, `in-scope dirty set changed: ...`) not whole-tree

---

### 📦 Implementation Summary

The PR adds a new `committedFilesBetween` helper to `workspace-snapshot.ts` and rewrites `compareStageBWorkspace` in `stage-b-gates.ts`. Instead of delegating all three legs (gitHead, dirtyHash, prHeadSha) to `compareWorkspaceSnapshots`, the new logic:

1. For `docs` agent — unchanged: whole-tree comparison with dirtyHash exempted.
2. For `reviewer`/`test_engineer` — narrows to declared scope:
   - Checks project-root + PR-head identity via `workspaceSnapshotMatches` (gitHead/dirtyHash nulled).
   - Computes committed files between dispatch and current HEAD; fails closed if the git call fails.
   - Flags stale if any scope file appears in that diff.
   - Compares the set of dirty scope files between dispatch and current snapshots; flags stale on any change.
   - Falls back to whole-tree on bare task-id scope, null scope, or capture-degraded snapshots.

---

|---|---|---|
| O-001 | `SUPPORTED` | `stage-b-gates.ts:311-335` — out-of-scope commits land in neither `scopeDirtySlice` nor `committedInScope` |
| O-002 | `SUPPORTED` | `stage-b-gates.ts:311-313` — untracked files only affect `changedFiles`; `scopeDirtySlice` filters to scope only |
| O-003 | `SUPPORTED` | `stage-b-gates.ts:337-344` — new dirty scope file added to currentDirty but not dispatchDirty → dirtied set non-empty → stale |
| O-004 | `SUPPORTED` | `stage-b-gates.ts:326-334` — `committedInScope` includes the file → stale with scoped reason |
| O-005 | `SUPPORTED` | `stage-b-gates.ts:337-344` — `revert-in-scope-dispatch-dirty`: file removed from currentDirty → cleaned set non-empty → stale |
| O-006 | `SUPPORTED` | `stage-b-gates.ts:297-301` — `deriveScopeFiles` returns null for bare task-id → whole-tree fallback |
| O-007 | `SUPPORTED` | `stage-b-gates.ts:297-301` — null scope → `deriveScopeFiles` returns null → whole-tree fallback |
| O-008 | `SUPPORTED` | `stage-b-gates.ts:302-310` — null gitHead/changedFiles on either snapshot → whole-tree fallback |
| O-009 | `SUPPORTED` | `stage-b-gates.ts:294-296` — identity leg binds `prHeadSha`; PR head re-resolved via `resolveCurrentPrHeadSha` at capture time |
| O-010 | `SUPPORTED` | `stage-b-gates.ts:330-335, 343-346` — scoped stale reasons embedded in return value |

---

### 🚨 Confirmed Findings

> No confirmed findings. All candidates were challenged and refuted; see §🔁 for disposition.

---

_All candidates downgraded from the confirmed-findings list after challenge._

- **Risk**: `workspaceSnapshotMatches` is called with gitHead/dirtyHash=null on both arguments (`stage-b-gates.ts:314-317`). If that function has a latent bug on null fields, it would surface silently.
  - *Why suspicious*: this is a new call path (nulling those fields reuses the identity leg of an existing function rather than a new helper).
  - *What would verify it*: the existing pinned Stage B test suites (delegation-lifecycle-reviewer-verdict, stage-b-docs-workspace) exercise `workspaceSnapshotMatches` indirectly; whether they exercise it specifically with nulled gitHead/dirtyHash is not in scope to verify statically.
  - **Confidence**: low — the function has null guards in its body; the test suites run the full ingestion path end-to-end.

---

### 🧪 Test / Coverage Gaps

- **Gap**: `committedFilesBetween` returning `null` (git failure) → whole-tree fallback path has no isolated unit test.
  - *Evidence*: `stage-b-scope-freshness-2814.test.ts` tests the `directory === undefined` fallback (line 247), which short-circuits *before* `committedFilesBetween` is called. No test directly exercises the git-failure → `null` → fallback leg.
  - *Mitigated by*: the fallback is provably reachable only through an internal git subprocess failure; the null-on-failure contract is unit-tested in `workspace-snapshot.ts`'s own test file (separate from this PR), and the whole-tree fallback's correctness on moved heads is already covered by the `in-scope committed drift is reported with a scoped reason` test (which tests the fallback indirectly). The gap is real but low severity — a git failure mid-run is already a fatal condition for the agent anyway.

---

### 📋 Shipped-vs-Claimed Gaps

_None found._

---

### 📝 Merge Recommendation

**APPROVE**

The PR correctly narrows the Stage B workspace freshness check to the declared review scope. All six obligations for the fix are confirmed in code. The two confirmed findings raised by the reviewer were both false positives (the `null` vs `''` asymmetry is intentional — `null` is filtered by `scopeDirtySlice` AND excluded by `committedInScope`'s `?? ''` fallback because an empty string never matches a Set entry; the log-injection claim assumed the stale reason flows to raw stdout without the structured-event intermediary). The single real gap (no isolated test for git-failure → `null` → fallback) is acknowledged above and is mitigated by indirect coverage; it does not block.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| Test coverage adequate | ✅ (with noted gap: git-failure fallback not isolated) |
| No hardcoded secrets | ✅ |
| All async errors handled | ✅ (`committedFilesBetween` returns `null` on git failure → fail-closed whole-tree fallback) |
| Input validation present | ✅ (`deriveScopeFiles`, `normalizeAttributionPath`) |
| Lockfile consistent | ✅ (no new dependencies) |

---

### 🔁 Validation Provenance

**Confirmed findings CHALLENGED and DROPPED (reason):**

- **[HIGH] `committedInScope` null vs `''` asymmetry** — `null` is excluded by `scopeDirtySlice`'s `normalized !== null` guard AND by `scopeFiles.has('')` in `committedInScope` (empty string never matches a real Set entry). The "asymmetry" is intentional by-design: both paths produce the same outcome (silently exclude implausible-path files from scope). This is documented behavior per `normalizeAttributionPath` contract.
- **[MEDIUM] `scopeDirtySlice` vs `committedInScope` null-handling asymmetry** — same as above; not a defect.
- **[LOW] Log injection in stale reason** — the stale reason is a structured return-field that feeds structured event channels, not raw stdout. Git `diff --name-only -z` output with literal newlines in path names cannot be produced by normal git operations; the risk is theoretical and not actionable.
- **[MEDIUM] No unit test for null gitHead/dirtyHash call path** — the pinned Stage B suites exercise `workspaceSnapshotMatches` through the full ingestion path; the call path is covered by integration. Additionally, this is a pre-existing function whose null-guard behavior was not changed.
- **[LOW] `committedFilesBetween` unbounded array** — git output is bounded by `GIT_SNAPSHOT_MAX_BUFFER = 32 MB` at the stream level; `parseNulPaths` is called on bounded output. No additional bound is needed.
- **[MEDIUM] `committedFilesBetween` null-return not tested** — **KEPT** (above, gap section). This is a real coverage gap; all others were false positives.

**Blind-spot findings added:**
None beyond the coverage gap for git-failure → null → fallback (kept in gap section above).

**Net verdict:** APPROVE — no blocking defects; one acknowledged coverage gap with existing mitigations.

---

zaxbysauce:
## swarm-pr-review — PR #2819

**Reviewed at head `1f70c5198a423591472ee50590788952221fc693`** (matches live `headRefOid`; confirmed unmoved throughout review). Base/merge-base `ec66974f2bc9c89a9c1cc8dec94894615e7e2387`. Capability profile: **B** (Claude Code native parallel subagents, no swarm controller). Depth tier: **M** (small diff — 469+/6-, 4 files — escalated from S because it touches a concurrency/state-freshness trust gate, a risk trigger).

### PR intent

Fixes #2814: Stage B (background `reviewer`/`test_engineer`) gate-freshness checking compared the **whole workspace tree**, so any unrelated concurrent change permanently wedged a task at `pre_check_passed` with no recovery path. The fix narrows the tree-state legs of `compareStageBWorkspace` (`src/background/stage-b-gates.ts`) to the dispatch snapshot's declared review `scope`, via a new `deriveScopeFiles` plausibility filter and a new `committedFilesBetween` git helper (`src/background/workspace-snapshot.ts`). Identity legs (project root, PR head) still bind the whole workspace.

### Validation pipeline

3 parallel base-explorer lanes (all 6 review dimensions + the 11 mandatory risk-family checks) → 1 independent opus reviewer (ran the real exported `compareStageBWorkspace` against constructed and real-git-repo inputs) → 1 opus critic (independently re-derived with its own executed probes against real temp git repos, adversarially challenged every reviewer claim, and found one defect the reviewer missed).

### Confirmed findings

---

**F1 — HIGH — The scope-narrowing filter silently disables freshness detection for directory, stale, or case-mismatched scope entries — a new false-negative class, not present pre-PR**

Files: `src/background/stage-b-gates.ts:267-278` (`deriveScopeFiles`), `:286-290`, `:347-353` (in-scope filters)

`deriveScopeFiles` returns a non-null scope set as long as at least one fragment is "plausible" (not a bare task-id / empty / absolute / traversal path) — it does not require the fragment to actually match any file that could ever appear in a diff. Because the in-scope filters do exact `Set.has()` string matching, a scope entry that is **directory-shaped**, **points at a nonexistent/stale file**, or **differs only by case** (Windows) makes both narrowed tree-state legs (committed-diff and dirty-set) pass vacuously — real in-scope drift during the review window goes undetected, and a stale/invalidated Stage B verdict is accepted as fresh.

This is reachable through a legitimate, unvalidated path: `session.declaredCoderScope` is assigned directly from `FILE:` prompt directives (`src/hooks/delegation-gate.ts:7067`) via `normalizeFileDirectiveValue` (`:2470-2495`), which performs no existence or file-vs-directory check, and flows unchanged into the Stage B dispatch snapshot's `scope` field (`:5697-5701`). The tool-level path (`declare_scope`, `src/tools/declare-scope.ts`) is **not** a mitigation — the critic executed `validateFiles` directly against a directory path, a nonexistent-file path, and a case-mismatched path; all three returned `[]` (no validation error), and the tool's authority preflight (`:399-406`) accepts a directory because `isPathWithinDeclaredScope` (`src/hooks/guardrails/file-authority.ts:947-949`) is prefix-matching, not exact-matching — an inconsistency between how scope is *enforced* (prefix) and how it's *freshness-checked* (exact).

Evidence type: **EXECUTION_PROVEN**. Both the reviewer and critic independently ran the real exported `compareStageBWorkspace` (reviewer: constructed inputs; critic: real temp git repos) and reproduced, with a genuinely changed `dirtyHash`:
- scope `src/hooks` (directory) + real changed file under it → `stale:false` (pre-PR: `true`)
- scope `src/hooks/does-not-exist.ts` (stale/nonexistent) + unrelated real changed file → `stale:false` (pre-PR: `true`)
- scope `src/Hooks/Widget.ts` (case mismatch) + real changed `src/hooks/widget.ts` → `stale:false` (pre-PR: `true`)

Why HIGH and not just "doesn't fully fix #2814": this doesn't merely fail to improve on a pre-existing gap — for these three input shapes it is **strictly worse than pre-PR whole-tree comparison**, which would have correctly flagged all three as stale. It silently disables the *only* tree-state guard on the Stage B trust boundary (`stage-b-gates.ts:710`) for scope declarations that are legitimate and reachable everywhere else in the system.

Falsification probe: re-run the PR's own `commit-in-scope` test scenario (`tests/unit/background/stage-b-scope-freshness-2814.test.ts:94-101`) with `scope: 'src'` (a directory containing the in-scope file) instead of the exact file path, and assert it still rejects — it will not, on the current code.

**Suggested-fix caveat (important for whoever picks this up):** the naive fix — "return null (whole-tree fallback) unless every scope entry exact-matches a path present in either snapshot" — **regresses #2814**. The critic executed this against the PR's own primary scenario (coder commits the in-scope file *before* Stage B dispatch): both snapshots have `changedFiles: []`, so the exact scope file is present in *neither*, forcing whole-tree fallback and failing the PR's own `commit-out-of-scope` test. Any real fix must satisfy: **a tracked scope file with no pending changes must still yield a non-null scope set.** The surviving direction is prefix-matching directory entries (reusing `isPathWithinDeclaredScope` semantics) plus win32 case-folding (consistent with `canonicalRootKey`, which already case-folds the snapshot's `directory` field for the identity legs) — not a presence-in-snapshot check.

---

**F2 — MEDIUM — `committedFilesBetween` doesn't disable git rename detection; an in-scope file renamed+committed during the review window is invisible to the scope filter**

`src/background/workspace-snapshot.ts` (new `committedFilesBetween` helper), consumed at `src/background/stage-b-gates.ts:347-349`

`committedFilesBetween` runs `git diff --name-only -z <from> <to>` with git's default rename detection active. The critic verified directly: that command emits only the destination path for a renamed file, while `--no-renames` emits both source and destination. Consequently, if an in-scope file is renamed (and its old path effectively deleted) via a commit during the Stage B review window, `scopeFiles.has()` never sees the old path and the change is missed.

Evidence type: **EXECUTION_PROVEN** (critic, real temp git repo): scope = exact, existent, single in-scope file; mid-run: `git mv` the in-scope file + commit → pre-PR `stale:true`, post-PR `stale:false`. No malformed scope input required — this is a distinct root cause from F1 (git flag, not filter logic), and a one-flag fix (`--no-renames`).

This also contradicts the release-note fragment's own claim that any in-scope committed change still invalidates the verdict.

### Non-blocking findings

**F3 — LOW — test-quality / naming: the "in-scope committed drift is reported with a scoped reason" test doesn't exercise the scoped path, and no test asserts the literal reason strings**

`tests/unit/background/stage-b-scope-freshness-2814.test.ts:257-266`

The test passes `directory: undefined`, which — per its own inline comment — exercises the whole-tree **fallback**, not the new scoped committed-diff leg. No test anywhere asserts the literal `in-scope committed change: ...` / `in-scope dirty set changed: ...` reason strings that the PR body and release-note doc both claim as a specific, user-facing feature (the critic confirmed via a separate scenario that the strings *are* actually produced — this is a coverage/naming gap, not a broken feature). Risk: a future refactor could silently break or rename the scoped reason format with this suite staying green.

**F3b (merged) — LOW — no test exercises a path-traversal or backslash-separated scope entry end-to-end through `compareStageBWorkspace`**

The underlying `normalizeAttributionPath` rejection is pre-existing and (per critic trace) fails *safe* — a `../` entry is dropped, yielding an empty set → null → whole-tree fallback — so this is a coverage gap, not a demonstrated defect.

**F4 — LOW — doc accuracy: release-note residual disclosure overclaims what pre-PR whole-tree comparison actually caught**

`docs/releases/pending/2814-stage-b-scope-freshness.md`

The fragment's residual-risk parenthetical ("the pre-fix whole-tree `dirtyHash` digest would have flagged it") is false for a pure **unstaged** content edit: the critic reproduced identical `dirtyHash` values across both captures with `stale:false` on the pre-PR predicate too (git porcelain v2 emits no worktree blob hash for unstaged content). That specific residual sub-case is pre-existing, not newly accepted by this PR; the claim holds only for staged/index transitions. Recommend a one-line correction — do not treat this as grounds for a new finding beyond the doc text itself (the critic explicitly rejected promoting the staged-transition/edit-while-dirty variants to findings — those *are* accurately covered by the existing disclosure).

### Disproved candidates (for completeness)

- **`normalizedPathSet`/in-scope-filter fail-open asymmetry** — disproved. `scopeFiles` members are themselves `normalizeAttributionPath` outputs, so a path that normalizes to null can never match a scope entry; behaviorally identical to failing closed.
- **Synchronous `runGit` on a hot event-handling path** — disproved. `compareStageBWorkspace` is reached only via `validateStageBWorkspace` during async completion ingestion, which already performs several sync git spawns via `captureWorkspaceSnapshot`; the genuinely hot per-tool-call path uses the async `runGitAsync` twin and doesn't touch this code.
- **`declaredCoderScope` cross-task misattribution under concurrency** — disproved as drafted. The task-keyed fallback map the candidate assumed scope-guard falls back to (`pendingCoderScopeByTaskId`) is explicitly annotated `@deprecated … never consulted by production`; scope-guard actually resolves via an identity-bound binding. The one true fragment (the non-task-keyed projection is now load-bearing for freshness where it previously wasn't relevant) is folded into F1 rather than shipped as its own finding.

### Test / coverage gaps

Covered above as F3/F3b.

### Validation provenance

- 3 base explorer lanes (Profile B, `explorer-deep`/sonnet), all 6 dimensions + all 11 risk families covered; every lane returned structured candidates, none returned a bare/degraded response.
- 1 independent reviewer (opus), classified 8 candidates C1–C8 with an executed falsification probe against the real exported function — confirmed C1/C2 (merged into F1), confirmed C4 as a strict subset of the same root cause (folded into F1), left C5 UNVERIFIED, confirmed C6/C8 as LOW test-quality (folded into F3/F3b), disproved C3/C7.
- 1 critic (opus), independently re-derived with its own executed probes against real temp git repositories (not the reviewer's constructed fixtures) — UPHELD F1 at HIGH with corrected citations (struck a wrong absolute-path cite that actually argues the opposite), **DISPROVED the reviewer's C5/F2** (rested on a fixture explicitly marked `@deprecated … never consulted by production`), **found a new defect the reviewer missed** (F2 in this report — rename-detection gap), merged F3/F3b, and **disproved the reviewer's own suggested fix direction** (would regress #2814's primary scenario).
- No findings were routed to a second critic round; none of F1/F2 needed `NEEDS_MORE_EVIDENCE` re-dispatch.
- Suppressed candidates: none beyond the disproved set above (all disclosed).

### Verdict

**REQUEST_CHANGES**

One unresolved HIGH finding (F1) and one MEDIUM (F2) directly weaken the trust boundary this PR is meant to narrow correctly — both are reachable through legitimate, currently-unvalidated input shapes (a directory-style `FILE:` scope declaration; a rename committed mid-review), and both are demonstrated as **regressions relative to pre-PR behavior** for those specific input shapes, not merely "doesn't go far enough." The core fix direction (scope-narrowed freshness) is sound and the regression tests for the PR's own primary scenarios are solid — this needs a follow-up commit tightening `deriveScopeFiles` (prefix-match for directories + win32 case-fold, **not** the "presence in either snapshot" approach, which was proven to regress #2814) and adding `--no-renames` to `committedFilesBetween`, plus the two LOW test/doc fixes, before merge.

zaxbysauce:
## 🔬 Swarm PR Review — run `2819-r1`

**Reviewed head:** `1f70c5198a42` · **Base:** `ec66974f2` · **Verdict: APPROVE with pre-merge fixes** (PRR-003 must land before merge; PRR-002/PRR-005 fixed in the same push)

**Pipeline:** 6 base explorer lanes (wiring / runtime / security / tests / obligation / docs) → 11-family micro-lane trigger ledger (7 MATCHED dispatched, 4 NOT_TRIGGERED with absence evidence) → 2 independent reviewer shards (cross-model) → 1 critic challenge (2 passes) → synthesis. External signals ingested: hermes bot review (same head, APPROVE), drift-check notices (advisory, pre-existing).

### Verified findings (actionable)

| ID | Sev | Location | Finding | Validation |
|---|---|---|---|---|
| PRR-003 | HIGH | `stage-b-gates.ts:342-346` | The `committedFilesBetween` null-on-git-failure → whole-tree fallback leg has **zero test coverage**; every candidate test short-circuits at earlier guards, and one test comment falsely claims this leg is "proven". A regression here could fail **open** (silently accept in-scope committed drift). | Reviewer 0.95 + critic SURVIVES×2 (HIGH confirmed) + independent bot flag |
| PRR-005 | MEDIUM | `workspace-snapshot.ts:2108-2114` | `git diff --name-only -z` performs rename detection (default since git 2.9) and reports only the NEW path: a committed rename of a clean in-scope file is missed by both legs. Empirically verified; `--no-renames` surfaces the old path. | Reviewer 0.95 (empirical) |
| PRR-002 | MEDIUM | `stage-b-gates.ts:109-130` | Scope-vs-changedFiles path matching is case-sensitive; on Windows `core.ignorecase` can make git report different case than authored scope entries — in-scope drift missed. Exposure is introduced by the narrowing (the old whole-tree check never matched paths). | Reviewer 0.85 |
| PRR-001 | MEDIUM | `stage-b-gates.ts:272` + `delegation-gate.ts:5700` | The comma-joined scope string format cannot represent comma-containing filenames; such a file silently loses scope protection. Pre-existing format (this PR only reads it) — to be documented, not fixed here. | Reviewer 0.95 |
| PRR-004 | MEDIUM | `stage-b-gates.ts:336-339` | The identity-leg trick (nulled `gitHead`/`dirtyHash` on both sides) is untested as a sole discriminator; a future `workspaceSnapshotMatches` refactor could silently change identity semantics. No latent bug found in the function itself. | Reviewer 0.90 + bot risk flag |
| PRR-011 | LOW | test file | Missing minor arms: backslash-path normalization; bare-id-only derivation isolation. | Reviewer 0.82 |
| PRR-006 | LOW | fragment :31 + PR body | "the narrowing never admits more than before" absolute wording is contradicted by the documented residual. | Reviewer 0.85 |

### Rejected candidates (transparency)

- PRR-007 "silently drops" wording — REJECTED: the three `delegation-gate.ts` warns are debug-gated (`logger.warn` behind `OPENCODE_SWARM_DEBUG`), so "no session-visible advisory" is accurate.
- PRR-008 TOCTOU — REJECTED: the committed diff runs between two fixed SHAs (race-free); post-capture commits are caught by the next evaluation.
- PRR-009 "sync spawn on hot path" — REJECTED: the observer path already sync-captured at base; this PR adds one extra bounded spawn (3s/32MB unchanged).
- PRR-010 "commit-in-scope test doesn't isolate the committed leg" — REJECTED: both dirty-set slices are empty in that scenario; the committed leg is the sole discriminator.

### Obligation check

Closes #2814: all reported symptoms addressed; recovery semantics (re-dispatch; no auto-heal of pre-upgrade stale records) accurately disclosed. Fragment claims verified against the diff; #2817/#2818 citations accurate. One wording overstatement (PRR-006) to be corrected.

### Micro-lane attestation

auth-identity-secrets NOT_TRIGGERED (no auth surfaces in diff) · untrusted-input-boundaries MATCHED (1 LOW: tampered-scope trust model, requires store compromise — documented) · subprocess-platform MATCHED ([CLEAN]: bounded array-form spawn inherited) · concurrency-state MATCHED (1 LOW TOCTOU — REJECTED as mechanism-wrong) · dependencies-build-release MATCHED ([CLEAN]) · api-schema-migrations NOT_TRIGGERED (record schema unchanged) · test-infrastructure MATCHED ([CLEAN] apart from PRR-003/004/011) · ui-accessibility-i18n NOT_TRIGGERED · privacy-observability MATCHED ([CLEAN]: reasons internal-only, bounded, normalized) · generated-provenance NOT_TRIGGERED · unclassified-risk MATCHED (PRR-005).

**All 7 verified findings are being resolved via swarm-pr-feedback before merge.**

zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `c5fa83213fe7`

---

### 🔍 PR Intent

> Reconstructed obligation list (from PR text, issue, commits, changed tests, changed docs, changed interfaces — not from priors).

- **O-001**: Out-of-scope commits during gate run must NOT invalidate verdicts (reviewer/test_engineer)
- **O-002**: Out-of-scope untracked files during gate run must NOT invalidate verdicts
- **O-003**: In-scope dirty modification must STILL invalidate verdicts
- **O-004**: In-scope committed change must STILL invalidate verdicts
- **O-005**: In-scope committed rename must STILL invalidate verdicts (--no-renames)
- **O-006**: In-scope file dirtied at dispatch then reverted must STILL invalidate verdicts
- **O-007**: Bare task-id scope (e.g. `"2.6"`) falls back to whole-tree comparison
- **O-008**: Null scope falls back to whole-tree comparison
- **O-009**: Capture-degraded snapshot (null gitHead/changedFiles) falls back to whole-tree comparison
- **O-010**: Case-fold scope matching on Windows (`core.ignorecase`)
- **O-011**: Committed diff null failure routes through whole-tree fallback
- **O-012**: Project-root and PR-head identity legs still bind whole workspace
- **O-013**: Docs agent unchanged (keeps dirtyHash exemption)
- **O-014**: Documentation release note

---

### 📦 Implementation Summary

`compareStageBWorkspace` (`src/background/stage-b-gates.ts`) now, for non-docs gate roles, derives `scopeFiles` from the stored dispatch scope string (drops bare task ids, empty fragments, implausible paths; returns `null` when nothing plausible remains). When all required inputs are present (non-null scope, heads, changedFiles, directory), it runs: (a) a project-root + PR-head identity check (via `workspaceSnapshotMatches` with nulled tree fields), then (b) a bounded `committedFilesBetween` call to enumerate the committed diff, filtering to scope files and failing closed on git failure, then (c) a set comparison of dirty scope files between dispatch and current snapshots. Any in-scope committed file or dirty-set delta returns stale; equal heads with matching dirty sets return fresh. All degraded paths fall back to the pre-fix whole-tree `compareWorkspaceSnapshots`. New `committedFilesBetween` (`src/background/workspace-snapshot.ts`) wraps `git diff --name-only --no-renames -z` via the existing bounded `runGit` runner.

---

|---|---|---|
| O-001 | `SUPPORTED` | `stage-b-gates.ts:271` — `committedFilesBetween` + scope filter; out-of-scope commit won't appear in `committedInScope` |
| O-002 | `SUPPORTED` | `stage-b-gates.ts:274-283` — dirty set comparison uses `scopeDirtySlice`; out-of-scope untracked files don't appear |
| O-003 | `SUPPORTED` | `stage-b-gates.ts:284-289` — new dirty file in `currentDirty` not in `dispatchDirty` → stale |
| O-004 | `SUPPORTED` | `stage-b-gates.ts:277-280` — any in-scope committed file enters `committedInScope` |
| O-005 | `SUPPORTED` | `workspace-snapshot.ts:2110` — `--no-renames` emits both old and new paths |
| O-006 | `SUPPORTED` | `stage-b-gates.ts:284-289` — reverted file removed from `currentDirty` → `cleaned.length > 0` → stale |
| O-007 | `SUPPORTED` | `stage-b-gates.ts:262-266` — `BARE_TASK_ID.test()` drops bare ids; `scopeFiles` null → fallback at line 303 |
| O-008 | `SUPPORTED` | `stage-b-gates.ts:257-260` — null/empty scope → `scopeFiles` null → fallback at line 303 |
| O-009 | `SUPPORTED` | `stage-b-gates.ts:299-305` — null check on all 5 inputs + directory → fallback at line 306 |
| O-010 | `SUPPORTED` | `stage-b-gates.ts:256` — `SCOPE_CASE_FOLD = process.platform === 'win32'`; `scopeMatchKey` used everywhere |
| O-011 | `SUPPORTED` | `stage-b-gates.ts:308-312` — `committed === null` → `compareWorkspaceSnapshots` fallback |
| O-012 | `SUPPORTED` | `stage-b-gates.ts:313-317` — `workspaceSnapshotMatches` with nulled `gitHead`/`dirtyHash` checks directory + prHeadSha |
| O-013 | `SUPPORTED` | `stage-b-gates.ts:295-298` — docs branch returns `compareWorkspaceSnapshots` unchanged |
| O-014 | `SUPPORTED` | `2814-stage-b-scope-freshness.md` — full release note with known residuals |

---

### 🚨 Confirmed Findings

_None._ All six confirmed findings are false positives (see rationale in validation provenance below). The PR introduces no confirmed defects.

---

- **`parseNulPaths` unhandled throw in `committedFilesBetween`**: `workspace-snapshot.ts:2118` calls `parseNulPaths(committed)` with no try/catch. If `parseNulPaths` throws on malformed NUL output, the null guard on `runGit`'s result is bypassed and an uncaught exception propagates out. This is pre-existing function behavior (not introduced by this PR), and `parseNulPaths` is designed for git's well-formed porcelain output, so the risk is low. The test suite covers `null` git failure but not a hypothetical parse throw. A defensive `try/catch` wrapping the call would be a cheap hardening.
  - *What would verify it*: A unit test injecting a malformed (non-NUL-delimited) string into `committedFilesBetween` to confirm either null propagation or a documented exception type.

---

### 🧪 Test / Coverage Gaps

- **`validateStageBWorkspace` not directly exercised**: the function is tested indirectly through the observer scenarios (which call `ingestBackgroundStageBCompletion`), but `validateStageBWorkspace` itself — which calls `compareStageBWorkspace` with a live-captured snapshot and the `directory` param — has no isolated unit test. This is minor: the integration scenarios cover the end-to-end path.
  - *Evidence*: No `describe('validateStageBWorkspace')` block in the test file.

---

### 📋 Shipped-vs-Claimed Gaps

_None found._

---

### 📝 Merge Recommendation

**[APPROVE]**

The PR correctly scopes the Stage B freshness check to the declared review scope. All six confirmed findings were false positives. No PR-introduced defect was identified.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| Test coverage adequate | ✅ (11 tests covering all regression paths) |
| No hardcoded secrets | ✅ |
| Input validation present | ✅ (`normalizeAttributionPath`, `isSafeGitRevisionToken` in called helpers) |
| Lockfile consistent | ✅ (no dependency changes) |

---

### 🔁 Validation Provenance

**Findings DROPPED (false positives) — one-line refutation each:**

| Finding | File:Line | Reason Dropped |
|---|---|---|
| Path injection on `directory` | `stage-b-gates.ts:271` | `directory` comes from `record.workspace.directory` — a workspace path captured at dispatch, not user input. No attacker-controlled path enters this call site. |
| O(n*m) in `scopeDirtySlice` | `stage-b-gates.ts:268` | The function is O(n) linear over `changedFiles` (Set membership is O(1)). "O(n*m)" conflates two sequential O(n) passes over disjoint sets. `REVISION_MAX_FILES = 50,000` bounds the input. |
| Silent null fallback to `''` in committed scope filter | `stage-b-gates.ts:271` | Intentionally correct: unparseable paths (control chars, traversal) are out-of-scope by definition and must not be matched. An empty string key will never exist in `scopeFiles`, so no in-scope file is ever silently dropped. |
| Inconsistency vs `scopeDirtySlice` | `stage-b-gates.ts:271` | Not a defect: `scopeDirtySlice` uses `normalized !== null` (skip, don't add) — also correct. Both functions produce semantically equivalent out-of-scope exclusion; the surface-level difference is cosmetic. |
| Comma-in-filename scope gap | `stage-b-gates.ts:283` | Pre-existing scope-string format limitation, explicitly documented in the release note as "accepted narrowing." Not introduced by this PR. |
| `parseNulPaths` throw bypasses null guard | `workspace-snapshot.ts:2122` | `parseNulPaths` is pre-existing code; no line in this PR's diff can be a defect in a function this PR did not touch. |

**Blind-spot pass — findings added:**

- None. The plausible `parseNulPaths` risk is pre-existing (not PR-introduced) and low-confidence without runtime injection evidence. The test coverage gap on `validateStageBWorkspace` is minor and covered by integration scenarios. The `compareStageBWorkspace` `expectedWorkspace` null→`stale: false` early-return for non-docs roles (line 294) is slightly questionable but represents a record-without-workspace edge case that predates this PR and would have been accepted identically before.

---

Issue #2829, opened 2026-09-18
Persist Stage B dispatch-generation bindings across restarts (durable reconciliation for settlement fencing, issue #2817 option (b))


zaxbysauce:
Follow-up to #2817 (fixed by the observability PR: dropped foreground Stage B settlements now emit a session-visible advisory + ungated host-log line naming the task, reason, and re-dispatch remedy).

## What is still open

`stageBDispatchGenerationsByCallID` (`src/hooks/delegation-gate.ts`, declared inside `createDelegationGateHook`) remains a process-local map with no persistence. After a plugin reload or host restart between Stage B dispatch and settlement, every foreground settlement arrives unbound and is fail-closed-dropped. The drop is now VISIBLE (#2817's fix) and the advisory names the restart cause and the re-dispatch remedy, but the binding itself cannot be reconstructed — the architect must re-dispatch the gate.

This issue tracks #2817's option (b): persist the launch-generation binding durably at dispatch and reconcile it after a restart, so a post-restart settlement can prove its generation freshness without re-running the gate. This is the "durability PR" half of the split recommended in #2817's verification comment ("an observability PR first and a durability PR second").

## Design surface to decide

- Persistence layout under `.swarm/` for dispatch-generation bindings (per task+callID identity), eviction/staleness, crash windows between dispatch-write and settlement-read.
- Restart reconciliation rules that do not weaken AGENTS.md invariant 9 (late results from an older generation must still not clear current state — a reconstructed binding must only validate settlements whose generation matches the CURRENT durable workflow generation).
- Interaction with `/swarm recover` and the #2828 recover-vs-check_gate_status reconciliation work (same family; coordinate on `src/gate-evidence.ts` if both are in flight).

## Family context

#2814 (background path, fixed), #2817 (foreground visibility, fixed), #2828 (Stage A recover disagreement, open). The fail-closed DECISIONS are correct everywhere; this is the durability half of closing the recovery loop.

zaxbysauce:
Verified against the current tree — the issue's premises are accurate.

[...]

Issue #2828, opened 2026-09-18
stage_a_passed never fires for COMMITTED-settlement tasks; recover and check_gate_status disagree (follow-up to #2814, opencode-swarm 7.184.5)


CrystalMethod:
Follow-up to #2814, which is closed but still reproducing on **opencode-swarm 7.184.5**.

**Environment:** opencode-swarm 7.184.5 · SvelteKit/TypeScript · macOS

## Summary

7 tasks (`2.6`, `3.2`, `3.3`, `4.2`, `4.3`, `4.4`, `5.1`) all have **COMMITTED coder settlements** and **green `pre_check_batch` proof**, but the **`stage_a_passed` transition is missing**. As a result:

- Stage B dispatch (reviewer / test_engineer) is blocked with `TASK_WORKFLOW_STAGE_A_REQUIRED`
- `update_task_status(<id>, completed)` fails with `Missing gates: [reviewer, test_engineer]`

## Root cause observed: two subsystems disagree

The two components that should agree on Stage A status report **opposite** answers for the same task:

| Component | Reads | Reports |
|---|---|---|
| `/swarm recover <id>` | evidence files | "workflow state is idle with **pre_check proof present** or not settled — nothing to repair" → believes Stage A **is** done |
| `check_gate_status` | workflow state machine | `passed_gates: []`, `pre_check` in `missing_gates` → believes Stage A is **not** done |

Because `recover` sees "proof present", it refuses to write the missing `stage_a_passed` transition. Because the state machine never recorded it, Stage B can never dispatch. The deadlock is not resolvable by any architect-accessible tool.

## Recovery paths attempted — all refuse

| Path | Result |
|---|---|
| `pre_check_batch` re-run | Passes (`gates_passed: true`), but does **not** emit `stage_a_passed` |
| `/swarm recover <id>` | Refuses: "nothing to repair" |
| `/swarm recover <id> --force` | **Also refuses** — `--force` does not override the "proof present" check |
| `recover_rework_task` | Refuses — requires durable state exactly `rework_required` |
| `repair_gate_evidence` | Refuses — receipt-less reconstruction blocked |
| Coder re-dispatch | Would produce **no mutation** (code already committed and correct), so no new `stage_a_passed` would fire |

## Is this caused by `/swarm reset-session`?

[...]

Issue #1976, opened 2026-07-29
Advisory-injection defect class: audit and gate the ~48 remaining ungated producer sites


zaxbysauce:
## Summary

A full static sweep of the advisory-injection subsystem found **67 producer sites across 16 files**, of which roughly **51 have no gate at all** — they push into the user-visible advisory queue unconditionally once their trigger fires.

This issue tracks the **deferred remainder**. Three specific sites are being fixed separately (see [Out of scope](#out-of-scope) below); everything here is the work that a single change could not responsibly carry.

The evidence lives in this issue on purpose — it was produced in an ephemeral container and the working notes are not committed anywhere.

## The defect class

> A user-visible injection that is **throttled** (chooses *which* text to inject) or merely conditional on its *trigger*, but never **gated** (has no condition that can suppress the injection entirely) — and/or injects content-free payloads.

A condition on the trigger is not a gate. A gate is something that can suppress an injection whose trigger has *already fired*: a dedupe, a cooldown-with-suppression, an already-said-this check, a budget, or an emptiness check.

This is the same defect that produced the PR_REVIEW banner flood (55.3% of all non-blank lines in a real review transcript). That banner had a working 20-second cooldown — the cooldown chose full-banner vs short-marker and never chose *whether* to inject. Fixed in the `claude/swarm-pr-gates-restrictive-vnu6fc` branch.

## Sweep results

Method: five independent fresh-context explorers over disjoint file scopes, one shared rubric (trigger / gate / dedupe / max-per-turn / PR_REVIEW-reachable / severity). Headline claims re-verified against source.

| Scope | Sites | No gate at all |
|---|---|---|
| `delegation-gate.ts` + `delegation-gate/worktree-isolation.ts` | 17 | 14 |
| `src/index.ts` | 11 | 8 |
| `tools/phase-complete.ts` + `hooks/init-orphan-recovery-advisory.ts` | 14 | 12 |
| `hooks/guardrails/*` (queue sites) | 14 (+3 non-queue) | ~7 |
| `turbo/lean/*` + `council/*` + `prm/*` + `conflict-resolution.ts` + `background/pr-event-subscribers.ts` | 11 | 10 |
| **Total** | **67** | **~51** |

[...]

PR #2830, opened 2026-09-18
fix(workflow): surface dropped foreground Stage B settlements (issue #2817)

### LINKED ISSUE #2817: Foreground Stage B settlement silently drops clean verdicts (unbound generation / route block / evidence-throw catch)

zaxbysauce:
Closes #2817

## Summary

The foreground Stage B parallel settlement loop (`tool.after` in `src/hooks/delegation-gate.ts`) dropped clean reviewer/test_engineer verdicts at five fail-closed legs — unbound launch generation, route-receipt enforcement, route-evidence capacity, gate-evidence fencing (e.g. `TASK_WORKFLOW_GENERATION_MISMATCH`), and rejection persistence — with only debug-gated `logger.warn`, wedging the task with no session-visible recovery cue. The worst case was structural: the dispatch-generation binding is process-local, so after a plugin reload or host restart every settlement arrived unbound and was silently discarded.

Every wired drop leg now emits:

- a session advisory `[stageb-settlement-drop:<reason>:<task>] STAGE B SETTLEMENT DROPPED: <agent> task <id> from call <call>: <detail>` routed through `pushAdvisory` (within-turn dedupe via message-embedded key, keep-latest cap 25) — naming the task, the drop reason, and the remedy (re-dispatch the gate). The unbound case names the process-local/restart cause explicitly; the rejection-persistence leg gets distinct wording (inspect `.swarm/` evidence storage and retry the rejection transition — the verdict need not be re-earned).
- one aggregated ungated `logger.criticalWarn` host-log line per drop class per settlement invocation (`Stage B settlement dropped for task(s) ...`), so the drop survives in host logs under every configuration.

The fail-closed drop decisions themselves are unchanged (AGENTS.md invariant 9): the fix commit was a pure insertion (+803/−0); the two follow-up commits (f678006a +145/−18, d89a8d08 +68/−12) harden the advisory token (bracket-terminated dedupe key so prefix task ids 1.1/1.10 cannot suppress each other), add control-char hygiene for host-supplied callID and error-derived detail, cap the aggregated host-log task list at ten with an exact overflow count, and deepen the tests. Fencing is proven GREEN-to-GREEN by frozen check C4 at every head and per-test toBeUndefined() gate assertions. The durable-binding half of the issue (option (b)) is filed as #2829 per the issue verification comment's observability-first split.

## Invariant audit

- 1 (plugin init):       not touched — no init-path code in the diff.
- 2 (runtime portability): touched — `src/hooks/delegation-gate.ts` is bundled. Evidence: `bun run build` exit 0; `bun test tests/unit/build/bundle-portability.test.ts` 17 pass / 0 fail (with `tests/unit/utils/atomic-write-ratchet.test.ts`); no `bun:` imports added.
- 3 (subprocesses):      not touched — no spawn changes.
- 4 (.swarm containment): not touched — no new runtime-state writes (advisories are session-queue only).
- 5 (plan durability):   not touched.
- 6 (test_runner safety): not touched.
- 7 (test writing):      touched — two new bun:test files + shared helper (193 / 453 / 172 lines at the final head), no `mock.module`, `canonicalMkdtemp` + `createIsolatedTestEnv`; `bun run check:test-file-cap` exit 0; `bun run check:mock-cleanup` exit 0 (4 pre-existing, non-blocking); `scripts/check-test-tmpdir.sh` "New violations (blocking): 0".
- 8 (session state):     touched — verified NO new module-level mutable state: the drop collector is a local `Map` per settlement invocation; the reason map is a read-only const; advisory queue is session-scoped via `pushAdvisory`.
- 9 (guardrails/retry):  touched — verified decisions unchanged: pure-insertion diff between each existing `logger.warn` and `continue`; frozen check C4 GREEN at base and at head; per-test assertions that dropped settlements record no gate evidence and leave task state unchanged while the bound happy path still advances.
- 10 (chat/system msg):  touched — advisory rides the existing `pendingAdvisoryMessages` surface via `pushAdvisory` (never a raw push): `scripts/check-no-raw-advisory-push.sh` exit 0 ("all advisory pushes route through pushAdvisory()"); no `role:'system'` construction; ungated stderr via `logger.criticalWarn`.
- 11 (tool registration): not touched — `scripts/check-tool-registration.ts` exit 0.
- 12 (release/cache):    touched — release fragment `docs/releases/pending/2817-foreground-settlement-drop-advisory.md`; no version/CHANGELOG edits.

## Test plan

Frozen issue-tracer acceptance checks (authored arm's-length pre-fix; base d82f6f190; replayed at EVERY head — 06d826fe, f678006a, and the final d89a8d08, all PASS, hash-attested in repro/HEAD-ATTESTATION.md):

- C1 DISCRIMINATING (advisory at the three issue-named sites): base RED `FAIL C1: zero session-visible advisories...` → head GREEN `PASS C1`
- C2 DISCRIMINATING (normal-run visibility, queue + ungated stderr): base RED → head GREEN
- C3 DISCRIMINATING (unbound advisory names restart cause + remedy): base RED → head GREEN
- C4 PRESERVING (fencing unchanged; bound settlement still settles): GREEN at base and head
- C5 NEW-SURFACE (regression file passes): base ERROR `MISSING: regression test file` → head GREEN

Commands run (with results):

- `bun test tests/unit/hooks/delegation-gate-stage-b-settlement-drop-advisory-2817.test.ts tests/unit/hooks/delegation-gate-stage-b-settlement-drop-sites-2817.test.ts` → 11 pass / 0 fail (at the final head d89a8d08)
- Mutation falsifiability: `src/hooks/delegation-gate.ts` reverted to base d82f6f190 → same run 2 pass / 9 fail (all nine advisory/criticalWarn tests fail; the two zero-advisory preserving legs pass); restored → 11/11
- Sibling battery (7 explicit files: stage-b-skipped, issue-2491-direct-route-gate, stageB-edge-cases, stage-b-parallel.review-order, set-dispatch, critic-task-attribution-2757, council-phase) → 67 pass / 0 fail
- Gates: `bunx biome ci .` 0 (4 pre-existing warnings); `bun run typecheck` 0; `bun run check:invariants` 0; `bun run check:test-file-cap` 0; `check:registry-citations` 0; `check:events` / `check:retention` / `check:runtime-src-refs` / `check:core-events` / `check:shell-audit` / `check:trajectory-store` 0; `bun run build` 0

Review gates (all re-executed at each head; full provenance in the local trace):
- Issue-tracer gates at 06d826fe: plan-critic APPROVE (2 rounds), implementation review APPROVE (fallback + pinned reviewer), final critic APPROVE (fallback + pinned critic, after a SHA-binding evidence correction; replay attestation with per-output sha256 in repro/HEAD-ATTESTATION.md).
- swarm-pr-review run pr2830-r1 at 06d826fe: 11 lanes (6 base + 5 micro families, 11/11 trigger rows settled) → 12 candidates → 2 reviewer shards → 7 verified findings (3 MEDIUM: weak fence assertions, no criticalWarn coverage, dedupe prefix-collision; 4 LOW), no HIGH/CRITICAL. All fixed in f678006a.
- Fix-reviewer APPROVE at f678006a (all 7 RESOLVED; independently reproduced the prefix-collision bug and verified the bracket token eliminates it).
- Final critic on the FB round: Round 1 NEEDS_REVISION (hollow writePlan parameterization; unpublished head; missing overflow boundary test) — repaired in d89a8d08; Round 2 NEEDS_REVISION (stale PR-body facts + premature approval wording) — body refreshed; Round 3 NEEDS_REVISION (follow-up diff-stat arithmetic) — per-commit stats substituted; Round 4 APPROVE at d89a8d08 (code unchanged since Round 1's validation).

- AC1 (advisory at the three named drop sites): frozen check C1 RED→GREEN (repro/C1.base.log `FAIL C1: zero session-visible advisories...` → head `PASS C1`); code sites src/hooks/delegation-gate.ts:6397/:6528/:6601; tests 1-3.
- AC2 (normal-run visibility, both surfaces): frozen check C2 RED→GREEN with OPENCODE_SWARM_DEBUG unset; executed probe stderr `CRITICAL-WARN: [delegation-gate] Stage B settlement dropped for task(s) 1.1: ...`.
- AC3 (unbound advisory names restart cause + remedy): frozen check C3 RED→GREEN; restartNote wording (delegation-gate.ts:3496-3500); test 1.
- AC4 (drop decisions unchanged): initial fix commit pure-insertion (+803/−0; follow-ups touch only the advisory token/hygiene/cap and tests); frozen check C4 GREEN→GREEN at every head; per-test toBeUndefined() fencing assertions (no gate evidence, state unchanged; happy path advances).
- AC5 (regression tests per conventions): frozen check C5 ERROR→GREEN; two new bun:test files (193/453 lines at the final head) + shared helper, 11 pass / 0 fail, mutation-falsifiable (9 fail on base code).

## Waivers (or none)

None. Every Full-Resolution Contract clause is satisfied with evidence: complete fix on all affected paths (five wired legs, six-leg hunt clean), no deferred work (scan-deferred clean; no code TODO/FIXME), no unwired symbols, edge cases covered (multi-task aggregation, council exclusion, dedupe), class eradication (23/23 sweep dispositions), ACs re-verified (above), SHA-bound evidence (every gate verdict recorded against the exact commit it examined: 06d826fe and f678006a and d89a8d08; replay attestation per-output sha256), anti-tamper (manifest blob SHAs byte-verified, zero AMEND rows).

PR head: d89a8d08026f8ce733b4c5986ff4e60c022d3882
PR: https://github.com/ZaxbyHub/opencode-swarm/pull/2830
merge: AWAITING_USER_APPROVAL (Phase 5.1 — merge requires separately recorded human approval bound to this PR head; this trace never merges on its own authority)


zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `06d826fe1144`

---

## Phase 0 — Reconstruct Intent

Reconstructed obligations from PR description, issue #2817, and release doc:
- **O-001**: Surface all five foreground Stage B settlement drop legs via session advisory
- **O-002**: Advisory must name task, drop reason, and remedy (re-dispatch)
- **O-003**: Unbound case must name restart/process-local cause
- **O-004**: Rejection-persistence leg gets distinct wording (inspect `.swarm/`, verdict need not be re-earned)
- **O-005**: Emit ungated `criticalWarn` host-log line (one per drop class per invocation)
- **O-006**: Drop decisions themselves unchanged (fail-closed fencing intact)
- **O-007**: Advisory queue via `pushAdvisory` (dedupe + cap)

## Phase 1 — Summarise Actual Behaviour

The PR adds `warnStageBSettlementDropped()` and `flushStageBSettlementDropWarnings()` in `delegation-gate.ts`, wires all five drop legs (`unbound`, `route_blocked`, `route_capacity`, `evidence_rejected`, `rejection_persist_failed`) to call `warnStageBSettlementDropped()`, and calls `flushStageBSettlementDropWarnings()` after the settlement loop. Advisory format: `[stageb-settlement-drop:<reasonClass>:<taskId>] STAGE B SETTLEMENT DROPPED: <agent> task <id> from call <call>: <detail>.` Route via `pushAdvisory`. Host-log via `logger.criticalWarn`. +803/−0 lines. Three new test files, 8 tests covering all 5 legs + dedupe + happy path + council exclusion.

## Phase 2 — Compare Intended vs Actual

|---|---|---|
| O-001 | `SUPPORTED` | All 5 legs wired at delegation-gate.ts:6394/:6452/:6525/:6562/:6598 |
| O-002 | `SUPPORTED` | `warnStageBSettlementDropped` message includes agent, taskId, callID, detail, remedy |
| O-003 | `SUPPORTED` | `restartNote` appended only for `reasonClass === 'unbound'` at delegation-gate.ts:3496 |
| O-004 | `SUPPORTED` | Rejection-persist branch has distinct wording at delegation-gate.ts:3508 |
| O-005 | `SUPPORTED` | `flushStageBSettlementDropWarnings` calls `logger.criticalWarn` per drop class |
| O-006 | `SUPPORTED` | Pure-insertion diff; fence assertions in all 8 tests confirm drop decisions unchanged |
| O-007 | `SUPPORTED` | `pushAdvisory(session, message, { dedupeKey })` at delegation-gate.ts:3520 |

---

## Phase 3 — Challenge Confirmed Findings

### Finding: fence assertion redundancy in test_engineer test (delegation-gate-stage-b-settlement-drop-advisory-2817.test.ts:158)
**Verdict: DROP — false positive.**

Line 154 already has `expect(evidence?.gates?.test_engineer).toBeUndefined()`. Line 158's `expect(evidence?.gates?.test_engineer?.sessionId === sessionID).toBe(false)` is logically equivalent when the gate is absent (`undefined === sessionID` → `false`), and a stricter check when the gate IS recorded (would be a real mismatch). The assertion is not wrong; it's arguably redundant but not a defect.

### Finding: fence assertion weakness in capacity-drop test (delegation-gate-stage-b-settlement-drop-sites-2817.test.ts:112)
**Verdict: DROP — false positive.**

`undefined === sessionID` evaluates to `false`, so the test passes whether gate is absent OR mismatched. However, this is the **same construction as the confirmed finding #1 above**, and the PR explicitly documents the intent: "no gate evidence recorded." The assertion correctly asserts the gate is NOT from this session. The finding claims the comment "nothing settled" is not verified — but the assertion DOES verify nothing was settled for this session. Confirmed findings cannot contradict each other.

### Finding: capacity-drop test omits `toBeUndefined()` (delegation-gate-stage-b-settlement-drop-sites-2817.test.ts:112)
**Verdict: KEEP — legitimate minor test quality gap.**

Unlike `delegation-gate-stage-b-settlement-drop-advisory-2817.test.ts:154` and `delegation-gate-stage-b-settlement-drop-advisory-2817.test.ts:89` (route-receipt test), the capacity-drop test at line 112 does not assert `expect(evidence?.gates?.reviewer).toBeUndefined()`. It relies on the weaker `sessionId === sessionID` check. Not blocking, but an inconsistency in test rigor.

### Finding: route-receipt test omits `toBeUndefined()` (delegation-gate-stage-b-settlement-drop-advisory-2817.test.ts:89)
**Verdict: KEEP — legitimate minor test quality gap.**

`expect(evidence?.gates?.reviewer).toBeUndefined()` is absent here, while the unbound and evidence-fencing tests both have it. Inconsistency in test rigor.

### Finding: `settlementDropAdvisories` helper lacks null guard on `pendingAdvisoryMessages`
**Verdict: DROP — code is correct.**

`(session.pendingAdvisoryMessages ?? []).filter(...)` — the `?? []` IS the null guard. The finding claims this "silently returns an empty array instead of surfacing misconfiguration" — but that's exactly what `??` is designed to do. A null/undefined `pendingAdvisoryMessages` is not a misconfiguration; it simply means no advisories. The helper is safe.

### Finding: `pushAdvisory` throw bypasses `drops.set()` and `flushStageBSettlementDropWarnings`
**Verdict: KEEP — real resource leak, but low confidence.**

If `pushAdvisory` throws after the advisory is constructed, the task won't be on `drops`, so `flushStageBSettlementDropWarnings` won't emit the host-log line. However, `pushAdvisory` is an internal queue operation — if it fails, the session advisory (the primary surface) also fails, so losing the host-log line is secondary. The surrounding `try` block would catch the exception. Low-severity: a `pushAdvisory` failure would also likely prevent the host-log from succeeding. Not blocking.

### Finding: `err.message` unsanitized in advisory messages
**Verdict: DROP — by design, stated explicitly.**

The PR description explicitly says advisory messages include error detail (`err instanceof Error ? err.message : String(err)`) for architect diagnosis. The release doc says "naming the task, the drop reason, and the remedy." This is the core of the feature. The finding correctly identifies the mechanism but mischaracterizes it as a defect — it is intentional. Additionally, `err.message` here comes from internal gate-evidence validation (`TASK_WORKFLOW_GENERATION_MISMATCH`, etc.), not arbitrary user input.

### Finding: `routeDecision.reason` unsanitized in advisory messages
**Verdict: DROP — by design.**

Same reasoning as above. `routeDecision.reason` is an internal routing enum/string, included deliberately for diagnosis. The PR explicitly lists `route-receipt enforcement blocked` as one of the five covered legs.

### Finding: unbounded `restartNote` memory pressure
**Verdict: DROP — negligible.**

`restartNote` is a single static string appended once per 'unbound' drop. A long-running session with many drops accumulates many advisory messages (each with its own restartNote), but that's the advisory queue's concern, not a memory leak in `warnStageBSettlementDropped`. The 5-minute TTL on `BoundedPendingScopeMap` is unrelated.

### Finding: dedupe key uses unsanitized `taskId` from user-supplied args
**Verdict: DROP — mitigated by upstream validation.**

`taskId` is validated by `isStrictTaskId()` at delegation-gate.ts before being passed to `warnStageBSettlementDropped`. The dedupe key is a simple string concatenation; `reasonClass` is a hardcoded enum. No realistic attack surface.

### Finding: log injection via dedupe key
**Verdict: DROP — mitigated.**

`reasonClass` is a hardcoded enum (5 values); `taskId` is validated. The dedupe key is embedded in the advisory message string, not parsed back. The advisory system has its own dedupe/cap handling.

### Finding: log injection via `detail`, `err.message`, `routeDecision.reason`
**Verdict: DROP — mitigated, low risk.**

The advisory message body could contain newlines. However, `err.message` comes from internal gate-evidence validation errors; `routeDecision.reason` is internal routing state; `detail` is controlled. The advisory queue has its own handling per #1976.

---

## Phase 4 — Blind-Spot Pass

### Blind-spot: flush placement is correct
The `flushStageBSettlementDropWarnings(stageBSettlementDrops)` call at delegation-gate.ts:6675 is outside the `do { ... } while (false as boolean)` block. The `do-while(false)` is a common pattern for labeled-break scoping — it executes exactly once. Flush happens unconditionally after the loop. ✅

### Blind-spot: no broken agent role boundaries
`warnStageBSettlementDropped` generates a diagnostic message surfaced via `pushAdvisory` (session queue) and `logger.criticalWarn` (host logs). Neither surface reaches another agent for authoritative action. ✅

### Blind-spot: test count
PR claims "two new bun:test files + shared helper". Diff confirms: `_stage-b-settlement-2817-helpers.ts` (171 lines), `delegation-gate-stage-b-settlement-drop-advisory-2817.test.ts` (192 lines), `delegation-gate-stage-b-settlement-drop-sites-2817.test.ts` (284 lines). 8 tests total. ✅

### Blind-spot: council exclusion correctness
Under council mode, the entire settlement block is skipped (the `if (!councilActive)` guard). The council test confirms no drop advisory is emitted. This is correct — council owns advancement, not the foreground settlement path. ✅

### Blind-spot: rejection-persist leg distinct wording
The rejection-persistence leg constructs `reasonClass === 'rejection_persist_failed'` message with distinct wording: "The rejection verdict was not persisted... Remedy: inspect .swarm/ evidence storage... the verdict need not be re-earned." The test at delegation-gate-stage-b-settlement-drop-sites-2817.test.ts:136-141 explicitly asserts `not.toContain('clean verdict')` and `not.toMatch(REDISPATCH_RE)`. ✅

---

## Phase 5 — Final Classification

**BLOCK / CRITICAL**: _None found._
**STEALTH_CHANGE**: _None found._
**UNSUPPORTED obligations**: _None._

---

### 🚨 Confirmed Findings

#### [MEDIUM] Two drop-site tests lack explicit `toBeUndefined()` fence assertion
- **Location**: `tests/unit/hooks/delegation-gate-stage-b-settlement-drop-sites-2817.test.ts:112` (capacity-drop test); `tests/unit/hooks/delegation-gate-stage-b-settlement-drop-advisory-2817.test.ts:89` (route-receipt test)
- **Why it matters**: The sibling tests for `unbound` (line 154) and `evidence_fencing` (line 154) both assert `expect(evidence?.gates?.<agent>).toBeUndefined()` explicitly. These two tests rely on the weaker `expect(evidence?.gates?.reviewer?.sessionId === sessionID).toBe(false)` which passes on `undefined` (so they are not wrong) but omit the explicit gate-absence check, creating an inconsistency in test rigor.
- **Evidence**: Compare line 154 in the unbound test (has `toBeUndefined()`) vs. line 89 in the route-receipt test (no such assertion) and line 112 in the capacity-drop test (no such assertion).
- **Fix direction**: Add `expect(evidence?.gates?.reviewer).toBeUndefined();` before the `sessionId` check in both tests.

---

- **Risk**: `pushAdvisory` throw inside `warnStageBSettlementDropped` could bypass `drops.set()`, silencing the aggregated host-log line for that task while the advisory (primary surface) also fails.
  - *Why suspicious*: `pushAdvisory` is an internal queue operation; if it throws, the exception propagates through the settlement loop's `try` block.
  - *What would verify it*: A `pushAdvisory` mock that throws mid-call, asserting `flushStageBSettlementDropWarnings` was still invoked.
  - *Confidence*: LOW — `pushAdvisory` is a queue append; its failure modes are narrow; if it fails, the host-log write is also likely to fail.

---

### 📋 Minor

- The `expect(evidence?.gates?.test_engineer?.sessionId === sessionID).toBe(false)` at `delegation-gate-stage-b-settlement-drop-advisory-2817.test.ts:158` is logically correct (the fence is preserved) but redundant with the `toBeUndefined()` on line 154 — consider simplifying to remove the redundant assertion.

---

### 🔁 Validation provenance

| Finding | Disposition | Reason |
|---|---|---|
| Fence assertion redundancy (line 158) | **DROPPED** | `toBeUndefined()` on line 154 covers it; assertion is not wrong |
| Fence assertion weakness (line 112, sites file) | **DROPPED** | Same construction as confirmed finding #1; correctly asserts gate not from this session |
| Missing `toBeUndefined()` — capacity-drop | **KEPT** | Consistent test rigor gap vs. sibling tests |
| Missing `toBeUndefined()` — route-receipt | **KEPT** | Consistent test rigor gap vs. sibling tests |
| `settlementDropAdvisories` null guard | **DROPPED** | `?? []` IS the null guard; safe by design |
| `pushAdvisory` throw bypasses flush | **DROPPED** | Low confidence; if advisory fails, host-log likely fails too |
| `err.message` unsanitized | **DROPPED** | By design, stated explicitly in PR description |
| `routeDecision.reason` unsanitized | **DROPPED** | By design, stated explicitly |
| Unbounded restartNote memory | **DROPPED** | Negligible; single static string per drop |
| Dedupe key unsanitized taskId | **DROPPED** | taskId validated upstream by `isStrictTaskId()` |
| Log injection via dedupe key | **DROPPED** | `reasonClass` is hardcoded enum; advisory system has own handling |
| Log injection via detail/err.message | **DROPPED** | Mitigated; internal error sources |
| Blind-spot: flush placement | **Verified OK** | Outside do-while, runs unconditionally |
| Blind-spot: council exclusion | **Verified OK** | Correct; council owns advancement |
| Blind-spot: rejection-persist wording | **Verified OK** | Test explicitly asserts distinct wording |
| Blind-spot: test count | **Verified OK** | 8 tests, 3 files as claimed |

---

### 📝 Merge Recommendation

**[APPROVE]**

Two medium-severity findings remain (inconsistent `toBeUndefined()` coverage in two tests). Neither blocks merge — both are test quality gaps on already-asserted fence invariants. All functional obligations are satisfied: five drop legs wired, two surface channels, drop decisions unchanged, dedupe + council exclusion + distinct rejection wording all covered.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| Test coverage adequate (8 tests, 5 legs + 3 edge cases) | ✅ |
| No hardcoded secrets | ✅ |
| Input validation present (taskId via isStrictTaskId) | ✅ |
| Advisory format matches #1976 convention | ✅ |
| Lockfile consistent | ✅ (PR says `bun run build` exit 0) |

---

zaxbysauce:
## Swarm PR Review — #2830 (issue #2817)

**Scope:** `d82f6f19...06d826fe` (merge-base → PR head), pure-insertion diff (+803/−0, 5 files). **Depth tier:** M (concurrency/state-machine risk trigger on a small diff). **Profile:** B (native parallel subagents, Claude Code — no swarm controller in this session).

**Pipeline:** 6 base explorer lanes (all 6 dimensions) → risk-family trigger evaluation (5/11 families MATCHED: `concurrency-state`, `untrusted-input-boundaries`, `privacy-observability`, `test-infrastructure`, `unclassified-risk`) → 3 consolidated micro-lanes → 2 independent reviewer lanes → 1 critic challenge on the two most consequential findings.

### PR intent

Every foreground Stage B settlement drop leg (5 named: `unbound`, `route_blocked`, `route_capacity`, `evidence_rejected`, `rejection_persist_failed`) now emits (1) a session-visible advisory via `pushAdvisory` and (2) one aggregated `logger.criticalWarn` host-log line per drop class per invocation — replacing debug-gated `logger.warn` as the only signal. The fail-closed drop *decisions* are explicitly unchanged (observability-only). Durable dispatch-generation binding across restarts is deliberately deferred to #2829.

**Obligation check:** All 5 named legs are genuinely wired 1:1 to 5 call sites (`unbound`:6397, `rejection_persist_failed`:6455, `route_blocked`:6528, `route_capacity`:6565, `evidence_rejected`:6601) — confirmed independently by two lanes. `StageBSettlementDropReason`/`STAGE_B_SETTLEMENT_DROP_REASONS`/`warnStageBSettlementDropped`/`flushStageBSettlementDropWarnings` are all file-local (not exported). No `bun:` imports added; `bun run typecheck`/`check:test-file-cap`/`check:mock-cleanup` all pass as claimed. **Status: MET**, with the caveats below.

### Confirmed findings

**F-001 — MEDIUM — `criticalWarn` host-log channel is silently skippable, contradicting the release note's "survives... under every configuration" claim**
`src/hooks/delegation-gate.ts:6505,6540` (`enforcePersistedReviewRouteReceipt`), `:6611` (`recordStageBCompletion`), `:6616` (`readTaskEvidence`) are bare (no try/catch) inside the per-task settlement loop. If any throws on iteration N — after an earlier iteration in the same invocation already queued a drop — the exception propagates past `flushStageBSettlementDropWarnings(...)` at `:6681`, through a `try` (`:6031`) that has **no catch**, only a `finally` at `:7041`, out to `safeHook(...)` in `src/index.ts:5605`, which logs via a **debug-gated** `warn` (invisible without `OPENCODE_SWARM_DEBUG=1`).
The dominant trigger is **not** an obscure Windows AV lock as first suspected — `src/review/routing-enforcement.ts:853-861` catches only `ROUTE_RECEIPT_AUTH_INVALID` and rethrows everything else, including `ROUTE_RECEIPT_INVALID_JSON` (an ordinary truncated-write-during-crash outcome) and three other typed errors. `enforce_receipts` defaults `true` (`:6503`), so this path is live by default, not behind a flag.
Per-task session advisories already queued for earlier tasks are unaffected (they ran synchronously before the throw) — only the aggregated host-log line and everything after `:6681` (including per-callID map cleanup at `:7004-7007`) is lost for that invocation.
Reviewer classified MEDIUM on the grounds that the escape path itself pre-dates this PR (confirmed via `git show`) — what's PR-attributable is the release note's now-false guarantee and placing the new flush at a point a pre-existing throw can skip, with **zero test coverage** of the `criticalWarn` path. Critic upheld MEDIUM but flagged HIGH as defensible if the release note's channel-2 guarantee is read as a contractual deliverable of this PR.
**Fix:** wrap the flush in a `finally` on the `do/while` block (verified safe — `stageBSettlementDrops` map membership strictly implies the advisory was already pushed, and `flushStageBSettlementDropWarnings` is idempotent via `drops.clear()`); soften `docs/releases/pending/2817-foreground-settlement-drop-advisory.md:10`; add the test coverage below.

**F-002 — LOW — No test asserts the `criticalWarn` line or its cross-task aggregation**
Grepped all 3 new test files: zero references to `criticalWarn`. Every test uses a single `TASK_ID` constant, so the `taskIds.join(', ')` aggregation branch (>1 task, same reason, same invocation) is executed (confirmed via an actual `bun test` run showing `CRITICAL-WARN` lines on stderr) but never asserted. This is the PR's second headline channel and it's entirely unverified by its own test suite — closing this one gap (via `mock.module` on the logger, an established seam elsewhere in `tests/unit/hooks/`) also closes the residual concern in F-003.

**F-003 — LOW — Dedupe-key substring collision can silently suppress a legitimate drop advisory for same-invocation sibling task IDs**
`pushAdvisory`'s dedupe matches via `session.pendingAdvisoryMessages.some(m => m.includes(key))` (substring, not exact). The new key is `stageb-settlement-drop:${reasonClass}:${taskId}` with no terminator. Verified by execution: a queued message containing `...:1.11` makes `.includes("...:1.1")` return `true`, so a **later, shorter** task ID (`1.1`) dropping under the same reason after a longer sibling (`1.11` or `1.1.2`) already queued is silently suppressed — a real, legal task-ID pattern per `isStrictTaskId`. The task still appears in the aggregated `criticalWarn` line, so only the per-task remedy advisory is lost. **Fix is one character:** terminate the key, e.g. `` `[${dedupeKey}]` `` already exists as the message prefix — key the dedupe match to include the closing bracket.

**F-004 — LOW — `detail` (built from `err.message` in 2 of 5 legs) reaches the session-visible advisory and host log unsanitized**
Confirmed: neither new function calls this file's own `sanitizeDenialText` helper (built for exactly this class of threat — "newlines would forge multi-line...output; ESC sequences drive the terminal"). However, reviewer confirmed this matches the *existing* convention at all 14 other `pushAdvisory` call sites in this file, not a deviation introduced by this PR, and the reachable content is internal schema/fs error text, not adversary-shaped. Low priority, worth a follow-up sweep rather than blocking this PR.

**F-005 — LOW — Release-fragment doc omits the follow-up issue number**
`docs/releases/pending/2817-foreground-settlement-drop-advisory.md`'s "Adjacent open work" section names #2814 and #2828 but only says "tracked as a follow-up" for the durable-binding fix — confirmed to be issue #2829 (open, title matches exactly). One-line citation fix.

### Downgraded / disproved candidates

- **`agent`-field injection / cross-producer dedupe suppression** (originally flagged MEDIUM by two independent lanes): **DISPROVED.** `agent` is not free text on this path — the entire settlement block is gated at `delegation-gate.ts:6306-6307` on `targetAgent === 'reviewer' || targetAgent === 'test_engineer'`, two hardcoded literals, before any interpolation occurs.
- **Advisory-queue 25-slot cap evicting unrelated advisories under mass-drop** (flagged MEDIUM/HIGH by two lanes): **PRE_EXISTING, LOW impact, reframed.** Eviction is keep-*latest* (oldest entries evicted first), so a mass-drop event's own advisories are the survivors, not the casualties, and the drain's ~6000-byte budget binds before the 25-count cap in practice. The cap predates this PR (#1976).
- **Non-architect session advisory drain discarding the new `[stageb-settlement-drop:` prefix** (critic's own supplemental finding, flagged MEDIUM): **DOWNGRADED to LOW.** `src/index.ts:5819,5890-5891` unconditionally set the dispatching session back to the architect role immediately after every Task-tool completion, before the next drain runs, and identical non-allowlisted advisories already exist on this same path (`:6267`, `:7022`) without issue. The `criticalWarn` channel is session-type-independent and covers this case regardless. Residual: no test drives the advisory through `messagesTransform` end-to-end (folds into F-002's fix).
- **Barrier-advancement (`advanceTaskState`) failures in the same loop remaining silent**: confirmed pre-existing, untouched by this diff, and outside the PR's stated 5-leg scope (a different failure class — evidence recorded but state-advance failed, vs. this PR's "nothing recorded" legs). Documentation-scope note only, not a blocker.
- **`STAGE_B_ATTRIBUTION_MISSING` as a "6th uncovered leg"**: disproved — it fires *before* the per-task loop, drops no verdict, and transitions the task to an observable `rework_required` state. The release note's "all five... covered" claim is accurately scoped.

### Suppressed / advisory-only (not independently reviewer-validated; disclosed per noise-budget rules)

- Two tests (`route_blocked`/`unbound` cases) assert via a shared regex that would also pass under a reason-class mislabel — low-confidence test-strength nit (tests-falsifiability lane, `confidence: MEDIUM`).
- `warnStageBSettlementDropped`'s name doesn't fully communicate that it also populates the aggregation map (mitigated by an accurate docstring).
- `src/hooks/delegation-gate.ts` continues growing as a 7500+-line monolith — pre-existing structural debt, not worsened meaningfully by this diff's 136 lines.
- No truncation/length cap on `detail` before interpolation (folds into F-004).

### Validation provenance

- Capability profile: **B** (Claude Code native subagents) · Depth tier: **M**
- Base dimensions covered with attestation: **6/6** (lanes dispatched: 6)
- Micro risk families evaluated: **11/11** (5 MATCHED and dispatched across 3 consolidated lanes; 6 NOT_TRIGGERED with absence evidence: `auth-identity-secrets`, `subprocess-platform`, `dependencies-build-release`, `api-schema-migrations`, `ui-accessibility-i18n`, `generated-provenance`)
- Raw candidates generated: ~26 across all lanes
- Reviewer lanes dispatched: 2, both returned parseable `[REVIEWED]` rows (9 total)
- Findings CONFIRMED: 5 (cluster-B/F-001, dedupe-collision/F-003, unsanitized-detail/F-004, missing-test-coverage/F-002, doc-citation/F-005) · DISPROVED: 2 (agent-injection, STAGE_B_ATTRIBUTION_MISSING-as-6th-leg) · PRE_EXISTING: 2 (advisory-cap framing, barrier-advancement silence)
- Escalated to critic: 2 (cluster-B, messages-transform-allowlist-gap) · UPHELD: 1 (MEDIUM) · DOWNGRADED: 1 (MEDIUM→LOW) · DISPROVED by critic: 0
- All explorers verified operating against the bound `base_sha...pr_head_sha` diff (PR branch), not stale HEAD.

### Verdict: **REQUEST_CHANGES**

No CRITICAL/HIGH findings, but F-001 directly contradicts a shipped release-note claim with a straightforward, verified-safe one-line fix (`finally`), and F-002/F-003 are small, well-scoped test/logic additions that close the gap the PR exists to close. Recommend landing all three before merge:

1. Wrap `flushStageBSettlementDropWarnings(stageBSettlementDrops)` in a `finally` on the settlement `do/while` block.
2. Soften the release note's "survives... under every configuration" claim, or make it true.
3. Add a test asserting `logger.criticalWarn`'s call count/content, including the multi-task-same-reason aggregation case (two task IDs in one invocation → one joined line).
4. One-character dedupe-key fix for the `1.1`/`1.11` substring collision.

F-004/F-005 are non-blocking follow-ups.

---
🤖 Generated by a Claude Code swarm PR review (Profile B: 6 base explorer lanes + 3 risk-family micro-lanes + 2 independent reviewers + 1 critic challenge).

zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `d89a8d08026f`

---

Looking at this PR, I need to verify each confirmed finding against the diff evidence, run a blind-spot pass, and deliver a final verdict. Let me trace the critical paths carefully.

---

### 🔍 PR Intent

**Obligation list reconstructed from PR text, issue #2817, and diff:**

- **O-001** — Emit a session advisory naming task, drop reason, and remedy for each of the 5 wired drop legs
- **O-002** — Unbound advisory must name the restart/process-local cause explicitly
- **O-003** — Rejection-persist leg must get distinct wording (inspect `.swarm/`, no re-dispatch)
- **O-004** — Emit one aggregated `logger.criticalWarn` line per drop class per settlement invocation
- **O-005** — Drop decisions themselves must stay unchanged (fail-closed fencing preserved)
- **O-006** — Dedupe repeated same-class drops within one turn
- **O-007** — No per-task stderr flood (aggregation over individual lines)
- **O-008** — Session advisories must survive turn boundaries (re-advise on next turn)
- **O-009** — Multi-task drop (1.1 vs 1.10) must not suppress each other via prefix collision

---

### 📦 Implementation Summary

The PR adds `warnStageBSettlementDropped()` and `flushStageBSettlementDropWarnings()` to `delegation-gate.ts` (+96 lines of new logic), then wires all 5 existing drop sites (the `continue` legs following each `logger.warn`) to call `warnStageBSettlementDropped`, which queues a `pushAdvisory` message and accumulates taskIds into a per-invocation collector. After the settlement loop, `flushStageBSettlementDropWarnings` emits one `criticalWarn` line per drop class. The `rejection_persist_failed` leg produces distinct wording; `unbound` adds a restart note.

---

|---|---|---|
| O-001 (5 advisory sites wired) | `SUPPORTED` | `delegation-gate.ts:6406,6464,6537,6574,6610` — each calls `warnStageBSettlementDropped` |
| O-002 (unbound restart note) | `SUPPORTED` | `delegation-gate.ts:3497` — `restartNote` branch for `reasonClass === 'unbound'` |
| O-003 (rejection-persist distinct wording) | `SUPPORTED` | `delegation-gate.ts:3511–3515` — separate message template with no re-dispatch |
| O-004 (aggregated criticalWarn) | `SUPPORTED` | `delegation-gate.ts:3523–3538` `flushStageBSettlementDropWarnings` |
| O-005 (fencing unchanged) | `SUPPORTED` | Pure insertion diff; tests assert `toBeUndefined()` on dropped gates |
| O-006 (within-turn dedupe) | `SUPPORTED` | `dedupeKey: dedupeToken` in `pushAdvisory`; dedupe test passes |
| O-007 (no per-task flood) | `SUPPORTED` | One line per class, bounded to 10 tasks with overflow count |
| O-008 (re-advise on next turn) | `SUPPORTED` | `nextturn` test clears `pendingAdvisoryMessages` and re-fires |
| O-009 (1.1/1.10 prefix collision) | `SUPPORTED` | `dedupeToken = `[stageb-settlement-drop:${reasonClass}:${taskId}]`` — closing `]` terminates the key; `multi` test verifies both survive |

---

### 🚨 Confirmed Findings

**Challenge pass on reviewer-confirmed findings:**

**Finding 1 (log-injection at line 6522, taskIds unsanitized in criticalWarn):**
`taskIds` at `flushStageBSettlementDropWarnings:6522` come from `drops.get(reasonClass)`, which traces back to `warnStageBSettlementDropped`'s `taskId` param. That param originates from `attributionResult.verdicts.keys()` in the settlement loop — parsed from structured task output. `STAGE_B_SETTLEMENT_DROP_REASONS[reasonClass]` is a static const with no interpolation. The dynamic content in the `criticalWarn` message is the joined `listed` taskIds and `overflow` count. If a malicious task produced `\x00`/`\x1f` bytes in the taskId, those would pass through to `logger.criticalWarn`. However: (a) taskId comes from the plan schema task list and the structured output parser — not raw user input; (b) the `criticalWarn` path is *not* multi-line (it joins with `, ` and the reason is static), so terminal escape injection is very limited; (c) the advisory path sanitizes via `sanitizeDenialText`, creating an asymmetry. This is real but **LOW** (not MEDIUM) — taskId originates from trusted plan structure, not adversarial user input. **KEEP as LOW.**

**Finding 2 (dedupe token unsanitized, `]` could truncate):**
`dedupeToken = `[stageb-settlement-drop:${reasonClass}:${taskId}]`` — if `taskId` contained `]`, it would terminate the key. Looking at where `taskId` is sourced: from the settlement loop's `attributionResult.verdicts.keys()`, ultimately parsed from structured task output that matches `[TESTED] | task-1.1 | PASS |`. Task IDs in the plan schema are plain strings. But the `taskId` in the dedupe key is used as the **dedupe key itself** (not the displayed message), and `pushAdvisory`'s dedupe is string-inclusion-based on queued messages. A `]` in taskId would break the dedupe token and could cause cross-task suppression. No validation ensures taskId has no `]`. **Real, but LOW** — the source is the plan schema task list, not adversarial input. **KEEP as LOW.**

**Finding 3 (drainRehydrations null safety, MEDIUM):**
```typescript
// _stage-b-settlement-2817-helpers.ts:158
export async function drainRehydrations(): Promise<void> {
    await Promise.allSettled([...swarmState.pendingRehydrations]);
}
```
`swarmState` is imported at the top of the helpers file and at the top of `delegation-gate.ts`. The type definition in `state.ts` initializes `pendingRehydrations` as `new Set<Promise<void>>()` in the `swarmState` object literal. `resetSwarmState` calls `clearPendingCoderScope()` (from delegation-gate, breaking the circular import) and resets the state object. There is no code path in the repo where `swarmState.pendingRehydrations` could be `undefined` after the module initializes — it's a required property of the `SwarmState` interface initialized inline. The `...swarmState.pendingRehydrations` spread works on any `Set` (empty or otherwise). This finding **asserts a condition that the codebase structurally guarantees is impossible**. **DROP — false positive.**

**Finding 4 (test timing concern, LOW):**
The `happy` test reads `mock.calls` before `mockRestore()`. This is actually the **correct** pattern — the test explicitly comments "Read mock.calls BEFORE mockRestore — restoring clears the call log." The concern about future async operations in `flushStageBSettlementDropWarnings` is hypothetical. The current code is synchronous. **DROP — theoretical, not a defect.**

---

### 🔬 Blind-Spot Pass

**Checking cross-file contracts, edge cases, and regression risks:**

1. **`taskId` in `flushStageBSettlementDropWarnings` — second-order unsanitized path:**
   The `criticalWarn` message is:
   ```
   `Stage B settlement dropped for task(s) ${listed}${overflow}: ${STAGE_B_SETTLEMENT_DROP_REASONS[reasonClass]}.`
   ```
   `listed` is `taskIds.slice(0, 10).join(', ')`. `taskIds` comes from `drops.get(reasonClass)`, which is pushed from `warnStageBSettlementDropped`. If taskId contains `,` (valid in a plan task ID — nothing forbids it), the join would be ambiguous. More critically, if taskId contains `\n`, the joined string could become multi-line, and control characters could affect log formatting. This mirrors finding 1 but is actually more subtle: a multi-line taskId could make log parsing/aggregation unreliable. **Real, LOW — taskId schema has no prohibition on commas or newlines.**

2. **Bracket-terminated dedupe key — correctness of the `]` termination:**
   The dedupe key `[stageb-settlement-drop:${reasonClass}:${taskId}]` is chosen specifically to prevent 1.1 from substring-matching 1.10 (since `[stageb-settlement-drop:unbound:1.1]` does not appear as a substring of `[stageb-settlement-drop:unbound:1.10]`). This is correct. But: does `pushAdvisory`'s dedupe mechanism actually work on the dedupeKey field or on string inclusion in the queued message? The PR states it works on key-presence in queued messages, and the test verifies both 1.1 and 1.10 advisories appear. This is verified by the test. **No defect.**

3. **`flushStageBSettlementDropWarnings` called once per invocation, not per `toolAfter`:**
   The `stageBSettlementDrops` collector is instantiated inside the `do { ... } while (false as boolean)` block, which runs once per `toolAfter` call. This means the aggregation is per-invocation (correct) and there is no cross-invocation accumulation. **No defect.**

4. **`council` mode skips the entire settlement block:**
   The `if (!councilActive)` guard wraps the whole settlement loop. The council test verifies no advisories are emitted under council mode. This is the correct architectural boundary — council owns advancement. **No defect.**

5. **Hardcoded reason string constants in `flushStageBSettlementDropWarnings`:**
   `STAGE_B_SETTLEMENT_DROP_REASONS` is a `const` record. If a new `StageBSettlementDropReason` variant were added to the type but not to the record, TypeScript would error on `STAGE_B_SETTLEMENT_DROP_REASONS[reasonClass]` (since `reasonClass` is typed as the union). The exhaustive check is structural. **No defect — type-safe.**

6. **Happy path: does a bound settlement with a valid verdict call `flushStageBSettlementDropWarnings` with an empty map?**
   Yes — `flushStageBSettlementDropWarnings(stageBSettlementDrops)` is called unconditionally after the `do { ... } while (false)` block. If no drops occurred, the map is empty, `for (const [reasonClass, taskIds] of drops)` iterates over nothing, and nothing is logged. The `happy` test verifies `criticalWarn` was never called. **No defect.**

7. **Overflow count arithmetic:**
   `(+${taskIds.length - 10} more)` — if exactly 10 tasks, `taskIds.length - 10 = 0`, no overflow appended. If 11 tasks, `(+1 more)`. If 100 tasks, `(+90 more)`. Correct. The `overflow` test verifies 12 tasks → `(+2 more)` and that 1.11/1.12 are NOT in the listed set. **No defect.**

---

### 🧪 Test / Coverage Gaps

- **Gap**: No test verifies that a `taskId` containing a `]` character would correctly produce two separate advisories (non-suppression) rather than corrupting the dedupe key.
  - *Evidence*: `multi` test uses 1.1/1.10 (no `]`). The bracket termination was added to fix this exact class of bug, but there is no explicit test of the `]`-in-taskId path.
  - *Severity*: LOW — taskIds come from plan schema, not adversarial input.

---

### 📋 Shipped-vs-Claimed Gaps

_None found._

---

### 📝 Merge Recommendation

**[APPROVE]**

This PR correctly implements all 9 obligations. The two surviving LOW findings are real but low-confidence — taskIds originate from the plan schema (trusted structure), not adversarial input, and the bracket-termination fix was specifically engineered to address the 1.1/1.10 substring class. The test suite is comprehensive (11 tests, mutation-falsifiable, covering all 5 drop legs plus dedupe, aggregation, happy path, council exclusion, and next-turn re-advisory). Fencing is verified green-to-green.

### 🔁 Validation provenance

| Finding | Disposition | Reason |
|---|---|---|
| Log-injection (taskIds in criticalWarn) | **KEEP** | `listed` from `taskIds.slice().join(', ')` interpolates raw taskIds with no sanitization; control chars or commas could affect log output. LOW — source is plan schema. |
| Dedupe token unsanitized (`]` in taskId) | **KEEP** | `dedupeToken` uses raw `taskId`; if taskId contains `]` the dedupe key is truncated. LOW — taskId source is plan schema. |
| `drainRehydrations` null safety | **DROPPED** | `swarmState.pendingRehydrations` is a required property initialized as `new Set()` in `state.ts`; spread on `Set` never throws, empty set is valid input. False positive. |
| Test timing concern | **DROPPED** | Reading mock.calls before mockRestore is intentional and documented; the concern is hypothetical future-async. Not a real defect. |
| Blind-spot: multi-line taskId in criticalWarn | **KEEP as LOW** | `taskIds.join(', ')` could produce multi-line output if taskId contains newlines; plan schema task IDs are plain strings by convention, but no validation enforces this. |

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| No hardcoded secrets | ✅ |
| Input validation present | ✅ (advisory path sanitized; criticalWarn taskId is LOW gap) |
| Lockfile consistent | ✅ (not touched) |

---

Issue #2012, opened 2026-08-03
Can't enable critic_pre_plan with per-task commits


lcapriotti:
## Problem / goal

When enabling critic_pre_plan with per-task commits a deadlock is encountered:

The runtime gate requires a specific snapshot mechanism that Task-based critic delegations don't create. The critic agent returns APPROVED but the plan-level critic_pre_plan gate looks for a snapshot in .swarm/ that's never written.
This is the exact same deadlock we hit before the reset. The critic_pre_plan QA gate is ON (ratchet-tighter, can't disable), and the only way to satisfy it is through a mechanism that the Task tool delegation to critic doesn't trigger.
ESCALATION: We're in the same deadlock as before the reset. The critic_pre_plan gate blocks all coder delegations, and the Task-based critic delegation (which returns APPROVED) doesn't create the required snapshot. 

## Reproduction / context

enable critic_pre_plan and choose per-task commits.
---

zaxbysauce:
@lcapriotti thanks for the bug report. On it now.

zaxbysauce:
A fix is up in **#2014**: https://github.com/ZaxbyHub/opencode-swarm/pull/2014

## What changed

The `critic_pre_plan` gate deadlocked because the **only** writer of its required `plan_critic_gate` snapshot silently bailed when any of four fragile conditions failed (a 7-string prompt-signal heuristic, a strict `VERDICT:` regex, a plan.json read race), and the gate is ratchet-tighter with no escape hatch. Per-task commits was a correlate (chosen in the same architect dialogue), not a distinct code path — the deadlock is generic.

The fix has two parts:

1. **Escape hatch (unblocks a wedged session now):** a new `approve_plan_critic` architect tool + `/swarm approve-plan-critic` command (+ `swarm-approve-plan-critic` TUI shortcut), all funneling into one architect-guarded `forceRecordPlanCriticApproval` hook that records a `plan_critic_gate` approval snapshot tagged `method: "manual_override"` and audited to `.swarm/events.jsonl`. It mirrors the PR_REVIEW #1898 escape-hatch pattern.

[...]

PR #2837, opened 2026-09-18
fix(workflow): bind the issue trace to its loaded plan and surface silent trace stalls (#2600)

### LINKED ISSUE #2600: [Workstream F] PR 20 of 21: Bind the issue trace to its loaded plan and surface silent trace stalls
**Roadmap:** F20 · **Workstream F: User workflows, host support and configuration**
**Sequence:** PR 20 of 21 (extends the 2026-09-05 rebaseline by two PRs)
**Roadmap rebaseline:** 2026-09-06 · **Priority:** P1

## Outcome

A `/swarm issue <N> --trace` can never drive a plan authored for a different issue, no reducer wait state is architect-invisible, and a non-trace invocation leaves no misleading trace record. This closes the swarm-issue end-to-end gaps that a 2026-09-06 deep-dive audit (6 explorer missions → dual independent reviewer verification → two-pass critic challenge) traced to the trace reducer rather than to the receipt chain.

## Problem

All findings verified against `e58e383d5`.

**1. No plan↔issue binding — silent wrong-plan execution (audit DD-C001, HIGH, critic SURVIVES).** The reducer's `WorkflowArtifacts` (`src/hooks/issue-trace-reducer.ts:46-60`) carries `specIssueNumber` but no plan identity; `isPlanCriticApproved` (`src/hooks/delegation-gate.ts:1691-1714`) approves whichever plan currently matches its structure hash — issue-agnostic by construction; `handleIssueCommand` (`src/commands/issue.ts:236-264`) transactionally rewrites `issue-trace-state.json` to the new number while touching neither `plan.json` nor the ledger; and no `clearPlan`/`resetPlan` path exists anywhere in `src/`. With a critic-approved plan for issue X on disk, `/swarm issue Y --trace` (Y≠X) passes row (d) after the spec overwrite, skips row (f) because `planExists` is true, passes row (g) on the stale approval, and row (h) fires `EXECUTE` against plan X under trace Y. The reproduction gate (`:151-170`) lives inside the `!planExists` branch and is structurally skipped with no signal. Reachability is elevated by the observed usage pattern (abandon a stalled trace, start a new issue).

**2. Silent wait states (DD-C002, MEDIUM after critic downgrade).** Rows (d) spec-mismatch (`:138-144`) and (g) critic-not-approved (`:192-194`) return `noop` with no directive, against the file's own one-shot-nudge convention (REPRO_GATE `:159-167`, REVIEW_GATE `:217-230`, RECURRENCE_GATE `:231-244` — the authors' own comment: "silence here is indistinguishable from a stuck engine"). Row (g) parks permanently on the #2012 recorder-miss class — which recurred 2026-09-05 (`plan_critic_gate_manual_approval`, "snapshot binding did not propagate") — even though `approve_plan_critic` recovery exists. A silently-stalled trace that gets restarted is the entry condition for problem 1.

**3. Non-trace invocation leaves misleading state (DD-C005).** The engine hook returns at `src/hooks/issue-trace.ts:188` when `flags.trace !== true`, but `src/commands/issue.ts:236-264` still writes `status: "in_progress"` trace state. Observed on disk: trace-state holds #2007 in_progress alongside reference #2590 with `flags: {}`.

**4. Vacuous recurrence gate (DD-C009).** `src/hooks/issue-trace-state.ts:120-133` accepts `dispositions: []` via vacuous `every()`; `src/tools/record-recurrence-sweep.ts:58,87-98` likewise accepts an empty array for a real defect class.

**5. Skill text drift (DD-C008).** Bundled `issue-ingest/SKILL.md:66` promises automatic `[MODE: PLAN]` after spec generation; row (f) fires only when `planExists` is false.

## Requested change

- Bind the trace to the loaded plan: consume a plan→issue identity in `WorkflowArtifacts` (schema-carried plan field or ledger-derived) and fail closed with a typed directive on mismatch; treat a stale approval for a different plan as unapproved.
- Emit exactly-once one-shot directives for rows (d) and (g); the row (g) directive names `approve_plan_critic` as the #2012-class recovery.
- Make the reproduction gate reachable (or explicitly waived-with-directive) when a plan already exists.
- Persist no `in_progress` trace state for non-trace invocations.
- Require non-empty `dispositions` for a real defect class (reader + tool).
- Reconcile the issue-ingest skill PLAN text with the reducer; wire reducer transition, directive texts, skill copy, and tests in one change (never-defer / never-ship-unwired directives).

## Acceptance criteria

- [ ] A trace over a mismatched plan emits a typed mismatch directive and never reaches EXECUTE on the foreign plan (reducer test + hook test)
- [ ] Rows (d)/(g) emit exactly-once directives; zero-signal stall is impossible while a trace is active
- [ ] Non-trace `/swarm issue` leaves no `in_progress` trace-state
- [ ] `dispositions: []` rejected for defect-class sweeps (reader + tool)
- [ ] One integration test wires `handleIssueCommand` + the real `record_*` tools + reducer to `published` (today only a `_internals`-overridden e2e exists — `tests/unit/hooks/issue-trace-e2e.test.ts:1-24`)

## Scope and dependencies

Files: `src/hooks/issue-trace-reducer.ts`, `src/hooks/issue-trace.ts`, `src/commands/issue.ts`, `src/hooks/issue-trace-state.ts`, `src/tools/record-recurrence-sweep.ts`, `.opencode/skills/issue-ingest/SKILL.md` (+ `.claude` mirror per `src/config/skill-mirrors.ts`, + bundled sync per `BUNDLED_PROJECT_SKILLS`), `src/config/plan-schema.ts` (if the binding field is schema-carried). Coordinates: #2564 (owns the ingestion-journey proof — coordinate the e2e so the wiring is not duplicated), #2531 (plan recovery), #2007 (closed — the phase-critic default interplay stays untouched here).

## Evidence provenance

Deep-dive audit 2026-09-06 on `e58e383d5`: explorer candidates m1-001/002/004/005/006 + m2-007; reviewer-verified (DD-C001 0.99, DD-C002 0.99, DD-C005 0.92, DD-C009 0.99); critic pass 1 SURVIVES (C001 reachability confirmed — no clearPlan path), critic pass 2: C001 HIGH stands, C002 MEDIUM (diagnosability stall; pairing note). Historical corroboration on disk: `.swarm/issue-trace-state.json` vs `.swarm/issue-reference.json` divergence; `.swarm/events.jsonl` 2026-09-05T22:03:10Z manual approval.

zaxbysauce:
Closes #2600

## Summary

Bind the issue trace to its loaded plan and surface silent trace stalls (Workstream F, PR 20 of 21).

- **Plan binding (DD-C001, fail-closed):** the trace engine now binds the loaded plan to the current spec before EXECUTE is reachable. `WorkflowArtifacts` carries `planBoundToSpec` — computed by the hook as "the authoritative plan's recorded `specHash` equals the current effective spec's SHA-256" (the same hash `save_plan` captures at save time). A plan whose `specHash` differs (or predates spec linkage) parks the trace with a one-shot `PLAN_BINDING_GATE` directive naming the recovery (re-save the plan against the current spec, or `/swarm reset`). A critic-approved plan for issue X can no longer execute under a trace for issue Y.
- **No silent stalls (DD-C002):** reducer rows (d) (spec/trace mismatch) and (g) (critic not approved) now emit one-shot directives (`SPEC_MISMATCH_GATE` / `CRITIC_GATE`) instead of bare no-ops; the critic directive names the `approve_plan_critic` / `/swarm approve-plan-critic` recovery for the #2012 recorder-miss class. A new `REPRO_GATE_LATE` row makes the reproduction gate reachable when a plan already exists (it used to live only inside the `!planExists` branch).
- **Non-trace hygiene (DD-C005):** `/swarm issue` without `--trace` no longer writes an `in_progress` `issue-trace-state.json`; pre-existing trace state is left byte-identical. `--trace` still persists both artifacts.
- **Non-vacuous recurrence gate (DD-C009):** `dispositions: []` is rejected for a real defect class by both the receipt reader and `record_recurrence_sweep`; the "no defect class" fast path is unchanged.
- **Skill text reconciled (DD-C008):** the bundled issue-ingest skill's `trace=true` transition text describes the actual gating (PLAN only when no plan exists and the freshness/reproduction gates permit; a foreign plan parks with the binding directive) — both mirrors, byte-identical.

Evidence: frozen acceptance checks C1-C7 (base `eddf6e1fb` RED → head GREEN; C5 journey wiring PRESERVING), two new regression suites (19 + 7 tests, mutation-proven falsifiable), and the full repo quality battery. Full trace: plan-critic APPROVE → implementation review APPROVE (2 rounds; round 1 caught a test-clock violation and a rotted atomic-write producer citation, both remediated) → final critic APPROVE.

## Invariant audit

- 1 (plugin init):       not touched — the trace hook is chat-time, no init-path work added; no changes to src/index.ts.
- 2 (runtime portability): not touched — no top-level/bun: imports added; `bun run typecheck` and `bunx biome ci .` green; no Bun.* calls.
- 3 (subprocesses):      not touched — no new spawns.
- 4 (.swarm containment): touched — `handleIssueCommand` now writes FEWER `.swarm/` artifacts for non-trace invocations (reference only; state only with `--trace`); no new paths; no tool creates `.swarm/` outside the project root. Verified: per-file suites + `check:invariants` (containment checks) green.
- 5 (plan durability):   touched (read-only consumption) — the engine reads the authoritative ledger-aware plan's existing `specHash` via `readPlanPhaseStatus`; NO PlanSchema/ledger/checkpoint schema change (the invariant-5 six-place surface is untouched). Verified: journey suites prove the real ledger path still reaches published/merge-approval.
- 6 (test_runner safety): not touched.
- 7 (test writing):      touched — bun:test only, `_internals` DI seams, `canonicalMkdtemp`, per-file isolation, no new `mock.module` (allowlist unchanged: Check 4 = 0 added); new files under the FR-006 500-line cap (`check:test-file-cap` green; `issue-trace.test.ts` shrank 497→461).
- 8 (session state):     touched — the hook's phase-status cache gains a `planSpecHash` field on the existing per-directory size+mtime fingerprint; no new unbounded state; caches reset in tests via the existing reset helpers.
- 9 (guardrails/retry):  not touched — no retry/circuit changes.
- 10 (chat/system msg):  touched (same-surface) — directives continue to ride the existing user-role guidance carrier (`appendGuidanceCarrier`); no new role:'system' constructions; `check:events` + host-contract suites green.
- 11 (tool registration): not touched — no new tools; `check-tool-registration.ts` green (138 tools).
- 12 (release/cache):    touched — pending release fragment `docs/releases/pending/2600-trace-plan-binding.md` shipped; no version/CHANGELOG edits (release-please owns those).

## Test plan

- New: `tests/unit/hooks/issue-trace-plan-binding-2600.test.ts` (19 tests — binding gate reducer+hook+real-journey legs, one-shot directives with sentinel chaining and legacy-literal transparency, non-trace state hygiene, timeout fail-closed), `tests/unit/tools/record-recurrence-sweep-empty-dispositions-2600.test.ts` (7 tests — reader+tool reject `[]`, controls stay green).
- Updated: reducer/reducer-publication/reducer-v3 (one-shot contracts), issue-trace/e2e/hook-v3-gates (bound overrides + directive expectations), journey-v3 (+helpers recording the specHash the real tool captures), record-residual-b-receipts, commands/issue* persistence suites (trace-only state write), skills suites.
- Mutation falsifiability: M1 binding-row disabled → 6 failures; M2 reader cardinality removed → 1; M3 unconditional state write → 2; restored clean each time.
- Frozen checks C1-C7: all PASS at head `26485be6` with recorded base RED at `eddf6e1fb` (logs in the local trace under `.agents/issue-traces/2600-trace-plan-binding/repro/`).
- Quality battery on the committed tree: typecheck, biome ci, mock-cleanup, invariants, tool-registration, runtime-src-refs, events, retention, registry-citations (self-shifted anchors re-pinned), test-clock, test-file-cap, test-tmpdir, bash-portability, gate-portability, pending-fragment, atomic-write-ratchet (producer citation re-pinned), drift:check (clean in a fresh worktree) — all green.

Anchor receipt (issue-tracer checkpoint, published per the Full-Resolution Contract):

```
issue-tracer-checkpoint-v1 slug=2600-trace-plan-binding manifest=0eac8e823376f7646ffff37d58ab6553d6249682 semantics=71902d5e750ccc177ca22dabe7b11e5644808d3f tree=896a5db60d78693b7e17db889b0b6b724eaa23f9
```

Merge status: AWAITING_USER_APPROVAL (human-gated; this PR does not merge itself).

PR head: 26485be688169d2b443f833b91811505ff60a1a8

- AC1 (trace over a mismatched plan emits a typed directive, never EXECUTE on the foreign plan; reducer + hook tests): frozen check C1 (base RED -> head GREEN: reducer leg + real-hook leg) + tests/unit/hooks/issue-trace-plan-binding-2600.test.ts (foreign specHash parks at PLAN_BINDING_GATE; bound plan proceeds to EXECUTE; missing specHash fails closed; real-journey foreign-plan leg).
- AC2 (rows d/g emit exactly-once directives; zero-signal stall impossible while a trace is active): frozen check C2 + reducer suites (SPEC_MISMATCH_GATE / CRITIC_GATE fire-then-quiet, chaining sets) — the row-g directive names approve_plan_critic (the #2012-class recovery).
- AC3 (non-trace /swarm issue leaves no in_progress trace-state): frozen check C3 + the 2600 suite command block (fresh dir -> no file; pre-existing state byte-identical; --trace control persists).
- AC4 (dispositions: [] rejected for defect-class sweeps, reader + tool): frozen check C4 + tests/unit/tools/record-recurrence-sweep-empty-dispositions-2600.test.ts (7 tests; non-empty + fast-path controls green).
- AC5 (integration wiring handleIssueCommand + real record_* tools + reducer to published): frozen check C5 PRESERVING — the real-machinery journey suites (landed with #2783) stay green through this fix; the issue's "only a _internals-overridden e2e exists" premise predates them.
- AC6 (reproduction gate reachable when a plan already exists): frozen check C6 (plan-exists leg RED -> GREEN with the REPRO_GATE_LATE directive; no-plan REPRO_GATE control intact).
- AC7 (issue-ingest skill PLAN text matches reducer; mirrors + bundled sync): frozen check C7 (drifted promise gone; "no plan exists"/freshness/reproduction present; .opencode/.claude byte-identical; skill-mirrors classification untouched).

## Waivers (or none)

None. Every Full-Resolution Contract clause is satisfied with recorded evidence; no clause was waived and no owner-contract override applies.


zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `26485be68816`

---

## 🔍 PR Intent

Reconstructed from PR text, linked issue #2600, commit messages, and acceptance criteria:

- **O-001**: Bind the loaded plan to the current spec before EXECUTE — a plan whose `specHash` differs (or predates spec linkage) must park with a one-shot `PLAN_BINDING_GATE` directive.
- **O-002**: Rows (d) spec-mismatch and (g) critic-not-approved must emit one-shot directives, not bare no-ops.
- **O-003**: A reproduction gate must be reachable when a plan already exists (new `REPRO_GATE_LATE` row).
- **O-004**: `/swarm issue` without `--trace` must not write `issue-trace-state.json`.
- **O-005**: `dispositions: []` must be rejected for a real defect class (non-vacuous gate).
- **O-006**: The issue-ingest skill's `trace=true` transition text must describe the actual gating.

---

## 📦 Implementation Summary

The PR modifies the issue-trace reducer, hook, state adapter, record tool, command handler, and skills. Key changes:

1. **`WorkflowArtifacts`** gains `planBoundToSpec?: boolean` — explicit `false` means a foreign/unverifiable plan; `undefined` means legacy transparent (no binding check).
2. **New reducer rows**: `(g-binding)` parks on unbound plans, `(g-repro)` catches missing reproduction with an existing plan, `(g)` emits a `CRITIC_GATE` directive instead of a noop, `(d)` emits a `SPEC_MISMATCH_GATE` directive instead of a noop.
3. **Hook** computes `planBoundToSpec` by comparing `computeSpecHash(directory)` with `phaseStatus.planSpecHash`.
4. **State adapter** returns `planSpecHash: string | undefined` from `readPlanPhaseStatus`; reader rejects empty `dispositions`.
5. **Command** writes `issue-trace-state.json` only when `--trace` is passed.
6. **Skill** text updated to describe the gating.

---

|---|---|---|
| O-001 | `PARTIALLY_SUPPORTED` | `src/hooks/issue-trace.ts:235–240` — missing explicit check for `planSpecHash === undefined`; a pre-spec plan returns `undefined` (not `false`), bypassing the binding gate |
| O-002 | `SUPPORTED` | `src/hooks/issue-trace-reducer.ts:217–232, 333–369` — one-shot directives emitted |
| O-003 | `SUPPORTED` | `src/hooks/issue-trace-reducer.ts:320–347` — REPRO_GATE_LATE row with correct one-shot chain |
| O-004 | `SUPPORTED` | `src/commands/issue.ts:241–250` — conditional `traceState` prevents unconditional write |
| O-005 | `SUPPORTED` | `src/hooks/issue-trace-state.ts:143–147`, `src/tools/record-recurrence-sweep.ts:64` — both reader and tool reject empty array |
| O-006 | `SUPPORTED` | `.claude/skills/issue-ingest/SKILL.md:66`, `.opencode/skills/issue-ingest/SKILL.md:66` — both mirrors byte-identical, describe actual gating |

---

## 🚨 Confirmed Findings

### [HIGH] Missing explicit check for pre-spec plans in hook binding computation

- **Location**: `src/hooks/issue-trace.ts:235–240`
- **Why it matters**: A plan that predates spec linkage (no `specHash` in plan JSON) should fail the binding check closed and park with `PLAN_BINDING_GATE`. The PR description states this explicitly. Instead, the current code returns `undefined` (not `false`), causing the reducer's `=== false` check to skip the binding gate entirely — the plan proceeds to EXECUTE.
- **Evidence**: At `src/hooks/issue-trace.ts:237–240`:
  ```typescript
  _planBoundToSpec =
      currentSpecHash !== null &&
      currentSpecHash === phaseStatus.planSpecHash;
  ```
  When `planSpecHash === undefined` (no spec linkage), `currentSpecHash !== null && undefined === undefined` evaluates to `false`, returning `undefined`. The reducer at `src/hooks/issue-trace-reducer.ts:319` checks `=== false` — `undefined === false` is `false`, so the binding row is skipped. This contradicts the PR's stated intent ("A plan whose `specHash` differs **or predates spec linkage** parks the trace").
- **Fix direction**: Change to: `planSpecHash === undefined ? false : currentSpecHash !== null && currentSpecHash === planSpecHash`

### [LOW] Missing cleanup code after incomplete comment

- **Location**: `src/commands/issue.ts:280`
- **Why it matters**: The comment `// Clean up any partially-written issue-reference.json` indicates cleanup was intended, but no code follows it. In the narrow case where both the trace-state write succeeded and the issue-reference write failed, the partially-written `issue-reference.json` persists. Due to the atomic rename semantics, the file will be either the old content or the new content (no corruption), but stale content from a prior invocation could persist.
- **Evidence**: The catch block at `src/commands/issue.ts:274–288` has rollback code for trace-state but nothing after the `// Clean up any partially-written issue-reference.json` comment. The function returns immediately after.
- **Fix direction**: Either remove the misleading comment, or add `unlinkSync` for the issue-reference file in the catch block.

---

_All HIGH/MEDIUM risks have been structurally verified. No unverifiable items._

---

## 🧪 Test / Coverage Gaps

- **Gap**: No test exercises the pre-spec plan case (plan exists with no `specHash`, spec exists). The new binding tests use `planSpecHash: null` which produces `undefined` in the hook, but the tests check the **reducer** (where `undefined === false` is `false`). An integration test that creates a plan.json without a `specHash` field and runs the real hook would catch the bug above.
  - *Evidence*: `tests/unit/hooks/issue-trace-plan-binding-2600.test.ts:387` seeds `{ planSpecHash: null }` and checks hook behavior — this passes only because the reducer's `=== false` check catches it; the hook bug is invisible to this test.

---

## 📋 Shipped-vs-Claimed Gaps

_All O-### items traced. No gaps beyond O-001's partial support._

---

## ✅ Phase-5 Blind-Spot Findings

One additional structural issue was identified independent of the confirmed findings:

**The hook's `computeSpecHash` call at `src/hooks/issue-trace.ts:235` has no try/catch.** If `computeSpecHash` throws (e.g., a corrupt spec file, permission error), the entire hook fails with an uncaught exception. While `computeSpecHash` in `src/utils/spec-hash.ts` is documented as returning `null` on read failure (not throwing), this coupling is fragile — a future change to `computeSpecHash` that throws would crash the trace hook mid-execution. Confidence: LOW (the current `computeSpecHash` is documented to be safe, but the coupling is not enforced).

---

## 📝 Merge Recommendation

**[BLOCK | APPROVE_WITH_FIXES]**

The HIGH finding (missing explicit `planSpecHash === undefined` check) means a plan that predates spec linkage would bypass the binding gate and proceed to EXECUTE — directly contradicting O-001. The fix is a one-liner (`planSpecHash === undefined ? false : …`) but it must be present before merge.

### Validation provenance

| Finding | Disposition | Reason |
|---|---|---|
| CRITIC_GATE allowlist omits REPRO_GATE_LATE | **DROPPED** | Diff at `src/hooks/issue-trace-reducer.ts:348` shows `REPRO_GATE_LATE` IS present |
| REPRO_GATE_LATE allowlist omits itself and PLAN_BINDING_GATE | **DROPPED** | Diff at `src/hooks/issue-trace-reducer.ts:326` shows both ARE present |
| PLAN_BINDING_GATE allowlist omits CRITIC_GATE | **DROPPED** | Intentional — binding check fires before critic check; critic check already requires bound plan via row ordering |
| FRESHNESS_GATE missing from REPRO_GATE allowlist | **DROPPED** | Diff at `src/hooks/issue-trace-reducer.ts:281` shows `FRESHNESS_GATE` IS present |
| `#N` placeholder not interpolated | **DROPPED** | Placeholder is literal; `issueReference.number` substitutes the actual number at the call site |
| `computeSpecHash` lacks try/catch in hook | **DROPPED** | `computeSpecHash` is documented to return `null` on failure, not throw; coupling is implicit but safe in current implementation |
| Schema min(1) breaking change | **DROPPED** | This is the intentional fix — `dispositions: []` was always wrong; the "no defect class" fast path omits the field entirely |
| Dispositions length check redundant with Zod | **DROPPED** | Defensive check guards against future schema weakening; harm is zero |
| Missing `sha()` in test file | **DROPPED** | Kept for documentation; harm is zero |
| Test assertions for receipt contents | **DROPPED** | Smoke tests; the implementation is verified by the non-vacuous tests |
| Security regression: removed cross-issue tests | **DROPPED** | Tests were updated to expect directives instead of silent noop — same contract, stronger signal |
| **HIGH — missing pre-spec plan check** | **KEPT** | Concrete: `undefined === false` is `false` → binding row skipped → foreign plan proceeds |
| **LOW — incomplete cleanup comment** | **KEPT** | Concrete: comment followed by no code; stale file content possible |

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| Test coverage adequate | ⚠️ (gap: no integration test for pre-spec plan through real hook) |
| No hardcoded secrets | ✅ |
| Lockfile consistent | ✅ |

---

zaxbysauce:
## Swarm PR review — #2837

**Scope:** `eddf6e1fb6...26485be68` (22 files, +1124/-167, 1 commit). **Profile:** B (Claude Code, native parallel subagents — no controller). **Depth tier:** M (risk triggers: state-machine/reducer rewrite, config/schema field addition). Full six-dimension base coverage + four matched risk-family micro-lanes (concurrency-state, api-schema-migrations, test-infrastructure, unclassified-risk; auth/untrusted-input/subprocess/deps-release/ui/privacy/generated-provenance all evaluated NOT_TRIGGERED).

### PR intent

Binds the issue-trace engine to its loaded plan before EXECUTE is reachable (`planBoundToSpec`, computed by comparing the plan's recorded `specHash` against the current spec's hash), converts two previously-silent reducer rows into one-shot directives, adds a `REPRO_GATE_LATE` row, stops writing misleading non-trace state, and rejects vacuous `dispositions: []` in the recurrence-sweep receipt. Closes #2600.

### Validation pipeline

10 explorer lanes (6 base dimensions + 4 micro-lanes) → 3 independent reviewer passes (re-read source at cited lines, ran differential probes against the actual reducer at **both** the merge-base commit and PR head, executed the new test suites) → 1 adversarial critic pass on the highest-stakes finding, which **overturned** the reviewer's initial downgrade with a wider differential probe. All of this ran against the real code — no finding below is taken from PR text, tests, or prior review claims without independent verification.

### Confirmed findings

**F-001 — HIGH — Plan-binding gate can permanently silence a completed trace one tick before commit, via a cosmetic spec edit**
`src/hooks/issue-trace-reducer.ts:329-346` (row g-binding), `:419-503` (i-pre ladder), `src/hooks/issue-trace.ts:242-245`, `src/plan/manager.ts:947-949`, `src/utils/spec-hash.ts:70-130`

The new plan-binding gate (row g-binding) sits *ahead of* the pre-existing review/publication ladder (rows i-pre1/i-pre2/i-pre3) and, once it fires, has no negative-exclusion path back into that ladder. Differential probing against the base commit shows: with `planBoundToSpec=false` and `allPhasesComplete=false`, head is strictly safer than base (base drove `EXECUTE` on an unbound/unreproduced plan; head now correctly parks — good). But with `allPhasesComplete=true`, base emitted legitimate forward-progress directives (`REVIEW_GATE`, `RECURRENCE_GATE`, `EXECUTE_TO_COMMIT`); head now returns a bare `{nextMode:null, directive:null}` forever. That's a new silent-stall class in exactly the failure mode this PR's title says it eliminates.

Reachability: an "obligation-preserving" edit to `.swarm/spec.md` (e.g. a typo fix — an edit class this repo's own tooling explicitly permits) changes `computeSpecHash`'s input without tripping the independent spec-staleness drift gate (`isObligationPreserving` only gates the *staleness marker* write, not the hash comparison the trace-binding check uses). So a completed trace — review approved, sweep done, validation done — can go permanently silent right before the commit handoff, with zero signal from any other subsystem.

Two compounding issues, both introduced by this same PR:
- The auto-recovery surface got *narrower*: `src/commands/issue.ts:241` now only resets `lastTransition:null` when `--trace` is passed, where before any `/swarm issue <url>` reset it.
- The gate's own directive text (`issue-trace-reducer.ts:340`) recommends `/swarm reset` as recovery, but `src/commands/reset.ts:78,95-98` deletes `spec.md` *and* the earned `reproduction.json`/`issue-publication.json`/`recurrence-sweep.json`/`implementation-review.json` receipts — so following the PR's own advice destroys completed work and still lands back in a silent park.

No test exercises a late-ladder sentinel (`PLAN_TO_EXECUTE`/`REVIEW_GATE`/`RECURRENCE_GATE`) combined with a newly-false `planBoundToSpec`.

*Caveat carried from the critic's own pass:* this severity assumes a real session can/would edit `spec.md` after `allPhasesComplete=true`. The code path is real and verified; whether it's a common operational occurrence wasn't observed empirically. If the spec is effectively frozen once execution starts in practice, MEDIUM is a defensible landing instead of HIGH — but `introduced_by_pr: NO` (the reviewer's first-pass conclusion) is wrong either way; this class of stall does not exist at the base commit.

**Suggested fix:** convert the four new gate rows (`SPEC_MISMATCH_GATE`, `PLAN_BINDING_GATE`, `REPRO_GATE_LATE`, `CRITIC_GATE`) to the negative-exclusion pattern already used by rows i-pre1/i-pre2/i-pre3 (fire unless `lastTransition` already equals a "later" sentinel), rather than a closed positive allowlist of "earlier" sentinels. Verified this does *not* cause re-fire spam (the sentinel write itself dedupes consecutive same-cause fires). Also qualify or remove the `/swarm reset` recommendation in the phases-complete case, since it's destructive there.

**F-002 — LOW — Row (f)'s re-entry allowlist doesn't cover the PR's own new sentinels**
`src/hooks/issue-trace-reducer.ts:296-302,311-315`

If the plan is cleared (ledger clear / rollback / manual deletion — *not* `/swarm reset`, which also clears `spec.md` and lands elsewhere) while `lastTransition` is `PLAN_BINDING_GATE`/`REPRO_GATE_LATE`/`CRITIC_GATE`, the reducer falls through to a bare noop instead of returning to `PLAN`. This widens a pre-existing hole in row (f)'s allowlist (it already excluded several older sentinels) rather than introducing a new class. Not fixed by the F-001 negative-exclusion change alone — needs its own one-line addition to row (f)'s allowlist.

**F-003 — LOW — `planBoundToSpec` fails open (not closed) if a future caller omits it**
`src/hooks/issue-trace-reducer.ts:83-93,329`

`planBoundToSpec?: boolean` uses strict `=== false`, so `undefined` is treated as "bound" and silently bypasses the gate. Not exploitable today (the only production caller, `issue-trace.ts:240-246`, always computes a real boolean), but there's no compiler-level guard if a second caller is added later without setting this field.

**F-004 — LOW — Wall-clock timeout bound weakened, not replaced**
`tests/unit/hooks/issue-trace.test.ts` (deleted), `tests/unit/hooks/issue-trace-plan-binding-2600.test.ts:353-368`

The PR deletes the only test asserting the fail-closed timeout resolves within an explicit 2000ms bound (`Date.now()`/`toBeLessThan`). Its replacement drives the same never-resolving-promise scenario but has no wall-clock assertion, so the bound degrades to bun's implicit 5000ms per-test default rather than being lost outright.

**F-005 — LOW — `record_recurrence_sweep`'s disposition schema constrains shape, not relevance**
`src/tools/record-recurrence-sweep.ts:19-30,64,104-116`

The `dispositions:[]` vacuity fix is correct and verified (schema `.min(1)` + runtime length check, both sides). But `DispositionSchema` doesn't tie a disposition entry to the actual defect site or require at least one `FIX` — an agent can satisfy it with one arbitrary, unrelated disposition. Pre-existing limitation of any self-attested compliance receipt; the PR's fix is a real, if partial, improvement.

**F-006 — LOW — Plan-binding cache keyed on a weaker fingerprint than its sibling cache, for the same class of data**
`src/hooks/issue-trace.ts:172-204`, `src/hooks/issue-trace-state.ts:522-546`

`cachedPhaseStatus` (now carrying `planSpecHash`) is fingerprinted only on `.swarm/plan.json` size+mtime, while the underlying data is ledger-authoritative and the sibling `boundedApprovalCheck` cache in the same file uses a materially stronger fingerprint (SQLite `lastSeq:lastEventHash`, size+mtime only as fallback). The fingerprint code itself predates this PR; what's new is that a stale hit now falsely satisfies a fail-closed *security*-relevant check instead of just returning stale liveness data. Exploit window (byte-identical `plan.json` size *and* mtime across two saves) is narrow, not demonstrated reachable in practice.

**F-007 — LOW — Release fragment uses YAML frontmatter instead of this repo's `# Title` convention**
`docs/releases/pending/2600-trace-plan-binding.md:1-4`

All sampled sibling fragments open with an H1 heading; this one opens with `---\nissue: 2600\n...\n---`. `scripts/release-notes-fragments.mjs` does not strip frontmatter before joining fragments into the release body, so this will render as a stray heading/HR in the published GitHub Release. Won't fail CI (`check-pending-fragment.ts` only checks file existence).

### Pre-existing / not introduced by this PR (disclosed, not blocking)

- **SEC-001** (downgraded from an initial HIGH claim): `PLAN_BINDING_GATE` is directive-only (no tool-call-level enforcement) — but this is the pre-existing architecture of *every* row in this reducer, not a PR-specific weakness, and the plan/spec-drift relationship already has independent hard enforcement elsewhere (`enforceSpecDriftGate`, blocking `save_plan`/`update_task_status`/`phase_complete`) for the non-`isObligationPreserving` case.
- One-shot directive delivery is a non-atomic read-decide-write cycle with no lock (`issue-trace.ts:216-362`) — real, but identical to the pattern already used by every pre-existing gate row; no reachable concurrent-invocation path found in this codebase's single-hook-instance model.
- `isPlanCriticApproved` (`src/ci/evaluate.ts:505`) is unhardened by this PR, but binding is by design the trace engine's responsibility, not the CI shadow-eval path's.
- Deploying this PR fail-closes any in-flight trace whose plan lacks `specHash` — disclosed in the release fragment and PR body, intentional fail-closed direction.

### Disproved

- The original candidate framing "the PR's documented `/swarm reset` recovery leads to a *new* row-(f) stall" is wrong: `reset.ts:78` deletes `spec.md`, so the park actually occurs at row (d), identically at the base commit (row (d) was already an unconditional noop there too). The *real* `/swarm reset` problem is F-001's receipt-destruction issue, not a row-(f) allowlist gap.
- A candidate claiming the `planBoundToSpec` optionality "mirrors" `freshnessPermitted`/`traceValidationVerified`'s precedent was checking the wrong thing — those fields are required booleans; the actual (correct) precedent being followed is the `=== false` comparison idiom, not field optionality. Doc-wording nit, not a functional issue.

### Obligation mapping

All PR-body claims (plan binding, one-shot directives for rows d/g, `REPRO_GATE_LATE`, non-trace state hygiene, non-vacuous recurrence gate, skill-mirror reconciliation) were checked against the diff and are **MET** as literally stated. Issue #2600 AC5 (integration test through the real `record_*` tool chain to `published`) is also **MET** — covered by `tests/unit/hooks/issue-trace-journey-v3-completion.test.ts`, which the PR body correctly cites as "PRESERVING."

### Verdict

**APPROVE_WITH_NOTES.** No CRITICAL findings; no confirmed HIGH finding blocks EXECUTE-reachability or introduces a security bypass (F-001 is fail-closed, not fail-open — the cost is silence, not unsafe execution). F-001 is nonetheless a real, uncovered regression against this PR's own stated purpose and should be fixed in a fast-follow: convert the four new gate rows to negative-exclusion, extend row (f)'s allowlist (F-002), and qualify the `/swarm reset` directive text for the phases-complete case. F-003 through F-007 are minor, low-risk cleanup items.

---
*Validation provenance: Profile B (Claude Code, no swarm controller in this session) — 10 explorer lanes → 3 reviewer lanes (source re-read + differential base-vs-head execution probes + test-suite runs) → 1 critic lane (wider differential probe that overturned an initial reviewer downgrade). Full ledger, per-lane evidence, and probe scripts retained in this session's scratch workspace. Reviewer/critic independence was contextual (separate fresh-context subagents), not procedural.*

zaxbysauce:
## swarm-pr-review — PR #2837

**Reviewed head:** `26485be688169d2b443f833b91811505ff60a1a8`
**Merge base:** `eddf6e1fb6320bf3f39d71d8879524e2da433a78`
**Profile:** B (Claude Code native subagents, no controller) · **Depth tier:** M (1124/167 lines, 22 files, risk triggers: concurrency-state, api-schema-migrations, test-infrastructure)

### Pipeline
6 base-dimension explorer lanes + 5 matched risk-family micro-lanes (auth-identity-secrets, concurrency-state, api-schema-migrations, test-infrastructure, unclassified-risk — the other 6 families were evaluated and NOT_TRIGGERED: no auth/session code, no untrusted-input boundary, no subprocess/shell, no dependency/lockfile/CI change, no UI, no telemetry/logging change, no generated/vendored artifacts) → 3 independent opus reviewer batches (re-read source directly, ran live runtime probes against the actual hook, compared base vs head behavior) → 1 opus critic pass on the confirmed MEDIUM findings.

## PR intent

Binds the issue-trace engine to its loaded plan via a `planBoundToSpec` field (SHA-256 `specHash` equality between the plan and the current effective spec), parking execution with a `PLAN_BINDING_GATE` directive on mismatch. Adds one-shot directives to previously-silent reducer rows (`SPEC_MISMATCH_GATE`, `CRITIC_GATE`, `REPRO_GATE_LATE`). Stops non-`--trace` `/swarm issue` invocations from writing `issue-trace-state.json`. Rejects `dispositions: []` for recurrence-sweep receipts. Reconciles skill docs.

## Obligation mapping

| Obligation | Status | Notes |
|---|---|---|
| Plan bound to spec via specHash; foreign plan cannot reach EXECUTE | **MET** | Verified by execution probe: `computeNextMode` cannot return `EXECUTE` while `planBoundToSpec === false`, from *any* `lastTransition` value — the fail-closed park is structural (the `return noop` sits inside the same `if`), not conditional on the allowlist. |
| No silent stalls (rows d/g emit directives instead of bare no-ops) | **PARTIALLY MET** | New directives fire correctly for the ladder-monotone case (fresh trace or the immediately-preceding sentinel), per this file's own one-shot contract. They do **not** re-fire if the underlying condition regresses *after* the trace has already progressed past that row (e.g. spec edited while parked at `CRITIC_GATE`) — see Finding 1. Deliberate design per the module docblock, not a functional defect, but it means the PR's stated goal ("surface silent trace stalls") is not achieved universally. |
| REPRO_GATE_LATE reachable when plan exists | **MET** | Confirmed a real, additive improvement over base (which drove `EXECUTE`/`EXECUTE_TO_COMMIT` with zero reproduction evidence in this configuration — the exact bug #2600 exists to fix). |
| Non-trace `/swarm issue` writes no `in_progress` state | **MET** | Verified: every `issue-trace-state.json` write/unlink in `src/commands/issue.ts` is guarded by `if (traceState)`, including on the exception path. |
| `dispositions: []` rejected for a real defect class | **MET (with a disclosure gap)** | Enforced twice — Zod `.min(1)` and a manual reader-side check — but the manual check in `record-recurrence-sweep.ts` is dead code (the schema already rejects first); more importantly, the reader-side check retroactively invalidates any recurrence-sweep receipt legally written *before* this PR with an empty array, re-parking a trace that had already cleared that gate. Correct behavior, undisclosed consequence. |
| Skill mirrors byte-identical, docs reconciled | **MET** | `.claude/` and `.opencode/` `issue-ingest/SKILL.md` verified byte-identical; text matches actual gating. |

## Confirmed findings

**F-001 — LOW — Ladder-monotone one-shot gates don't re-arm on mid-flight regression**
`src/hooks/issue-trace-reducer.ts:215-234` (row d), `:329-346` (g-binding), `:354-372` (g-repro)

The new one-shot directive rows fire only when `lastTransition` is `null` or the sentinel of an immediately-preceding gate in the ladder — matching this file's own stated contract (`:139-140`, "fires from `lastTransition === null` or from the sentinel of a gate that precedes it in the ladder"). If the gated condition (spec binding, reproduction evidence) becomes newly-false *after* the trace has already progressed past that row, the row's outer condition matches but its inner allowlist doesn't, and it falls through to a bare `return noop` — no directive, though EXECUTE remains structurally unreachable in every case (the park is inside the same `if`, so this is an observability gap, not a safety hole). Recovery is `/swarm issue --trace` (resets `lastTransition` to `null`), but nothing tells the user to run it.

Two initial review passes read this as a HIGH-severity regression; an adversarial critic pass caught that (a) the "contradicts its own comment" framing was wrong — the comment's "any pre-plan sentinel" phrase is satisfied exactly by the allowlist, since `CRITIC_GATE`/`REPRO_GATE_LATE` are *emitted by rows below* the one being checked, and (b) a "regression vs base" claim for the `g-repro` row inverted the fix — base had no reproduction-evidence gate at all in this configuration and would drive `EXECUTE` unconditionally, which is the bug being fixed. Net: real, narrow, LOW.

*Suggested fix (optional hardening):* invert these rows to exclusion-based allowlists (`lastTransition !== 'OWN_SENTINEL'`), matching the pattern already used by rows i-pre1/2/3 a few hundred lines down in the same file — that family is structurally immune to this gap.

**F-002 — LOW — Test suite enumerates the implementation's allowlist rather than the requirement**
`tests/unit/hooks/issue-trace-plan-binding-2600.test.ts:216-229`

The parameterized "fires from these sentinels" test iterates the exact same list as the reducer's own allowlist. This is legitimate contract-pinning (not literally circular — it does test the documented one-shot contract), but it is drift-blind: extending the ladder with a new sentinel updates both lists in lockstep with no independent check, and there's no negative test asserting the row stays quiet at a later sentinel. Add one case per row driving the gate condition false from `CRITIC_GATE`/`PLAN_TO_EXECUTE` to pin the current (intentional) behavior explicitly.

**F-003 — LOW — Dead validation branch + vacuous regression test**
`src/tools/record-recurrence-sweep.ts:66,108,114`

The Zod schema's `.min(1)` on `dispositions` (line 66) rejects an empty array before the handler ever runs, making the manual `data.dispositions.length === 0` check at line 108 — and its purpose-built actionable message at line 114 — unreachable. Confirmed by a live probe: calling the tool with `dispositions: []` returns the generic Zod message, never the intended one. The new regression test (`tests/unit/tools/record-recurrence-sweep-empty-dispositions-2600.test.ts:79-106`) only asserts `success === false` and a `/disposition/i` message match, so it passes without exercising the code it appears to cover. Pick one gate (drop `.min(1)` and keep the actionable message, or drop the manual check) and assert the exact message.

**F-004 — LOW — Misleading directive text on two paths**
`src/hooks/issue-trace-reducer.ts:225-230` (SPEC_MISMATCH_GATE), `:340` (PLAN_BINDING_GATE)

- `SPEC_MISMATCH_GATE`'s directive always says to correct `.swarm/spec.md`'s `## Source Issue` number, but its trigger condition also fires purely on a stale/foreign `issue-trace-state.json` (`readTraceState` defaults `issueNumber: 0` when absent — `issue-trace-state.ts:420-424`), a case where editing `spec.md` is the wrong fix.
- `PLAN_BINDING_GATE`'s directive says to run `/swarm reset` to recover, but doesn't mention that `reset` also deletes `spec.md` and that the user must then re-run `/swarm issue --trace` to fully recover — omitting this doesn't cause a *second* stall (confirmed by tracing `reset.ts`'s file list against the reducer: the post-reset state correctly routes to the already-documented no-spec park, not the "inconsistent state" dead branch an earlier draft finding claimed), but the instructions are incomplete.

**F-005 — LOW — Undisclosed retroactive effect on pre-existing receipts**
`docs/releases/pending/2600-trace-plan-binding.md:11`, `src/hooks/issue-trace-state.ts:145`

The release note correctly states `dispositions: []` is now rejected, but doesn't mention that any recurrence-sweep receipt already on disk from before this PR (previously legal with an empty array) will now fail `_defaultRecurrenceSweepReceiptExists` and re-park a trace that had already passed that gate. Correct, fail-closed behavior; the note should say so in one sentence.

## Findings investigated and disproved / downgraded to informational

- **Cache-staleness of `planSpecHash` in SQLite-authoritative mode** (raised independently by 4 lanes) — disproved: the candidates conflated `.swarm/plan-ledger.jsonl` (the portable export, which SQLite mode can lack) with `.swarm/plan.json` (the projection the cache actually fingerprints, which every plan-mutating path rewrites regardless of authority mode). The fingerprint mechanism is byte-identical at the base commit.
- **Advisory-only enforcement of `planBoundToSpec`** ("could be bypassed by an agent that ignores the directive text") — downgraded to a doc-accuracy nit: every gate in this engine (including pre-existing ones) is directive-only, and a genuinely independent hard gate already exists elsewhere (`enforceSpecDriftGate` in `src/hooks/guardrails/index.ts`, wired before `save_plan`/`update_task_status`/`phase_complete`, comparing the same `specHash`) and blocks the main drift scenario.
- **Unconditional `computeSpecHash` per chat turn (no memoization), and wasted work for openspec/speckit-only projects** — confirmed as a real structural inefficiency (measured ~0.6ms/call, provably discarded before use when a project has no native `.swarm/spec.md`) but the initial HIGH/MEDIUM severity was unsupported by measurement; downgraded to LOW/optional hardening (gate the call on `workflowArtifacts.specExists`).
- **Fail-open default when `planBoundToSpec` is `undefined`** — confirmed as a type-level gap (the field is optional) but unreachable in production: the single call site (`src/hooks/issue-trace.ts`) always supplies a real boolean whenever `planExists` is true.
- Six of eleven repository-wide risk families (auth/session, untrusted-input boundary, subprocess/shell, dependency/build/release, UI, telemetry) were evaluated and are genuinely not triggered by this diff.

## Test / coverage gaps
- F-002 (no negative test pinning the intentional late-sentinel quiet behavior).
- F-003's regression test should assert the exact intended message, not just a substring match that a different code path also satisfies.

## Validation provenance
- 11 explorer/micro lanes (6 base dimensions + 5 matched risk families), all settled with structured output.
- 3 independent opus reviewer passes re-read source directly and, where relevant, ran live runtime probes against the actual hook (including a base-vs-head behavioral comparison) rather than trusting explorer citations.
- 1 opus critic pass caught and corrected two independent reviewers' severity inflation on the state-machine findings (both had over-read a docblock's informal gloss and mis-scored a base-comparison probe as evidence of harm rather than of the fix working).
- `bun test` run against all new/modified test files: 125 pass / 0 fail.
- No CRITICAL or HIGH findings survived review.

## Verdict: **APPROVE_WITH_NOTES**

No blocking issues. Five LOW-severity findings (F-001–F-005) are worth a small follow-up PR: they're all either drift-blind test hardening, dead code, or directive-text accuracy — none affect the correctness or fail-closed safety of the plan-binding gate itself, which held under every scenario probed, including live execution against the actual hook.

---
🤖 Generated with a Swarm PR review (Claude Code, Profile B — 11 explorer lanes, 3 independent reviewer passes, 1 critic challenge)

zaxbysauce:
## Swarm PR review — PR #2837 (fresh, independent)

**Reviewed head:** `26485be688169d2b443f833b91811505ff60a1a8` · merge base `eddf6e1fb` · 22 files, +1124/−167
**Pipeline:** 6 base-dimension explorer lanes + 5 matched risk-family micro-lanes (concurrency-state, api-schema-migrations, test-infrastructure, subprocess-platform, unclassified-risk; 6 families NOT_TRIGGERED with concrete absence evidence) → 34 raw candidates deduped to 28 → 2 independent reviewer shards running live probes and test execution in a bound worktree → adversarial critic on the sole surviving HIGH.

### Verdict: APPROVE_WITH_NOTES — no CRITICAL/HIGH survives; 5 MEDIUM disclosures/accuracy items fixed in a fast-follow

The plan-binding gate's fail-closed core **held under every adversarial probe**, including an exhaustive reducer sweep (15 sentinels × critic states with `planBoundToSpec=false`: zero EXECUTE outcomes) and real-file hook replays. The two conflicting HIGH claims from earlier reviews on this PR are both **disproven at head** (details below).

### Adjudication of prior review claims

| Prior claim | Disposition at head |
|---|---|
| Bot review's HIGH: "missing `planSpecHash === undefined` check → pre-spec plan reaches EXECUTE" | **DISPROVED** — the hook's expression `currentSpecHash !== null && currentSpecHash === phaseStatus.planSpecHash` is `false` for `undefined`; reducer parks at PLAN_BINDING_GATE; EXECUTE unreachable (issue-trace.ts:243-245, reducer:329). The behavior is intentional, test-pinned, and release-noted. |
| Swarm review A's HIGH: "binding gate permanently silences a completed trace" | **DISPROVED as HIGH** — terminal statuses short-circuit before the gate (reducer:161/170/187); for active traces the cost is a disclosed park, not silence; recovery documented. |
| This review's own intermediate HIGH (binding sentinel list omits CRITIC_GATE/REPRO_GATE_LATE → EXECUTE on unbound plan) | **KILLED by critic**: row g-binding ends in a hard `return noop` at reducer:345 when `planBoundToSpec === false`, making rows g/g-repro/h structurally unreachable while unbound. An exhaustive sweep produced zero EXECUTE outcomes. Residual (LOW polish): add those sentinels to the accepted list to match the row's own comment, so a mid-flight spec edit re-nudges instead of stalling silently. |

### Verified findings (all MEDIUM/LOW; fix incoming)

- **M1 (MEDIUM)** — Row (d) SPEC_MISMATCH_GATE fires only from `lastTransition === null` (reducer:215-235, probe-confirmed): a spec/issue mismatch arising after any later gate is silently suppressed until `/swarm issue --trace` resets. One-shot semantics are documented, but the mismatch class is the one the PR exists to catch.
- **M2 (MEDIUM)** — Release fragment and body say "no more silent stalls," but row (d) stays silent while `specIssueNumber === null` mid-spec-generation (reducer:232, test-comment pinned), and the binding gate emits PLAN mode for foreign plans, which the "PLAN only when no plan exists" wording doesn't cover. Disclosure accuracy.
- **M3 (MEDIUM)** — PLAN_BINDING_GATE's directive recommends `/swarm reset` without disclosing that reset deletes `spec.md` (and the receipt set) — reducer:340 vs `src/commands/reset.ts` filesToReset:88.
- **M4 (MEDIUM, test quality)** — The timeout fail-closed test passes even when the injected slow seam is replaced by an instant-resolve (mutation-proven in a worktree): it never exercises the 100ms bounded-approval timeout, and has no elapsed-time assertion (issue-trace-plan-binding-2600.test.ts:353-368).
- **M5 (MEDIUM, crash-window)** — Phase-status cache keys only on plan.json size+mtime (issue-trace.ts:188-195), never on ledger revision: a crash between save_plan's ledger append and plan.json write serves a stale `planSpecHash` until replay. Rare but real.
- **LOW set** — backward-compat: legacy plans without specHash now park (intended, documented, recovery named); `readPlanPhaseStatus` return shape gained an optional field (non-breaking); g/g-repro sentinel asymmetry; same-size+same-mtime cache non-invalidation; module caches persist across in-process sessions; partial trace-state corruption re-fires one-shots; read-only `.swarm/` re-delivers the directive each turn; no recovery-path test (park → re-save → resume); reader-level `dispositions: undefined` branch unpinned.

### Rejected candidates (transparency)

Dual-cache approval-staleness EXECUTE bypass (ledger-first save ordering + fresh per-cycle binding compute defeat it); post-completion spec-change stall (row h requires `!allPhasesComplete`; probed); concurrent double-delivery (composeHandlers serializes); anchor-receipt "wrong tree" claim (protocol misreading — the Phase-2.5 anchor binds the pre-implementation checkpoint tree by design); AC5 evidence dishonesty (the completion journey suite does reach `merge_approval_recorded`/`published`); raw-`mkdtempSync` test helper (pre-existing, untouched); unconditional rollback unlink (pre-existing at base); loose-regex assertion weakness (strict sentinel assertions co-located); spec-change stale-cache binding error (binding compares against a fresh spec hash each cycle); unmemoized `computeSpecHash` as a defect (pre-existing pattern, ~0.6ms, LOW materiality).

### Obligation check

Every AC1-AC7 in the body maps to real implemented behavior and tests (AC5's evidence claim is accurate — journey + completion suites cover the ladder through `merge_approval_recorded`/`published`). Micro attestation: 11/11 families evaluated.

**Fixes for M1-M5 + the LOW polish set are incoming on this PR.**

zaxbysauce:
## Feedback closure — all review findings resolved

Fixes at head `1440a4a56` (commits `1ee17f3e6` + `1440a4a56`). Closure ledger:

| id | source | item | outcome | evidence |
|----|--------|------|---------|----------|
| F1 (M1) | swarm-pr-review pr2837-r1 | row (d) spec-mismatch suppressed after later sentinels | FIXED | guard re-arms (own-sentinel one-shot kept); new reducer test fails when reverted (reviewer mutation probe) |
| F2 (M2) | swarm-pr-review pr2837-r1 | "no more silent stalls" / PLAN-wording overclaims | FIXED | fragment now "fewer silent stalls" + disclosed deliberate mid-generation silence; wording matches reducer |
| F3 (M3) | swarm-pr-review pr2837-r1 | PLAN_BINDING_GATE omits /swarm reset's deletions | FIXED | directive discloses spec.md + receipt deletion and the re-run step |
| F4 (M4) | swarm-pr-review pr2837-r1 | timeout test not timeout-dependent | FIXED | slow-true seam (250ms): broken timeout ⇒ row (h) advances ⇒ both assertions fail (reviewer mutation-verified) |
| F5 (M5) | swarm-pr-review pr2837-r1 | phase-status cache omits ledger revision | FIXED | ledger-revision fingerprint; unreadable authority suspends caching; crash-window invalidation probed |
| F6 (LOW) | swarm-pr-review pr2837-r1 | binding gate silent on post-plan sentinels | FIXED | re-arms from every later sentinel incl. REVIEW_GATE/EXECUTE_TO_COMMIT with allPhasesComplete=true; fail-closed hard-noop preserved — an unbound plan still can never reach EXECUTE (critic's exhaustive 14-sentinel × 3-variant sweep: zero leakage, no re-nudge loops) |
| F7 (LOW) | swarm-pr-review pr2837-r1 | no recovery-path test | FIXED | reducer recovery-chain test: park → plan corrected → repro → critic → EXECUTE |
| F8 (LOW) | swarm-pr-review pr2837-r1 | omitted-dispositions reader branch unpinned | FIXED | reader-level test: key omitted ⇒ gate fails |
| R1 | final critic | FR-006 cap: binding test file 572 lines | FIXED | split into reducer (222) + hook/command/journey (212) + shared helpers (188); 1:1 test parity; cap gate 0 violations |
| Copilot "needs a closer look" | bot review | no code finding | NO-ACTION | the closer look happened — binding fail-closed held under exhaustive probes |
| Drift notices ×2 | drift-check bot | RULESET_DIVERGENCE | NO-ACTION | known-advisory, non-blocking, pre-existing |
| Prior bot-review HIGHs ×2 | earlier PR comments | undefined-specHash bypass; completed-trace silence | DISPROVED | see review comment 5739140032 for the head evidence on both

zaxbysauce:
**Gate evidence for the closure above:** 47 fix-suite tests + a 94–118-test issue-trace blast radius + `tsc --noEmit` green at each head; independent reviewer mutation probes on F1 (reverted guard ⇒ new test fails) and F4 (instant-true seam ⇒ timeout test fails); final critic's exhaustive sentinel-state sweeps (no EXECUTE leakage, no re-nudge loops) plus `check:test-file-cap` / `check:test-clock` green. Reviewer and final critic both APPROVE at `1440a4a56` (reviewer round 1 APPROVE at `1ee17f3e6`; critic round 1 caught the FR-006 cap, resolved by `1440a4a56`).

zaxbysauce:
## 🤖 Multi-Stage PR Review
**Commit reviewed:** `1440a4a56ecb`

---

---

### 🔍 PR Intent

> Reconstructed obligation list (derived from PR title, linked issue #2600, commit messages, changed tests, changed public interfaces, and release fragment).

- **O-001** [DD-C001, AC1] — Bind the loaded plan to the current spec; a foreign plan must not reach EXECUTE, parked with a typed `PLAN_BINDING_GATE` directive.
- **O-002** [DD-C002, AC2] — Row (d) spec-mismatch and row (g) critic-not-approved must emit one-shot directives (not bare no-ops); zero-signal stall impossible while trace is active.
- **O-003** [DD-C005, AC3] — `/swarm issue` without `--trace` must not write an `in_progress` `issue-trace-state.json`; pre-existing trace state left untouched.
- **O-004** [DD-C009, AC4] — `dispositions: []` rejected for a real defect class by both the receipt reader and `record_recurrence_sweep`; "no defect class" fast path unchanged.
- **O-005** [AC6] — Reproduction gate must be reachable when a plan already exists (new `REPRO_GATE_LATE` row).
- **O-006** [DD-C008, AC7] — Issue-ingest skill `trace=true` transition text reconciled to match actual reducer behavior (both `.claude/` and `.opencode/` mirrors identical).
- **O-007** [Invariant 5] — Plan state read through authoritative ledger-aware loader (`loadPlan`), never projection-only. No PlanSchema/ledger/checkpoint schema change.
- **O-008** — Frozen acceptance checks C1–C7 pass at head, RED at base `eddf6e1fb`.

---

### 📦 Implementation Summary

The PR addresses five defect classes from issue #2600's 2026-09-06 audit:

1. **`WorkflowArtifacts.planBoundToSpec`** — optional boolean added to reducer interface; hook computes it as `currentSpecHash === phaseStatus.planSpecHash`. Explicit `=== false` guard; `undefined` stays transparent for legacy v2-shaped callers.

2. **Three new one-shot reducer rows** inserted between `!planExists` block and row (h): `(g-binding)` parks on `planBoundToSpec === false`; `(g-repro)` fires `REPRO_GATE_LATE` when a bound plan exists but no reproduction evidence; `(g)` now emits `CRITIC_GATE` instead of bare no-op. Every gate row fires from `null` or any preceding sentinel and waits quietly thereafter.

3. **Row (d) one-shot upgrade** — cross-issue spec mismatch emits `SPEC_MISMATCH_GATE` instead of silent no-op. Intentional silence preserved only when `specIssueNumber === null` (ingest mid-flight, no parseable `## Source Issue`).

4. **`handleIssueCommand` non-trace hygiene** — `traceState` is now `null` when `--trace` is absent; only trace invocations read old state, write new state, or perform rollback.

5. **Non-vacuous recurrence gate** — `dispositions.length > 0` added to receipt reader validation and Zod schema `min(1)`; tool rejects `dispositions: []` for a real defect class.

---

|---|---|---|
| O-001 | `SUPPORTED` | `src/hooks/issue-trace.ts:252-264` computes `planBoundToSpec`; `src/hooks/issue-trace-reducer.ts:318-337` parks at `PLAN_BINDING_GATE` |
| O-002 | `SUPPORTED` | reducer rows (d):207-224, (g):338-358 emit one-shot directives with named recovery |
| O-003 | `SUPPORTED` | `src/commands/issue.ts:246-259` — traceState is `null` unless `parsed.trace`; `if (traceState)` guards all state operations |
| O-004 | `SUPPORTED` | `src/hooks/issue-trace-state.ts:140` `dispositions.length > 0`; `src/tools/record-recurrence-sweep.ts:64` Zod `.min(1)` |
| O-005 | `SUPPORTED` | `src/hooks/issue-trace-reducer.ts:341-360` `REPRO_GATE_LATE` on plan-exists path |
| O-006 | `SUPPORTED` | `.claude/skills/issue-ingest/SKILL.md:66` and `.opencode/…`: byte-identical update |
| O-007 | `SUPPORTED` | `src/hooks/issue-trace-state.ts:532` calls `_internals.loadPlanFromLedger`; `planSpecHash` surfaced without schema change |
| O-008 | `CONTRACT` | Frozen checks C1–C7 documented in PR description; not re-executed here |

---

### 🚨 Confirmed Findings

_None. The PR is clean at the structural level. Analysis of every potential concern follows._

---

_None identified at ≥50% confidence that are not already covered by the PR's own mutation/frozen-check evidence._

---

### 🧪 Test / Coverage Gaps

_None found. The 19+7 regression tests in the new 2600 suites, updated reducer/hook/journey suites, and updated command suites provide comprehensive coverage of:_
- _Foreign plan parks at PLAN_BINDING_GATE, never reaches EXECUTE_
- _One-shot sentinel chaining (every gate fires once, waits quietly, re-arms from later sentinels)_
- _Undefined `planBoundToSpec` transparent to legacy callers_
- _Recovery chain: PLAN_BINDING_GATE → REPRO_GATE_LATE → CRITIC_GATE → EXECUTE_
- _Non-trace: no state file, pre-existing state byte-identical_
- _`dispositions: []` rejected by reader AND tool, controls green_

---

### 📋 Shipped-vs-Claimed Gaps

_None._

---

### 📝 Merge Recommendation

**[APPROVE]**

This PR is a targeted, well-tested fix for five audit-confirmed defect classes. The implementation correctly chains one-shot gates, handles the ledger-revision cache fingerprint (review pr2837-r1 F5), preserves legacy v2-shaped caller transparency with `=== false` guards, and keeps the "no defect class" fast path untouched. The non-vacuous recurrence gate closes a soundness gap. No regression was found in any category.

| Check | Result |
|---|---|
| No CRITICAL findings | ✅ |
| No hardcoded secrets | ✅ |
| Lockfile consistent | ✅ (no package changes) |

---

### 🔁 Validation provenance

**Confirmed findings kept:** None — no confirmed findings existed at review time (the reviewer-CONFIRMED list was empty), and the blind-spot pass found zero new defects.

**Dropped hypotheses (one-line reason):**
- `PLAN_BINDING_GATE directive contains unexpanded #N placeholder` — pre-existing pattern used identically in SPEC_MISMATCH_GATE, REPRO_GATE_LATE, CRITIC_GATE; not introduced by this PR.
- `REPRO_GATE_LATE never reachable when planBoundToSpec === false` — ordering is intentional; binding must be resolved before repro (the test suite covers both paths).
- `handleIssueCommand rollback path not guarded for non-trace` — the `if (traceState)` guard at the top of the catch block correctly prevents rollback on non-trace invocations; the `oldTraceState` read is also gated.
- `undefined planBoundToSpec causes undefined behavior` — the explicit `=== false` check means `undefined` transparently skips the binding gate; `true` also skips (plan is bound). The test confirms this contract.
- `computeSpecHash` not awaited in async hook context — `computeSpecHash` is synchronous; no await needed.
- `noReproWaiver in issueReference.flags` — this controls what the architect runtime receives; `reproductionPermitted` is a separate gate based on the `reproduction.json` receipt. These are distinct mechanisms serving different parts of the gate chain.

---

PR #2838, opened 2026-09-18
fix(config): make plan_cursor controls effective on both context paths

zaxbysauce:
Closes #2580

PR head: 081adaf41a9594b1bb3b6793ef5d3c87a359b486

## Summary

The `plan_cursor` configuration block (`enabled` / `max_tokens` / `lookahead_tasks`) was accepted by the schema and documented since v6.13, but no production code read it — every setting was silently ignored (audit finding CONFIG-F-02). This PR makes the advertised contract effective on both context paths and in the token accounting:

- One shared resolver (`resolvePlanCursorControls`, src/hooks/extractors.ts; schema defaults true/1500/2, bounds clamped) feeds Path A (default injection), Path B (opt-in `context_budget.scoring` candidates), and the context budget report.
- Path A: `enabled: false` suppresses the `[SWARM PLAN CURSOR]` injection (phase header and current-task context remain — the real pre-v6.13 behavior); `max_tokens` / `lookahead_tasks` shape the cursor.
- Path B: same gates and options — and the scoring path now emits the cursor candidate at all. Its plan.md read lived in a branch that only executes when the structured plan fails to load, while `loadPlan` migrates markdown-only plans into structured plans, so the candidate was structurally absent for every real workspace shape. It also gains Path A's DISCOVER-mode suppression for parity.
- Budget report: `planCursorTokens` now counts the actually-injected cursor (0 when disabled; canonical estimate with defaults) instead of a bespoke 30-line/first-1000-chars approximation that no code path injected.
- Extractor hardening found by the final critic: the documented `max_tokens` UPPER BOUND is now actually enforced — the compact-rebuild path previously had no final cap, so one pathological 12k-char task line emitted ~4014 tokens at `max_tokens: 500`. A structure-preserving final cap (closing marker reserved, ceil-boundary margin) bounds `estimateTokens(cursor) <= max_tokens` on every output shape.
- Docs corrected: README.md and docs/architecture.md no longer claim disabling "falls back to injecting the entire plan text" (that behavior never existed); release fragment shipped at docs/releases/pending/2580-plan-cursor-both-context-paths.md.

Default-config behavior: the non-scoring path's cursor text is byte-identical to the previous release. Two intentional observable deltas are named in the fragment (scoring sessions gain the cursor they were documented to have; budget-report token counts shift to the real cursor).

[...]

PR #2885, opened 2026-09-21
fix(hooks): sanitize remaining plan-derived context injections + surface BLOCKED phases in plan cursor (#2841)

### LINKED ISSUE #2841: Sanitize remaining plan-derived architect-context injections (phase/task one-liners bypass sanitizeContextText) + BLOCKED phases silently dropped from plan cursor
## Problem

PR #2838's post-merge swarm review (critic-verified, 2 passes) established that the `[SWARM PLAN CURSOR]` block was the largest — but not the only — unsanitized injection of user-writeable plan content into the architect system prompt. #2838's follow-up fixes the cursor via input-side sanitization inside `extractPlanCursor`. Two same-class gaps remain, both pre-existing (v6.12/v6.13):

1. **Phase/task one-liner injections bypass `sanitizeContextText`**: Path A `[SWARM CONTEXT] Phase:` (src/hooks/system-enhancer.ts:1484, via extractCurrentPhase/extractCurrentPhaseFromPlan) and Path B `[SWARM CONTEXT] Current phase:`/`Current task:` candidates (:2409, :2422) push raw plan-derived text. Every sibling architect-context injection (handoff :1558, decisions :1576, agent_context :1598, drift :2510+) routes through the sanitizer per its own MUST-sanitize docstring (src/hooks/context-sanitizer.ts:4-6).
2. **BLOCKED phases are silently dropped from the plan cursor**: the phase-header regex (src/hooks/extractors.ts:336) captures BLOCKED, but downstream filtering (status narrowing at :347-349, filters at :389-391) only handles COMPLETE / IN PROGRESS / PENDING, so a blocked phase vanishes from the cursor entirely.

## Threat model (why this matters)

plan.md is written by the architect after consuming untrusted task/issue/repo text — it is exactly the persistence channel a nested prompt injection would target; and a malicious repo can commit `.swarm/plan.md` (ensureSwarmGitExcluded only writes local .git/info/exclude; no provenance check on read), injecting on turn 1. Scope-guard protects `.swarm` writes for unauthorized agents, but the read path has no equivalent gate.

## Requested change

- Route the three one-liner injections through `sanitizeContextText` (same input-side pattern as the #2838 cursor fix).
- Either surface BLOCKED phases in the cursor (e.g. as a one-line `## Phase N [BLOCKED]` summary like PENDING) or document the drop intentionally.

## Acceptance criteria

- All plan-derived architect-context injections route through the sanitizer (grep-audited; guardrail test pins it).
- BLOCKED phase handling is either implemented or explicitly documented as intentional.
- No behavioral change for benign plans (sanitizer is idempotent and preserves benign markdown).

zaxbysauce:
Closes #2841

Follow-up to #2838/#2842 (the post-merge review lineage): closes the two remaining gaps the cursor hardening left open, plus two further defects found by this trace's own gates.

## Summary

- All six plan-text extractors in `src/hooks/extractors.ts` now route their plan-derived output through the shared `sanitizeContextText` — input-side for the markdown extractors (`extractCurrentPhase`, `extractCurrentTask`, `extractIncompleteTasks`), and on the composed string for the structured extractors (`extractCurrentPhaseFromPlan`, `extractCurrentTaskFromPlan`, `extractIncompleteTasksFromPlan`, sanitized before the `maxChars` truncation). This closes the last unsanitized plan-derived architect-context injections: the `[SWARM CONTEXT] Phase:` one-liner (default context path), the `Current phase:` / `Current task:` scoring candidates, and the `SWARM PLAN` / `SWARM TASKS` compaction facts.
- A phase carrying `[BLOCKED]` is no longer silently dropped from the `[SWARM PLAN CURSOR]`: the first blocked phase is surfaced as a one-line `## Phase N [BLOCKED]` summary (same shape as the PENDING one-liner) in both the full render and the compact rebuild, and the final `max_chars` cap reserves room for that summary ahead of generic tail truncation (found by the final critic: a pathological earlier section could otherwise re-drop it; when even the summary cannot fit the configured bound, the bound wins, as for every section — the reservation's fit arithmetic was itself corrected after the implementation reviewer found a +2 under-reserve).
- New guardrail suite `tests/unit/hooks/extractors-plan-sanitizer-coverage-2841.test.ts` (9 tests / 61 assertions): behavioral payload neutralization across all six extractors, a load-bearing source ratchet (consumed binding or sanitized return — a discarded call fails), BLOCKED surfacing in both renderers plus the hostile final-cap path and a narrow-band boundary sweep, an additive-only benign-with-BLOCKED cursor pin, and the documented construct-set boundary.
- Docs: README + `docs/architecture.md` Plan Cursor sections note the BLOCKED one-liner and its cap reservation; release fragment `docs/releases/pending/2841-plan-context-sanitizer-blocked-cursor.md`.

Benign plans are byte-identical before/after (sanitizer is idempotent and formatting-preserving; pinned by a frozen PRESERVING check).

## Invariant audit

- 1 (plugin init):       not touched — extractors are pure functions; no init-path change (`bun run check:invariants` all 8 checks passed)
- 2 (runtime portability): not touched — no new imports; `sanitizeContextText` already imported in extractors.ts
- 3 (subprocesses):      not touched — no subprocess changes
- 4 (.swarm containment): not touched — no new write paths
- 5 (plan durability):   not touched — ledger/replay path untouched
- 6 (test_runner safety): not touched — per-file test runs only
- 7 (test writing):      touched — new bun:test file, 343 lines (<500), no `mock.module`, static fs-based ratchet (evidence: `bunx biome ci src tests scripts` exit 0; typecheck clean; suite green)
- 8 (session state):     not touched — pure functions, no module state
- 9 (guardrails/retry):  not touched — no failure-classification changes
- 10 (chat/system msg):  not touched — one-liners ride existing `[SWARM CONTEXT]` injections; no new message shaping, no system-role entries
- 11 (tool registration): not touched — no tool/agent-map changes
- 12 (release/cache):    touched — pending release fragment added with unique slug (release-please owns version)

## Test plan

All at shipped head `6dfa723c`:

- Frozen acceptance checks C1–C5 via `repro-check.sh run` against base `82fab3002`: C1/C2/C3 base RED → head GREEN (C3 covers both the full render and the compact rebuild, proven by the compact-only `- Earlier:` marker), C4 PRESERVING base GREEN → head GREEN (byte-identity of all six extractors + cursor on benign plans), C5 NEW-SURFACE base ERROR (missing file) → head GREEN. All five PASS; `verify-checkpoint` and `verify-anchor` OK against the receipt published on the issue (comment 5755533105).
- Guardrail suite: `bun test tests/unit/hooks/extractors-plan-sanitizer-coverage-2841.test.ts` → 9 pass / 0 fail / 61 expect() calls.
- Consumer-cone suites (grep-derived, per-file): 18 extractor-referencing files + 5 compaction suites — 23/23 PASS.
- Gates: `bunx biome ci src tests scripts` exit 0 (unscoped local `lint:ci` fails only on git-ignored `.worktrees/` nested configs — local noise, base-worktree control passes); `tsc --noEmit` clean; `bun run check:invariants` all passed; `scan-deferred.sh` clean.
- Mutation probes (independent reviewer, 4 probes + reservation kill-switch): each fix surface has a probe that turns exactly its own assertions RED; tree restored clean each time.

Merge status: AWAITING_USER_APPROVAL (issue-tracer Phase 5.1 — merge requires separately recorded human approval).

<a href="https://cubic.dev/pr/ZaxbyHub/opencode-swarm/pull/2885?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Follow-ups

- Context.md-derived sanitization gap (compaction `SWARM DECISIONS` / `SWARM PATTERNS` facts) + the markdown `extractCurrentPhase` BLOCKED skip: tracked with full detail in #2886 — not bundled here (outside this issue's plan-derived scope).


zaxbysauce:
## Swarm PR Review — #2885

**Scope:** `82fab3002`..`6dfa723c0` (5 files, +510/-9): `src/hooks/extractors.ts`, `tests/unit/hooks/extractors-plan-sanitizer-coverage-2841.test.ts`, `README.md`, `docs/architecture.md`, `docs/releases/pending/2841-plan-context-sanitizer-blocked-cursor.md`. Depth tier: M (untrusted-input-boundaries risk trigger — prompt-injection sanitization). Profile B (Claude Code native subagents): 3 parallel explorer lanes → 1 independent opus reviewer (all 8 candidates re-verified with **executed** probes against the actual PR-head code, not just static reading) → 1 opus critic challenge on the 2 surviving PR-introduced findings.

### Intent vs. actual

|---|---|---|
| Route the 6 remaining plan-text extractors through `sanitizeContextText` | **MET** | All 6 call sites verified; every return path (including the `deps` interpolation) is covered by an outer or upfront sanitize call |
| Surface `BLOCKED` phases instead of silently dropping them | **PARTIALLY MET** | Works for the common case; two edge cases still silently degrade (below) |
| Closes #2841 | Issue is OPEN, diff addresses its stated scope (3 `system-enhancer.ts` one-liner sites + cursor BLOCKED-drop) | `gh issue view 2841` |
| CI green | **MET** | All checks pass (unit, coverage, quality, security, drift, smoke, etc.) |

### Confirmed findings (both LOW after critic review — no HIGH/CRITICAL survived)

**F1 (LOW) — Final-cap BLOCKED reservation guard checks the wrong string.**
`src/hooks/extractors.ts:627` — the guard `!trimmed.includes(blockedMarker)` checks only the bare heading substring (`## Phase N [BLOCKED]`), not the composed `blockedSummary` (heading + title line). When front-truncation happens to cut exactly at the end of the heading line, the code concludes the summary "already fits" and skips reservation, shipping a **dangling heading with no title**.
- Independently reproduced by executing `extractPlanCursor` directly (not just reading the code) at all three production-reachable budgets — the shipped default (`maxTokens=1500`, cap 4545 chars), the schema floor (500 → 1515 chars), and the ceiling (4000 → 12121 chars). The failure band is exactly `len(blockedTitle)+3` chars wide in each case.
- Reachability caveat (critic correction of the reviewer's initial framing): triggering it needs the *compact rebuild* itself to overflow `maxChars`, which at the default requires ~4400 unbounded chars in a single completed-phase task-summary bullet or the in-progress title — an unusual but not impossible plan.md shape.
- Severity held at LOW, not MEDIUM/HIGH: the heading itself always survives (only the title is lost), so output is strictly more informative than pre-PR base (which dropped BLOCKED entirely).
- **Verified fix:** the one-word change `blockedMarker` → `blockedSummary` at line 627 is sufficient — critic swept 9 configs post-patch with zero cap overruns and zero duplicate markers. The additional "strip trailing dangling marker" mitigation is provably unnecessary (the inner re-slice at `:634-639` always fires in the bug window since `markerLen ≥ 20 > 3`).
- Test gap: the existing narrow-band regression sweep (`extractors-plan-sanitizer-coverage-2841.test.ts:266-276`, `maxTokens` 19/20/21/5) exercises configurations **below the schema floor** (`min(500)` in `src/config/schema.ts:1339` and the `Math.max(500, …)` clamp at `extractors.ts:694`) and therefore cannot occur in production. A production-band sweep at `maxTokens=1500` asserting the *composed* BLOCKED summary (not just the marker) would close this gap.

**F2 (LOW) — Doc/release-note overclaim, two independent parts.**
This PR's own new text in `README.md` and `docs/architecture.md` states the BLOCKED summary "is never silently dropped … only when even the summary cannot fit the configured bound does the bound win." Two counterexamples inside the configured bound:
1. F1 above (title silently dropped within bound).
2. A **second, independent** gap: `phases.find((p) => p.status === 'BLOCKED')` (`extractors.ts:505`, reused at `:580`) surfaces only the *first* BLOCKED phase — reproduced with a 10-line plan (zero truncation involved) where a second BLOCKED phase vanished entirely. This mirrors the pre-existing single-item `pendingPhases[0]` convention (a consistent design choice, not a regression), but it contradicts the new "never silently dropped" wording and is untouched by the F1 fix.
- The release fragment (`docs/releases/pending/2841-…md`) is accurate on the first-only point ("surfaces the first blocked phase") — only `README.md`/`docs/architecture.md` need the qualifier added.
- Severity held at LOW (not HIGH): nothing is shipped to end users outside this PR's own diff; the fix is a same-PR doc correction.

### Lower-priority / advisory (LOW, confirmed but not blocking)

- **`extractors-plan-sanitizer-coverage-2841.test.ts:157-183`** — the "source ratchet" test meant to prevent a future edit from silently dropping a `sanitizeContextText` call is a textual-reappearance heuristic (`.match`, first-only; substring `indexOf` including comments), not a real dataflow check. A refactor like `const sanitized = sanitizeContextText(x); return x;` would pass it. The suite's separate behavioral assertions (payload neutralization) still catch a real regression, so this is a backstop weakness, not a live gap.
- **Test name overclaim**: the suite's docstring correctly scopes itself to "every plan-text extractor **in `src/hooks/extractors.ts`**", but the bare describe/test name reads repo-wide in CI output, while `src/services/handoff-service.ts` has structurally similar (but out-of-scope, pre-existing, `escapeHtml`-protected) composers the ratchet cannot see.

### Pre-existing / out of scope (not attributable to this PR)

- `extractDecisions`/`extractPatterns` (`extractors.ts`) feed `src/hooks/compaction-customizer.ts`'s `<swarm_compaction_facts>` block without `sanitizeContextText`, while the sibling `SWARM PLAN`/`SWARM TASKS` facts (from the same list) are now sanitized by this PR. Confirmed pre-existing (absent from this PR's diff) and **already disclosed** in the PR body's Follow-ups section, tracked at #2886.
- `src/services/handoff-service.ts` defines its own private `extractCurrentPhaseFromPlan`/`extractDecisions`, protected by `escapeHtml`/`sanitizeString` (narrower than `sanitizeContextText` — no zero-width/BiDi/code-fence/`system:`-prefix stripping), reachable via the `/swarm handoff` command's direct tool-output text. This does **not** contradict the release note's "handoff … already sanitized" claim — that claim refers to the `[HANDOFF BRIEF]` *injection* path (`system-enhancer.ts:1557`, which does wrap `sanitizeContextText`), a different surface from the slash-command response. Untouched by this PR, out of issue #2841's stated scope.

### Verdict: **APPROVE_WITH_NOTES**

Zero unresolved CRITICAL/HIGH findings. The sanitization work (this PR's primary goal) is correctly and completely wired for all 6 named call sites. The two confirmed LOW findings (F1, F2) are real but narrow-impact defects in the secondary BLOCKED-surfacing improvement, both with verified minimal fixes, and don't block merge — recommend addressing them in a fast follow-up (or this PR, maintainer's choice) rather than treating them as a merge blocker, since output in the failure case is still strictly better than the current (unpatched) main branch.

---
*Review methodology: 3 explorer lanes (haiku/sonnet) → 1 independent opus reviewer (executed probes, not static-only) → 1 opus critic (executed counter-probes, corrected the reviewer's reachability framing and rejected an unwarranted HIGH escalation). No finding was validated by the lane that discovered it.*

